PR #9, opened 2015-02-17
Incoming

zhangweikop:
I have add the the stringtrim function, please the review the code and give some suggestions.


zhangweikop:
I have committed the  ReadGraphemeCluster function and reverse. I guess I could also create the test via file for it.

PR #15, opened 2015-03-26
Add a parameter to FPSync, used to communicate the frontPanel Control that should be updated or marked as dirty. 

cipix2000:


PaulAustin:
I'll go ahead and do the pull. Tomorrow the function implementation might be changed.
though I didn't think this case would cause Eval.

cipix2000:
Paul.

Thank you.

I think I have messed something on the pull request... It appeared as two
commits, one of them an empty merge. I planned to try to create a clean
pull request and ask you to merge that.

I'm glad that I don't have to do that :)

Cipix

PR #17, opened 2015-03-30
Incoming

zhangweikop:
change numericsring.cpp
add more testing


PaulAustin:
Notes for the future:  For parsing simple integers just use SubString.ReadInt(). Constatnt's from enums or #defines generally prefixed with 'k' ( shouldn't look like locals). JavaScript experiments should be in their own file. Dont use emscripten functions that call Eval.

PR #30, opened 2015-05-15
Fixed Typos and proposed some variable name changes.

cglzaguilar:


PaulAustin:
Thanks Carlos, I agree on the size -> length change, Length is the term used mostly in Vireo, though other terms slip in from time to time.   Largest exception is AQSize. we can look at that and others Monday.

PR #48, opened 2015-08-10
WebSockets implementation

tnelligan:
WebSockets API implemented into Vireo


PaulAustin:
Need to remove the esh.exe .exp and .lib from make-it. 
Also vireo.js ( which is a binary of sorts) is moving.

PR #54, opened 2015-11-17
Incoming

zhangweikop:
please review the change


source/core/Array.cpp
```diff
@@ -127,6 +128,66 @@ VIREO_FUNCTION_SIGNATUREV(ArrayIndexEltNDV, ArrayIndexNDVParamBlock)
...
+VIREO_FUNCTION_SIGNATURE6(ArrayIndexElt2DV, TypedArrayCoreRef, StaticType, void, StaticType, void, void)
+{
+    TypedArrayCoreRef arrayIn = _Param(0);
+    Boolean pickRow = false;
+    Boolean pickCol = false;
+    IntIndex row, col;
+    IntIndex rank = 2;
+    TypeRef type = _ParamPointer(1);
+    IntIndex* lengths = arrayIn->DimensionLengths();
+    if (type->IsA("Int32")) {
```
PaulAustin: Should no need to check type at runtime for indexes. If just checking for "unwired" optional parameter then simply check data pointer for null.

source/core/Array.cpp
```diff
@@ -361,6 +543,74 @@ VIREO_FUNCTION_SIGNATURE3(ArrayRotate, TypedArrayCoreRef, TypedArrayCoreRef, Int
...
+    IntIndex endIndex = offset + length > arrayIn->Length()? arrayIn->Length() : offset + length;
+
+    if (endIndex <= 0) {
+        return _NextInstruction();
+    }
+    IntIndex arrayOutLength = arrayIn->Length() - (endIndex - startIndex);
+    arrayOut->Resize1D(arrayOutLength);
+    if (startIndex > 0) {
+        arrayOut->ElementType()->CopyData(arrayIn->BeginAt(0), arrayOut->BeginAt(0), startIndex);
+
```
PaulAustin: avoid extra blank lines at end of block. After the block is Ok

test-it/ArrayDelete.via
```diff
@@ -0,0 +1,42 @@
...
+ 	e(a(.Int32 *)     outArray)
+	e(a(.Int32 *)     deletedArray)
+	e(dv(.Int32 71) elem)
+	e(a(.Int32 *)     outArrayNeg)
+	e(a(.Int32 *)     deletedArrayNeg)
+	e(dv(.Int32 0) elemNeg)
+    )
+    clump(1 
+	Println("testing array delete")
+	//arguments: output deleted input length offset
```
PaulAustin: Why is the indentation so ragged in the test files?

PaulAustin:
I added some comments to the PR. There are some other question we can cover
1 on 1
-Paul

PR #58, opened 2016-05-06
Incoming

zhangweikop:
clean up the work from last pull request. 
The Memory work is deleted from this request and moved to the open issue.


zhangweikop:
Add MaxAndMin instruction.

PaulAustin:
Thanks for adding the test file.  As a next step it it will help to added some some more detailed tests for strings.    'abcd' and 'abcde' along with some comments on why the answer should be what it is.   this is a great start.

PR #64, opened 2016-06-08
Implement RoundToNearest Vireo Function and associated test.  Fix Pi* tests.

spathiwa:


cglzaguilar:
In the code there was a case with kEncoding_IEEE754Binary, with a switch for aqSize, in which the literals 0, 4, 8 were used. I would be great if we update these magic numbers with named constants to improve the clarity of the code as we continue to work in Vireo.

spathiwa:
These really are byte sizes, not magic numbers, so I think it is fine leaving them as is; in fact, since case 4 uses Int32s, case 8 uses Int64, having the actual values makes it really clear what the numbers mean.

Craig

PR #73, opened 2016-07-13
Fix endianness issues in flatten support; remove tests from blacklist

spathiwa:
Also add python script to patch emscripten library bug preventing it from outputting strings with embedded \0's in them.


cglzaguilar:
@spathiwa: I noticed the travis-ci failed on native tests for the "binary Flatten tests to use Big Endian Flat..." submission. I'm guessing this is already fixed on recent submissions. I just want to confirm.

spathiwa:
Yup. The emscrpten Python patcher needed a tweak to work with debughing off.

PaulAustin:
Fred Visser had vireo running on a MIPs based SOC he found on kickstarter the I think the endian test were the only ones that failed, this might have fix those.

PR #78, opened 2016-07-19
Updated Blacklisted Tests

DevinCarr:
Removed a few of the tests from the Blacklist that we will not expect to fix. They are mostly related to the GPIO and having Vireo talk to embedded systems and hardware. These are not expected to be integrated into the main Vireo core currently.


spathiwa:
Thanks for updating the README with the new options/-h output.
I forgot to do this with my change.

PR #95, opened 2016-08-22
Add NumberTo{Float,Exponential,Engineering,Hex,Octal,Binary}String

spathiwa:
Add:
        NumberToFloatString(number width precision stringOut)
        NumberToExponentialString(number width precision stringOut)
        NumberToEngineeringString(number width precision stringOut)
        NumberToHexString(number width stringOut)
        NumberToOctalString(number width stringOut)
        NumberToBinaryString(number width stringOut)
    All polymorphic on number (which can be a scalar number of array/cluster
    of numeric type), and stringOut (which is scalar string or array/cluster
    of strings mathching shape of input).


spathiwa:
You also mentioned wanting Boolean to Number.
The LabVIEW "Boolean to (0,1)" actually only produces an Int16 output (for historical reasons), and the Vireo version of this already exists as "BooleanConvertInt16(boolIn int16out)"

PR #104, opened 2016-11-02
Add polymorphic predicate functions IsEmptyString, IsDecimalDigit, etc.  See commit comment for full ist.

spathiwa:


cglzaguilar:
Hi Craig, 

Have you removed all your changes and see if travis-ci still passes?

Also consider updating travis-ci to test only once vireo.js for now. I'll create a defect to look into this issue later on.

PR #114, opened 2016-11-29
Fix a compiler error and polymorphic trigonometric functions

sanmut:


cglzaguilar:
Sankara, please update the spaces in the test-it/testList.json file.

PR #115, opened 2016-11-29
Fix compiler error and polymorphic trigonometric functions

sanmut:


spathiwa:
Please add tests for the newly polymorphic trig functions as well.  Modifying existing tests for those functions to also call them on an array or cluster is sufficient.

PR #121, opened 2016-12-06
Emscripten JS modularization

rajsite:
Vireo is now broken up into several JS modules instead of having the JS implementations largely in a single file.

Initial JS lint rules have been created but it has not been enabled in CI builds yet.


cglzaguilar:
@spathiwa and @sanmut, please look at the change set and make comments.

spathiwa:
I already reviewed the code and discussed it with Milan in person.

rajsite:
Adding @jattasni to help review JS changes

jattasNI:
Milan and I talked in person and the changes look good. Only action item was to add a bit more documentation for each of the modules, either in a higher-level read-me doc or inline in each of the module definitions.

PR #124, opened 2016-12-07
Support >2D arrays with ArraySubset and ArrayReplaceSubsetNDV

spathiwa:


spathiwa:
Just made another pull request renaming the multi-dim ArrayReplaceSubset function.
Please pull this before implementing DFIR changes to use it.

See my other email about NDV vs. other array functions.

Thanks
Craig

PR #125, opened 2016-12-07
Format string fixes and minor cleanup of Makefile,etc

sanmut:


spathiwa:
The term 'localization codes'  is incorrect/imprecise.  Please use 'localized decimal separator' everywhere.  Localization codes could mean lots of other things.

spathiwa:
The StringFormat prim needs to support \xx escape codes too.
Printf is only used for stdout debugging and isn't useful inside G programs.
The Format() function in NumericString.cpp needs to be modified the same way you did Printf.

sanmut:
@spathiwa  Updated code to use 'localized decimal separator'. StringFormat and Printf both use  Format() function in NumericString.cpp -- which is what is fixed in the commits in this pull request. I verified that with the built vireo.js that a failing LabVIEW VI with "Format Into String" node actually passes now.

spathiwa:
Ok, sorry, I didn't notice that was where the change was so it will affect StringFormat().
In that case, it looks fine.  It would probably still be a good idea to modify one of the StringFormat unit tests to explicitly test this case as well.  Thanks

sanmut:
spathiwa wrote " It would probably still be a good idea to modify one of the StringFormat unit tests to explicitly test this case as well.".
Done now as a separate commit.

PR #126, opened 2016-12-07
Rename ArrayReplaceSubsetNDV to just ArrayReplaceSubset, overloading 1D

spathiwa:
Add ArrayIndex prim, and rename ArrayReplaceSubsetNDV to just ArrayReplaceSubset, overloading 1D version.  The NDV versions of other functions take column major indexes, so the new one needs to avoid using that name since it's row major.


cglzaguilar:
Sounds good to me. I like smaller simple names. We can remove the other functions once the new ones had been integrated.

PR #127, opened 2016-12-07
Add StringFormat tests for localized decimal separator and ASCII hex …

sanmut:
…codes


sanmut:
Addressing Craig's comments by adding tests in StringFormat

spathiwa:
Looks good.  You just need to add the test to testList.json so it runs.

sanmut:
@spathiwa I have modified an existing .via file that is already present in testList.json

PR #134, opened 2016-12-14
Updates to Wait, WaitUntilMultiple and GetTickCount primitives and other minor fixes

cglzaguilar:
2. Fixed millisecond assumption on VireoOS_emscripten
3. Updated Wait, WaitUntilMultiple and GetTickCount primitives
4. Added tests for Wait, WaitUntilMultiple and GetTickCount primitives
5. Assigned proper value to _SystemLocaletimeZone under Windows


cglzaguilar:
@rajsite and @spathiwa if you have a change, please review the change set.

spathiwa:
s/WaitUntilMilliSecondsMultiple/WaitUntilMillisecondsMultiple/
to be consistent with capitalization elsewhere.

spathiwa:
I would also try to avoid all the copied code in all the different arg type variants, but we can clean that up later.

PR #135, opened 2016-12-16
Refactoring Timestamp.cpp

cglzaguilar:
It contains more responsabilities than desired. Splitted the content in three
	Date.cpp, that holds the Date class
	Timestamp.cpp, that holds the Timestamp class
	TimeFunctions.cpp, that contains the definition for Vireo Functions


cglzaguilar:
@spathiwa I was unable to fix the XCode project. When you have a chance please fix it.

PR #136, opened 2017-01-03
String formatting improvements and error hanndling

sanmut:
Added a new format specifier called "%z" that can format any type. Restrict "%s" specifier to only supported types and throw error on unsupported types.


spathiwa:
Uhm, what motivated this change?
Also, why did you allow Boolean to work for %s but disallow all other non-string types?

CreateMimatchedFormatSpecifierError() is misspelled.

sanmut:
I was trying to make %s compatible with LabVIEW so that existing VI based tests can be run on Vireo as well. LabVIEW supports boolean type with %s modifier. Fixed the misspelled CreateMimatchedFormatSpecifierError now. Thanks :-)

PR #137, opened 2017-01-03
Cleaned code for time related functions. Partially removed usage of h…

cglzaguilar:
…ardcode numbers and removed abbreviations

### DIFF
--- a/source/core/Date.cpp
+++ b/source/core/Date.cpp
@@ -14,9 +14,6 @@ SDG
 #include "TDCodecVia.h"
 #include "Date.h"
 
-#include <math.h> /* fabs */
-#include <float.h> /* DBL_EPSILON */
-
 #if defined(VIREO_DATE_TIME_STDLIB)
 #if kVireoOS_win32U
     #include <windows.h>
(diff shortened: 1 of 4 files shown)


source/core/Date.cpp
```diff
@@ -250,17 +251,18 @@ namespace Vireo {
...
-        _SystemLocaletimeZone = -int((timeZoneInfo.Bias + timeZoneInfo.DaylightBias)* 60);
+        _systemLocaleTimeZone = -int((timeZoneInfo.Bias + timeZoneInfo.DaylightBias) * secondsPerMinute);
 #endif
-        return _SystemLocaletimeZone;
+        return _systemLocaleTimeZone;
     };
+
     //------------------------------------------------------------
-    Int32 Date::isDTS()
+    Int32 Date::isDayTimeSaving()
```
sanmut: It is commonly referred to as DaylightSavingTime (or abbreviated as DST) and I would go with either that full name or abbreviation.

source/core/Date.cpp
```diff
@@ -58,37 +55,42 @@ extern "C" {
 
 namespace Vireo {
 #if defined(VIREO_TYPE_Timestamp)
+
+    static const Int64 secondsInYear = 31536000;
+    static const Int64 secondsInLeapYear = 31622400;
+    static const Int32 secondsPerMinute = 60;
+    static const Int32 secondsPerHour = secondsPerMinute * 60;
```
sanmut: I think the convention to name the constants is prefix with 'k' (such as kTempFormattingBufferSize). I would rename the constants.

source/core/Date.cpp
spathiwa: Agree with sanmut.  I suspect DTS was just a typo in the first place.

PR #139, opened 2017-01-03
Make timestamp conversion prims take cluster as well as built-in LVDateTimRec type

spathiwa:


source/core/TimeFunctions.cpp
```diff
@@ -226,9 +226,10 @@ DEFINE_VIREO_BEGIN(Timestamp)
     DEFINE_VIREO_FUNCTION_CUSTOM(Convert, TimestampConvertDouble, "p(i(Timestamp) o(Double))")
     DEFINE_VIREO_FUNCTION_CUSTOM(Convert, ToTimestamp, "p(i(Double) o(Timestamp))")
 
-    DEFINE_VIREO_TYPE(LVDateTimeRec, "c(e(Double fractional_sec) e(Int32 second) e(Int32 minute) e(Int32 hour) e(Int32 day_of_month) e(Int32 month) e(Int32 year) e(Int32 day_of_week) e(Int32 day_of_year) e(Int32 dst))");
-    DEFINE_VIREO_FUNCTION(DateTimeToTimestamp, "p(i(LVDateTimeRec) o(Timestamp))")
-    DEFINE_VIREO_FUNCTION(TimestampToDateTime, "p(i(Timestamp) o(LVDateTimeRec))")
+#define kLVDateTimeTypeStr  "c(e(Double fractional_sec) e(Int32 second) e(Int32 minute) e(Int32 hour) e(Int32 day_of_month) e(Int32 month) e(Int32 year) e(Int32 day_of_week) e(Int32 day_of_year) e(Int32 dst))"
+    DEFINE_VIREO_TYPE(LVDateTimeRec, kLVDateTimeTypeStr);
+    DEFINE_VIREO_FUNCTION(DateTimeToTimestamp, "p(i(" kLVDateTimeTypeStr ") o(Timestamp))")
+    DEFINE_VIREO_FUNCTION(TimestampToDateTime, "p(i(Timestamp) o(" kLVDateTimeTypeStr "))")
```
sanmut: Do we need to keep the existing two lines as-is (as below) and add the two new functions, if I understand the title of this commit correctly?
DEFINE_VIREO_FUNCTION(DateTimeToTimestamp, "p(i(LVDateTimeRec) o(Timestamp))")
DEFINE_VIREO_FUNCTION(TimestampToDateTime, "p(i(Timestamp) o(LVDateTimeRec))")

spathiwa:
No.  The type system seems to handle down conversion from LVDateTimeRec to this structure automatically, even though it doesn't work the other way.

PR #148, opened 2017-01-05
Reduces noise in build output from karma

### LINKED ISSUE #147: Try to reduce karma build output
Looks like some less noisy reporters exist, worth a shot: https://github.com/jaapz/karma-failed-reporter

http://stackoverflow.com/questions/18121197/karma-runner-console-output-only-failed-tests

Tried dots at one point https://github.com/ni/VireoSDK/commit/01ae063e4861abdfa38b2b7b1668ed6db4fddc02

but didn't seem to help. maybe because of concurrency? Can try one browser at a time instead and dots might behave better...

rajsite:
I tried this same thing before by switching to the dots reporter : https://github.com/ni/VireoSDK/commit/01ae063e4861abdfa38b2b7b1668ed6db4fddc02

But the problem was that concurrency was still set to greater than 1. Because of that each dot would print along with the name of the browser

By setting concurrency to 1, the browser name only prints once and the output looks much better

image

Since we only test two browsers right now reducing concurrency is not a big impact on test time

Fixes #147


spathiwa:
Awesome, thanks for addressing this.

PR #149, opened 2017-01-06
Updated DateTimeRecordToTimestamp to take into account UTC

cglzaguilar:


test-it/DateTimeRecordToTimestamp.via
```diff
@@ -0,0 +1,74 @@
...
+        Copy(local39 dataItem_Actual )
+        //Printf ("All tests pass: '%s'.\n" dataItem_Actual)
+        Printf ("    UTC On pass: '%s'.\n" dataItem_UTCOn)
+        Printf ("        value: '%s'.\n" dataItem_UTCOnDateAndTimeString)
+        //Printf ("    UTC Off pass: '%s'.\n" dataItem_UTCOff)
+        //Printf ("        value: '%s'.\n" dataItem_UTCOffDateAndTimeString)
+/* Clump Ended. */    )
+)))
+enqueue (_%54estVI_DTRecordToTimestamp%2Egvi)
+//Finished!! :D
```
sanmut: Please insert a newline at the end of this file.

test-it/DateTimeRecordToTimestamp.via
```diff
@@ -0,0 +1,74 @@
...
+        IsEQ(local33 g26 local35)
+        Copy(local35 dataItem_UTCOff )
+        IsEQ(local31 g27 local37)
+        Copy(local37 dataItem_UTCOn )
+        And(local35 local37 local39)
+        Copy(local39 dataItem_Actual )
+        //Printf ("All tests pass: '%s'.\n" dataItem_Actual)
+        Printf ("    UTC On pass: '%s'.\n" dataItem_UTCOn)
+        Printf ("        value: '%s'.\n" dataItem_UTCOnDateAndTimeString)
+        //Printf ("    UTC Off pass: '%s'.\n" dataItem_UTCOff)
```
sanmut: As we are not going to make the UTCOff case pass but handle it manually in the near future, I would fork this test into a separate test and put the .via and .vtr files under test-it/ManualTests and test-it/ManualTests/results respectively. We will soon make test.js to run manual tests. We can try to automate the locale dependent manual tests at some point.

source/include/Date.h
```diff
@@ -43,13 +43,15 @@ class Date {
...
-        static void getDate(Timestamp timestamp, Int64* secondsOfYearPtr, Int32* yearPtr,
+    private:
+        void getDate(Timestamp timestamp, Int64* secondsOfYearPtr, Int32* yearPtr,
             Int32* monthPtr = NULL, Int32* dayPtr = NULL, Int32* hourPtr = NULL,
             Int32* minutePtr = NULL, Int32* secondPtr = NULL, Double* fractionPtr = NULL,
             Int32* weekPtr = NULL, Int32* weekOfFirstDay = NULL, char** timeZoneString = NULL);
+
+    public:
+        Date(Timestamp timestamp, Int32 timeZoneOffset);
+        ~Date();
```
sanmut: It looks like you are making the existing ~Date() to fail. I suspect you have not pulled from ni/incoming recently. I think you should pull and merge your changes on the latest.

source/core/TimeFunctions.cpp
```diff
@@ -166,11 +166,14 @@ namespace Vireo {
...
     //------------------------------------------------------------
-    VIREO_FUNCTION_SIGNATURE2(DateTimeToTimestamp, LVDateTimeRec, Timestamp)
+    VIREO_FUNCTION_SIGNATURE3(DateTimeToTimestamp, LVDateTimeRec, Boolean, Timestamp)
     {
         LVDateTimeRec *dt = _ParamPointer(0);
+        Boolean isUTC = _ParamPointer(1) ? _Param(1) : false;
+        Int32 timeZoneOffset = isUTC ? 0 : Date::getLocaletimeZone();
         Timestamp timestamp(dt->fractional_secs, dt->second, dt->minute, dt->hour, dt->day_of_month, dt->month, dt->year);
-        _Param(1) = timestamp;
+        Timestamp local = timestamp - timeZoneOffset;
```
sanmut: `Timestamp local = timestamp - timeZoneOffset;`
This is misleading/confusing to follow the code (and can possibly be slightly incorrect). We should subtract only Timestamp._integer part and Timestamp._fraction should not be touched. It is unfortunate that there is an operator+ overload for an Int64 argument but there is not an operator- overload for an Int64 argument. Instead of adding a new operator- overload with Int64 argument to fix this and make the code clearer,  I think we should remove the existing operator+(Int64) overload and add a new method in Date class (called UpdateTimeZoneOffset) and call that.

source/core/Date.cpp
```diff
@@ -60,14 +60,6 @@ namespace Vireo {
     Int32 Date::_systemLocaleTimeZone = 0;
 
     //------------------------------------------------------------
-    Int64 SecondsFromBaseYear(Int64 year, Int64 baseYear)
```
sanmut: If this is written fairly recently and may be useful in the near future, you can just #ifdef out this for now. We can delete it once the feature is close to completion.

test-it/DateTimeRecordToTimestamp.via
spathiwa: I actually plan on fixing this sooner rather than later, so I do not think we need to fork off tests just yet.

spathiwa: (Fixing UTFOff case I mean)

source/core/TimeFunctions.cpp
spathiwa: I mildly disagree with this.  I think overloading + is well-defined to operate only on the _Integer portion.

sanmut: I am a bit biased too on this. I am ok with adding an overload for operator- (which is what Carlos is doing). I hate that no source code browser tool / IDE can show the usages of an overloaded operator function in a C++ project even today :-(

PR #151, opened 2017-01-06
Adding substract overload to timestamp

cglzaguilar:
Addressing Shankara feedback.

### DIFF
--- a/source/core/Timestamp.cpp
+++ b/source/core/Timestamp.cpp
@@ -163,6 +163,14 @@ namespace Vireo
         return answer;
     }
     //------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


source/core/Timestamp.cpp
```diff
@@ -163,6 +163,14 @@ namespace Vireo
         return answer;
     }
     //------------------------------------------------------------
+    Timestamp const Timestamp::operator-(const Int64& value)
```
sanmut: This is ok, but it is a good idea to invoke operator+() directly with "-value" as an argument.

spathiwa: Agree.

PR #155, opened 2017-01-10
Refactor Date class to use localtime_r for platforms which support it, including emscriptem/JS.  Enable locale-specific tests.

spathiwa:


spathiwa:
The locale-specific tests expect results to match the Central Time Zone.
TravisCI has this set in the environment, but if anyone runs the test suite manually in another TZ, those tests may fail.  Milan and I discussed making TravisCI run a different set of tests than the default test list used by test.js, so the default doesn't include locale tests but Travis does, but I haven't made that change yet since it involves splitting some tests into parts.

PR #161, opened 2017-01-16
Implement dots report for test.js

rajsite:
When the --dots flag is used will show a dot for passing tests and the name of the test for failing tests. Without the flag the original format is used.

You can see example passing output here: https://travis-ci.org/rajsite/VireoSDK/builds/192517628#L1325

And example failing output here: https://travis-ci.org/rajsite/VireoSDK/builds/192520625#L1327

Let me if you are cool with this change @cglzaguilar @sanmut @spathiwa :+1:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,10 +15,10 @@ vjs js:
 test: testnative testjs
 
 testjs:
-	cd test-it; ./test.js -j
+	cd test-it; ./test.js -j --dots
 
 testnative:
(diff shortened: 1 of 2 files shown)


spathiwa:
Looks good to me.

sanmut:
Looks good. If there is not a pressing need for the behavior to see the test results output without dots, then I would remove the command line argument and just do it by default.

PR #162, opened 2017-01-16
Add RefNum support, and simple unit test framework with test for RefN…

spathiwa:
…ums.

### DIFF
--- a/make-it/EmMakefile
+++ b/make-it/EmMakefile
@@ -10,6 +10,7 @@ SITEDIR=$(VIREO_SDK_DIR)/Documents/gh-pages/playground
 TARGETDIR=$(VIREO_SDK_DIR)/dist
 CORESOURCEDIR=$(VIREO_SDK_DIR)/source/core
 IOSOURCEDIR=$(VIREO_SDK_DIR)/source/io
+UTSOURCEDIR=$(VIREO_SDK_DIR)/source/unittest
 
 .PHONY: clean install help vjs dox
 .DEFAULT_GOAL=help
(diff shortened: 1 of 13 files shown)


make-it/EmMakefile
```diff
@@ -10,6 +10,7 @@ SITEDIR=$(VIREO_SDK_DIR)/Documents/gh-pages/playground
 TARGETDIR=$(VIREO_SDK_DIR)/dist
 CORESOURCEDIR=$(VIREO_SDK_DIR)/source/core
 IOSOURCEDIR=$(VIREO_SDK_DIR)/source/io
+UTSOURCEDIR=$(VIREO_SDK_DIR)/source/unittest
```
rajsite: I guess at some point in the future we want a release build where this is not compiled in?

spathiwa: Actually, the body of the cpp files are commented out #if VIREO_UNIT_TEST is not 1.
The only code that's always compiled in is the code that conditionally runs the test list if it's not empty, which is pretty small.

rajsite: Excellent :+1:

source/core/UnitTest.cpp
```diff
@@ -0,0 +1,39 @@
+/**
```
sanmut: Did you consider using one of the existing Unit Test frameworks? GoogleTest is widely used in many teams in NI. As the test code and framework should not be compiled into and thus not really affect the side of the generated vireo.js, it sounds like a good idea to use an existing established framework. Thoughts?

spathiwa: I did, and I think it's something we could fairly easily migrate too, but it seemed overkill for a handful of tests, and more work to incorporate into the build for all platforms.

PR #164, opened 2017-01-20
analog waveform support

sanmut:


source/core/Waveform.cpp
```diff
@@ -0,0 +1,54 @@
...
+{
+    _Param(0)._t0 = _Param(1);
+    _Param(0)._dt = _Param(2);
+
+    TypedObjectRef* argY_source = _ParamPointer(3);
+    TypedObjectRef* waveY_dest = &_Param(0)._Y;
+    SubString waveY_elemType =  (*waveY_dest)->ElementType()->Name();
+    if (! (*argY_source)->ElementType()->Name().Compare(&waveY_elemType) )
+    {
+        THREAD_EXEC()->LogEvent(EventLog::kHardDataError, "AnalogWaveformBuild() Type of argument-3 does not match with type of output waveform");
```
spathiwa: s/does not match with type of/does not match type of/

source/core/Waveform.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+VIREO_FUNCTION_SIGNATURE4(AnalogWaveformBuild, AnalogWaveform, Timestamp, Double, TypedObjectRef)
+{
+    _Param(0)._t0 = _Param(1);
+    _Param(0)._dt = _Param(2);
+
+    TypedObjectRef* argY_source = _ParamPointer(3);
+    TypedObjectRef* waveY_dest = &_Param(0)._Y;
+    SubString waveY_elemType =  (*waveY_dest)->ElementType()->Name();
+    if (! (*argY_source)->ElementType()->Name().Compare(&waveY_elemType) )
```
spathiwa: Can you use IsA here instead of directly comparing type names?
I've think type comparison based on names is iffy, and while I realize that's what IsA actually does, I'd rather have the code isolated so it can be refactored more easily.

source/core/Waveform.cpp
```diff
@@ -0,0 +1,54 @@
...
+DEFINE_VIREO_TYPE(DigitalWaveform, "c(e(a(UInt8 * *) data)e(a(UInt32 *) transitions))")
+DEFINE_VIREO_END()
+#endif
+
+VIREO_FUNCTION_SIGNATURE4(AnalogWaveformBuild, AnalogWaveform, Timestamp, Double, TypedObjectRef)
+{
+    _Param(0)._t0 = _Param(1);
+    _Param(0)._dt = _Param(2);
+
+    TypedObjectRef* argY_source = _ParamPointer(3);
```
spathiwa: Probably use TypedArrayCoreRef* for these type (and for the last func arg) to more clearly indicate it's an array

spathiwa:
Except for those minor quibbles, the code looks good to me.

source/core/Waveform.cpp
sanmut: Done. Thanks.

source/core/Waveform.cpp
sanmut: Agree with all your comments. My bad. Done.

source/core/Waveform.cpp
sanmut: Good catch. Will make it more readable and explicit. Thanks.

PR #170, opened 2017-01-25
Add named queue support.

spathiwa:
QueueRefNumManager class keeps track of look up table from name
to canonical refnum allocation in TypedRefnum storage, then clones
the refnum so every user of the name gets its own refnum value.
The clones map back to the original, which is ref-counted and
used for all actual prim operations except release. When the last
clone is released, the original is disposed.


cglzaguilar:
Looks good to me, just one overall question. I noticed the code has several classes and methods with queue on it and I wonder if we need to support another kind of refnum, how reusable the code is?

spathiwa:
The typed refnum functionality comes from the base class, which is completely reusable.  The extra methods added have to do only with maintaining named queues, and queues are the only LabVIEW feature which supports such a concept.

PR #172, opened 2017-01-26
Support missing args in WaveformBuild

sanmut:
Add more Waveform test


spathiwa:
Looks good.

cglzaguilar:
Looks good to me too. But I would use the term unwired parameters rather than missing arguments.

spathiwa:
Actually, since Vireo is really its own language, missing (or optional) arguments is the better term.
Vireo doesn't have wires.

sanmut:
Is "wildcard argument" the Vireo terminology for this?

spathiwa:
It's called that at least in the macro defining a symbol name for the '*' string, but I think 'default argument' is used more.

PR #174, opened 2017-01-27
Http error code alignment

rajsite:
Aligned error codes and rename tests


cglzaguilar:
Already buddy the code with Milan in person. Looks good.

PR #179, opened 2017-02-05
Initial support for enums

spathiwa:
Example:
    e(dv(Enum16 ("abc def" %47hi jkl mno) 2) bar)
    e(dv(Enum32 (red orange yellow green blue purple)) foo)

Definition, string output and formatting with %d and %s work
Copy and Convert from int types to enums pin output to be in range.

Arithmetic prims currently do not pin, but having DFIR generate a
Copy from the enum to itself after arith ops would correct this in
the meantime.


source/core/GenericFunctions.cpp
```diff
@@ -21,12 +21,20 @@ namespace Vireo
...
     // If the source or dest are not aligned to aqSize bytes, return null.
     if (((uintptr_t) pSource % aqSize != 0) || ((uintptr_t) pDest % aqSize != 0))
         return null;
-
+    if (isEnum) {
+        switch (aqSize) {
+            case 1:     return "CopyEnum1";   break;
+            case 2:     return "CopyEnum2";   break;
+            case 4:     return "CopyEnum4";   break;
+            case 8:     return "CopyENum8";   break;
```
cglzaguilar: I'm curious why we don't have CopyEnum16 and CopyEnum32. I think a comment explaining why they're not necessary is helpful here, otherwise the code feels that is incomplete.

cglzaguilar: Typo on CopyENum8?

source/core/TDCodecVia.cpp
```diff
@@ -273,7 +274,9 @@ TypeRef TDViaParser::ParseType(TypeRef patternType)
     }
     
     Boolean bTypeFunction = _string.ComparePrefix('(');
-    if ((tt == TokenTraits_SymbolName) && (!bTypeFunction)) {
+    if (typeFunction.ComparePrefixCStr(tsEnumTypeToken) && (typeFunction.Length()==4 || isdigit(typeFunction.Begin()[4]))) {
```
cglzaguilar: Having magic numbers like 4, makes the code more difficult to read. Use constants in this case, with a self explanatory name, that way we avoid the need of adding a comment.

source/core/TDCodecVia.cpp
```diff
@@ -964,9 +1004,6 @@ void TDViaParser::ParseData(TypeRef type, void* pData)
             }
             }
             break;
-        case kEncoding_Enum:
```
cglzaguilar: I'm glad we finally removed this.

source/core/VirtualInstrument.cpp
```diff
@@ -621,6 +621,12 @@ void ClumpParseState::AddDataTargetArgument(SubString* argument, Boolean addType
         // StaticTypeAndData formal parameters get passed the type and pointer to the data.
         // they are fully polymorphic.
         InternalAddArg(null, ActualArgumentType());
+        static SubString etad("EnumTypeAndData");
```
cglzaguilar: etad? why not enumTypeAndData? As the code evolves, variable definitions tend to get separated from actual use, making harder to undertand the meaning of a variable.

source/include/TDCodecVia.h
```diff
@@ -223,6 +224,7 @@ void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], Strin
 #define tsParamBlockTypeToken   "p"  // Used for defining param blocks used by native functions.
 #define tsPointerTypeToken      "ptr"
 #define tsRefNumTypeToken       "refnum"
+#define tsEnumTypeToken         "Enum"
```
cglzaguilar: By looking at the other elements in the list does it make sense to have "enum", instead of "Enum"?

spathiwa:
Fixed the typo, added constant for enum string length and comments.

There is no CopyEnum16/32 because those are byte sizes and we do not have enums w/rep larger than UInt64.

The Copy functions next to it go up to 32 for efficiency reasons, not because those represent integer sizes.
I added comments to explain this.

‘Enum’ is upper case because it’s a user-visible type in Vireo, and all of our other type names (Int32, UInt32, etc.) are upper case.

The lower case names defined next to it (refnum, ptr), are NOT used in user Vireo code and only appear in internal type definitions.

They should probably be separated in the header file but I did not attempt this, as there are some whose usage I am unsure about.

Craig

PR #183, opened 2017-02-09
Fix Copy() of refnums (wasn't flagged as being flat).

spathiwa:
For now, use macro to define full inline error clust definition for prims
which take error clusters instead of using type names, because comparisons
of identical structured clusters with different top level names doesn't work correctly.


cglzaguilar:
Looks good.

sanmut:
spathiwa wrote "because comparisons of identical structured clusters with different top level names doesn't work correctly."

An arbitrary cluster with matching datatypes and even with the same names for the fields cannot be used in lieu of the typedef'ed Error cluster defined in the language and violation of this is an edit-time type system error in G.

spathiwa:
Yes, but this is only the case for the Error Cluster in NXG G, and not a rule for all clusters, correct?
Vireo does not need to have this prohibition, as the rule would be enforced before Vireo ever saw it.
(And CG LabVIEW does not have the prohibition either.)

PR #184, opened 2017-02-09
Add indexing of arrays in object paths.

spathiwa:
Now individual elements of arrays can be accessed via EggShell.
Also added Vireo prim GetSubElementFromPath which exposes this
funcitonality in via files, and test file GetObjPathArrayIndex.via


source/core/TypeAndDataManager.cpp
```diff
@@ -1676,20 +1676,38 @@ TypeRef ArrayType::GetSubElementAddressFromPath(SubString* path, void *start, vo
         SubString pathHead;
         SubString pathTail;
         path->SplitString(&pathHead, &pathTail, '.');
+        *end = null;
 
-        gPlatform.IO.Print(" Using array indexes in paths\n");
+        //gPlatform.IO.Print(" Using array indexes in paths\n");
```
cglzaguilar: Let's delete the comment. It doesn't add any value.

source/core/TypeAndDataManager.cpp
```diff
@@ -2876,6 +2894,22 @@ VIREO_FUNCTION_SIGNATURE4(TypeGetSubElementFromPath, StaticType, void, StringRef
   //  tm->PointerToSymbolPath(_Param(1), _Param(2), _Param(3));
     return _NextInstruction();
 }
+//------------------------------------------------------------
+//! Parse through a path from a starting point to target subelement, return subelement value if it matches expected output type.
+VIREO_FUNCTION_SIGNATURE5(GetSubElementFromPath, StaticType, void, StringRef, StaticType, void)
```
cglzaguilar: Do we need to worry about optional parameters?

cglzaguilar:
Other than that looks good to me. Waiting on Sankara to review the code.

source/core/TypeAndDataManager.cpp
spathiwa: No, why would we here?  There's a starting object (usually the VI), a string path, and a destination object.  Nothing's optional.

sanmut:
Changes in TypeAndDataManager.cpp seem to be risky enough that we should create a version of Vireo before and after this and validate with G tests also. I understand we are validating with Vireo unit tests. However, if we break something, it might be hard to debug which change broke. Ideally, it is a good idea to run the existing G tests also locally before submitting a risky change, IMHO.

test-it/GetObjPathArrayIndex.via
```diff
@@ -0,0 +1,61 @@
...
+        Println(w15 )
+
+        Copy("abc" w15)
+        GetSubElementFromPath(videf1 "w13.2" w15)
+        Println(w13)
+        Println(w15)
+        GetSubElementFromPath(videf1 "arrint.4" i1)
+        Println(arrint)
+        Println(i1)
+        GetSubElementFromPath(videf1 "arr2d.6" i1)
```
sanmut: Could you please add an extra argument to specifically say this is linearized index?  Just curious. Where would that be needed?
That way, we can add errors for index out of bounds.

source/core/TypeAndDataManager.cpp
cglzaguilar: So we can't pass '*' as a parameter to the function?

test-it/GetObjPathArrayIndex.via
```diff
@@ -0,0 +1,61 @@
...
+        e(.String w17)
+        e(.Type ty)
+    )
+    clump(1 
+        ArrayConcatenate( w11 c3 c2 )
+        ArrayConcatenate( w13 c8 c10 c7 c6 )
+        Println(w15 )
+
+        Copy("abc" w15)
+        GetSubElementFromPath(videf1 "w13.2" w15)
```
sanmut: Can we make the second argument as the actual variable name (not the string) so we can check this better? Can the indices be separate arguments as opposed to string? Should we really mix cluster fields and array indices in the same string? Seems risky for type checking and erroring out?

cglzaguilar:
Good point Sankara, but I think we should rely on the Vireo unit tests, to validate Vireo itself. I know there are gaps and we need to work to close them.

source/core/TypeAndDataManager.cpp
spathiwa: Optional parameters only make sense when there is a well-defined behavior without them.
The first two parameters are required to function, and if the only output is optional, the whole thing is a no-op.

test-it/GetObjPathArrayIndex.via
```diff
@@ -0,0 +1,61 @@
+// Autogenerated Vireo assembly file
+// Generated on Tue Oct 08 15:12:36 2013
+define(Point c(e(.Int32 x)e(.Int32 y)))
+//---------------------------------------------------
+//VI Definition: stringManipulation.vi
+define(videf1  dv(.VirtualInstrument (
```
spathiwa: Ok, I had one in there but overwrote it with a different test case.  It returns default data for out of bounds.

source/core/TypeAndDataManager.cpp
```diff
@@ -1676,20 +1676,36 @@ TypeRef ArrayType::GetSubElementAddressFromPath(SubString* path, void *start, vo
...
+                        subType = array->ElementType();
+                        *end = (AQBlock1*)array->BeginAtND(rank, dimIndex);
+                    }
+                } else {
+                    subType = array->ElementType();
+                    *end = (AQBlock1*)array->BeginAt(IntIndex(index));
+                }
+            }
+        }
+        if (*end && pathTail.Length()) {
```
spathiwa: Length can't be negative, although it's not actually an unsigned type.  That would be an assertion/invariant.

test-it/GetObjPathArrayIndex.via
```diff
@@ -0,0 +1,61 @@
...
+        e(dv(a(.Int32 * *) ((11 12 13 14 15) (21 22 23 24 25)) ) arr2d)
+        e(dv(.Int32 0) i1)
+        e(dv(Point) clust)
+        e(.String w17)
+        e(.Type ty)
+    )
+    clump(1 
+        ArrayConcatenate( w11 c3 c2 )
+        ArrayConcatenate( w13 c8 c10 c7 c6 )
+        Println(w15 )
```
spathiwa: I disagree.  Hard-coded line numbers make it hard to add new cases to the tests in the middle near other similar cases.  I think this sort of issue should be fixed in the test-runner infrastructure; it can easily provide differing line numbers or other context.  The test.js runner already does, even with colorized diffs.

test-it/GetObjPathArrayIndex.via
spathiwa: NO.  This prim is for the express purpose of testing the actual string-based implementation as used by EggShell.  Changing its API would defeat that purpose.  That's the way the object paths already used by EggShell work.  I just extended the syntax to add array subindexes, and added this prim only for testing that functionality as a Unit test.  

(The prim was added to the VIREO_INSTRUCTION_REFLECTION set of prims for that reason, because it deals with language features as strings, as do all the other prims in that section.  LV/DFIR would never emit this instruction directly.).

test-it/GetObjPathArrayIndex.via
spathiwa: You can tell a linearized index already.  For 1D arrays, it doesn't matter, they are equivalent.  For N-D arrays, if you have one index it's linearized, if you have multiple, it's not.
I could add an error out, but again, this is a testing prim for a feature to be used by EggShell.  The existing object path API EggShell uses ALREADY doesn't have an error mechanism (for bad cluster fields or known local names, for example).
All I did was extend it to work on array indexes.  Extending the object path mechanism further and adding error reporting would need to be discussed with the team.
EggShell will know the sizes ahead of time and will decide to use the linearized index or multidim index based on what turns out to be more efficient, and so it may not ever need error reporting because the field names and array indexes are precomputed.

source/core/TypeAndDataManager.cpp
spathiwa: Actually, the > 0 in the other case was because of a previous edit where it was temporarily > 1 in order to help debug an issue.  When I changed it back by hand I changed it to > 0 instead of removing the inequality.  I have reverted that edit in the other place now.

rajsite:
Can we add a JS test that verifies the new peek /poke behavior for arrays?

Can look at [errorcluster.test.js](https://github.com/ni/VireoSDK/blob/incoming/test-it/karma/publicapi/errorcluster.test.js) or [peekpoke.test.js](https://github.com/ni/VireoSDK/blob/incoming/test-it/karma/publicapi/peekpoke.test.js) as examples

rajsite:
We would still be relying on the EggShell_ReadValueString entry point correct? https://github.com/ni/VireoSDK/blob/incoming/source/core/CEntryPoints.cpp#L164

How will a caller know if we have indexed passed the end of the array? Will EggShell_ReadValueString return null if the path indexes outside the array?

spathiwa:
It would've return an empty string; I just tweaked my submission so that it will return null instead for out of range indexes.

PR #186, opened 2017-02-14
Make Queue prims tolerant of unwired refnum (even though it's required).

spathiwa:
Implement IsNotARefnum, IsEQ, and IsNE on queue refnums.


cglzaguilar:
@spathiwa, please add tests for IsEQQueueRefnum, IsNEQueueRefnum and IsNotARefnum when dealing with a refnum value. You can do this later.

PR #187, opened 2017-02-15
Fix queue name output parm of ReleaseQueue and GetQueueStatus

spathiwa:
And remainingElts output of ReleaseQueue.

### DIFF
--- a/source/core/Synchronization.cpp
+++ b/source/core/Synchronization.cpp
@@ -582,6 +582,7 @@ public:
     QueueRefNumType &RefNumManager() { return _QueueRefNumTypeStorage; }
     NamedRefNumMapType &NamedRefNumMap() { return _namedRefMap; }
     RefNum NewRefnumAlias(RefNum refnum);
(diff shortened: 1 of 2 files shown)


source/core/Synchronization.cpp
```diff
@@ -751,22 +759,65 @@ VIREO_FUNCTION_SIGNATURE6(QueueRef_Obtain, RefNumValType*, Int32, StringRef, Boo
...
+                ++it;
+            }
+            if (it != ite)
+                (*stringRef)->Type()->CopyData(&it->first,stringRef);
+            else
+                (*stringRef)->Resize1D(0);
+        }
+        else
+            (*stringRef)->Resize1D(0);
+    }
```
cglzaguilar: Shouldn't the else case for if (stringRef) return an empty string, like the other cases.

cglzaguilar: Never mind, you're checking for null.

PR #188, opened 2017-02-15
withCredentials to false to enable Anon. Requests

rajsite:
Anonymous Requests require withCredentials to be false and is currently our most frequent use case in testing.

However, this will prevent making HTTP requests to Cross-Origin servers using Cookies or the HTTP Auth Header for authentication. So we need a way to expose configuration of this property to users.

It should not have an effect making requests to same origin servers.


cglzaguilar:
Reviewed this change with Milan.

PR #189, opened 2017-02-15
Bug fixes for HTTP and Queue primitives

cglzaguilar:


cglzaguilar:
Bug fixes for HTTP and Queue primitives. Removed testing of 8bit resolutions timers. They wrap around and case false failures.

PR #192, opened 2017-02-17
Add MatchPattern Vireo prim, modeled after LV Match Pattern prim.

spathiwa:
MatchPattern(i(String string) i(String pattern) i(Int32 offset)
   o(String beforeStr) o(String matchStr) o(String afterStr) o(Int32 offsetPastMatch)

Match Pattern works on a very simple subset of regular expressions and
does not support | alternation or capture groups or modern PCRE
features.
The full Match Regular Expression prim is not yet supported.

Added minimal Match Pattern unit tests, but it needs to be further
fleshed out.


cglzaguilar:
Looks good to me.

PR #193, opened 2017-02-20
Add CallChain prim.

spathiwa:
Usgage:  CallChain(o(Array(String *)))

Returns an array of caller VIs, element zero being the most recent on
the 'stack'.  It supports reentrant VIs but does not yet support adding
":#" to the end, because we don't yet support clone pools.

Added a SubString field to VirtualInstrument so it can know its own
name.  (Previously this was only known to the type system via 'define').


source/CommandLine/main.cpp
```diff
@@ -46,36 +46,38 @@ int VIREO_MAIN(int argc, const char * argv[])
             if (strcmp(argv[arg], "-d") == 0)
                 gShells._pRootShell->DumpPrimitiveDictionary();
 			gShells._pUserShell = TypeManager::New(gShells._pRootShell);
```
sanmut: Indentation seems to be messed up. Could you please fix it?

source/CommandLine/main.cpp
```diff
@@ -46,36 +46,38 @@ int VIREO_MAIN(int argc, const char * argv[])
...
-				emscripten_set_main_loop(RunExec, 40, null);
-#else
-				while (gShells._keepRunning) {
-					RunExec(); // deletes TypeManagers on exit
-				}
-#endif
-            } else {
-                gShells._pUserShell->Delete();
-                gShells._pRootShell->Delete();
+            {
```
sanmut: Why do we need this curly brace here?

source/CommandLine/main.cpp
```diff
@@ -46,36 +46,38 @@ int VIREO_MAIN(int argc, const char * argv[])
...
+                TypeManagerScope scope(gShells._pUserShell);
+                STACK_VAR(String, buffer);
+                fileName.AliasAssignCStr(argv[arg]);
+                if (fileName.Length() && argv[arg][0] != '-') {
+                    {
```
sanmut: Why do we need this curly brace here?

source/CommandLine/main.cpp
```diff
@@ -46,36 +46,38 @@ int VIREO_MAIN(int argc, const char * argv[])
...
+                        gPlatform.IO.ReadFile(&fileName, buffer.Value);
```
sanmut: Indentation seems to be messed up and hard to eyeball and follow the flow. I suspect it is due to mixed use of tabs and spaces. Please fix it.

source/include/VirtualInstrument.h
```diff
@@ -140,11 +145,10 @@ class VIClump : public FunctionClump
     VirtualInstrument*  TopVI()      {
         VIClump *caller = this;
         do {
-            while (caller->_waitingClumps)
```
sanmut: Was the old logic of using _waitingClumps incorrect? That you ran into while implementing callchain? Or, is it some optimization?

source/core/ExecutionContext.cpp
```diff
@@ -198,6 +198,41 @@ VIREO_FUNCTION_SIGNATURE1(Branch, InstructionCore)
...
+            ++count;
+        } while (caller);
+
+        _Param(0)->Resize1D(count);
+
+        count = 0;
+        vi = runningQueueElt->OwningVI();
+        do {
+            SubString s = vi->VIName();
+            _Param(0)->At(count)->CopyFromSubString(&s);
```
sanmut: Is it a good idea to make sure "count" is less than what we have allocated before from the previous while loop to avoid any potential access violation errors.

source/CommandLine/main.cpp
cglzaguilar: Consider adding curly braces for the previous if, it is not clear what statements belong to it.

source/include/VirtualInstrument.h
```diff
@@ -140,11 +145,10 @@ class VIClump : public FunctionClump
...
-                caller = caller->_waitingClumps;
-            while (caller->_caller)
+            if (caller->_caller)
```
cglzaguilar: Add curly braces for the if code block, so the intent is clear and we don't rely on indentation along to convey the meaning.

source/CommandLine/main.cpp
spathiwa: I added a comment explaining the use of a set of curly braces that are intended only to control scope and when destructors are run.

source/CommandLine/main.cpp
spathiwa: I will delete this set of braces that used to be used for that same purpose as the braces mentioned above, but the stack object was moved and so these were no longer necessary.

source/include/VirtualInstrument.h
spathiwa: The old logic was wrong because it did not handle the Trigger'd clumps that weren't actively being waited on.

source/include/VirtualInstrument.h
spathiwa: I think it's obvious from context this is just a NULL-test guard here.

source/core/ExecutionContext.cpp
spathiwa: This would be an assertion-level change, because the two loops have identical control flow and there's no way count can go out of bounds the second time.
I'm hesitant to add a new variable just to be able to add the assertion

spathiwa: I added a comment emphasizing the importance that the two loops match.

source/include/VirtualInstrument.h
sanmut: I see. Thanks.,

PR #194, opened 2017-02-21
Updated Enum code to return numeric values when generating JSON

cglzaguilar:
Also updated EnMakefile-AAL, to look for Analysis Library files.

### DIFF
--- a/make-it/EmMakefile-AAL
+++ b/make-it/EmMakefile-AAL
@@ -21,7 +21,7 @@ endif
 ifneq (,$(wildcard $(VIREO_SDK_DIR)/private/*))
   ANALYSIS_WRAPPERS_SOURCE_DIRECTORY=$(VIREO_SDK_DIR)/private
 else
-  ANALYSIS_WRAPPERS_SOURCE_DIRECTORY=../vi.lib/Analysis
+  ANALYSIS_WRAPPERS_SOURCE_DIRECTORY=../../vi.lib/Analysis
 endif
 
 ifeq (,$(wildcard $(ANALYSIS_WRAPPERS_SOURCE_DIRECTORY)/AnalysisWrapper*.cpp))
@@ -32,7 +32,7 @@ $(OBJS)/NoAnalysisLibrary.bc: | no-analysis
 	rm $(OBJS)/NoAnalysis.cpp
 
 no-analysis::
-	$(info Note: Analysis library not found in $(VIREO_SDK_DIR)/private or ../vi.lib/Analysis; omitting from build.)
+	$(info Note: Analysis library not found in $(ANALYSIS_ROOT_DIRECTORY) or Analysis Wrappers not found in $(ANALYSIS_WRAPPERS_SOURCE_DIRECTORY); omitting from build.)
 else
 
 ifeq (,$(AAL_ONLY))
--- a/test-it/results/EnumBasic.vtr
+++ b/test-it/results/EnumBasic.vtr
@@ -1,13 +1,13 @@
-2 jkl
-3 mno
+2 2
+3 3
 1
-Ghi
-c0 mno
-c1 abc def
-red
-orange
-yellow
-green
-blue
-purple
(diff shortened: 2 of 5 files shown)


test-it/results/EnumBasic.vtr
```diff
@@ -1,13 +1,13 @@
-2 jkl
```
rajsite: Looking at the EnumBasic.via file we do want the result of StringFormat primitive to output the enum string and not a number.

However for readJSON we want the number value to be returned when we read the terminal.

make-it/EmMakefile-AAL
```diff
@@ -32,7 +32,7 @@ $(OBJS)/NoAnalysisLibrary.bc: | no-analysis
 	rm $(OBJS)/NoAnalysis.cpp
 
 no-analysis::
-	$(info Note: Analysis library not found in $(VIREO_SDK_DIR)/private or ../vi.lib/Analysis; omitting from build.)
+	$(info Note: Analysis library not found in $(ANALYSIS_ROOT_DIRECTORY) or Analysis Wrappers not found in $(ANALYSIS_WRAPPERS_SOURCE_DIRECTORY); omitting from build.)
```
sanmut: I would do these changes in a separate commit as this seems to be unrelated. You can still do one pull request for  multiple commits.

cglzaguilar:
Closing pull request to submit changes to make file independently.

rajsite:
@cglzaguilar Was the StringFormat behavior fixed?

JackDunaway:
What's the motivation for this change? Since the string representation carries more information than the (usually arbitrary) number it represents, it seems a better fit for JSON representation. A number is likely to be more performant, yet, we prefer to preserve semantic information over raw speed when choosing JSON (or other human-readable serializations). It's more robust against mutating enums also -- normal things like adding/removing/reordering items -- where names are likely to stay the same. Said another way -- the deserializer is far more likely to make the correct choice when mapping to the correct underlying number. This also presents a natural means for the serializer and deserializer to not necessarily map to the same underlying numbers -- their contract agrees upon the *symbol*, not necessarily the underlying number it represents.

JackDunaway:
Real-world stories:

**The _My Life Sucks_ Example**

Application writes its configuration to a file, including an enum field **serialized as a number**. I update the application, adding and removing some values from that enum. The code that I must write to properly deserialize the enum is error-prone and frustrating to maintain, since it's just maps of numbers to numbers. (Plus -- I need to encode extra information into the configuration file, such as `ConfigFileVersion`, and properly bump that version in tandem as well)

**The _My Life Is Not So Terrible_ Example**

Application writes its configuration to a file, including an enum field **serialized as a string**. I update the application, adding and removing some values. The code that I must write to properly deserialize the enum checks whether the string maps to the current enum or not; the happy case is simple (enum hasn't changed; items reordered; items added), and the unhappy case (when the symbol no longer exists) has some semantic-looking mutation code making the proper re-mapping to the appropriate default value.

---

The same general cases apply for inter-application comms, but could be even messier when it is separate codebases exchanging information.

cglzaguilar:
@rajsite I didn't apply the change, @JackDunaway the reason was for compatibility with a product we are developing that consumes Vireo.

JackDunaway:
> compatibility with a product we are developing that consumes Vireo

Is it worthwhile for that product to reconsider this case? Or, has this ship sailed?

cglzaguilar:
@JackDunaway, yes we are reconsidering this case.

JackDunaway:
Thank you! As a precedent of when a popular language chose this same path, it turned into a long tail of devs looking for the "better" option: http://stackoverflow.com/questions/2441290/json-serialization-of-enum-as-string

I reckon it's easier for the Vireo team right now to just change Vireo in order to conform to suboptimal requirements/protocols/syntaxes, but I sincerely appreciate your work and opportunities to reverse the dependency vector and feed better design back the other way.

PR #196, opened 2017-02-22
Temporary change for JSON encoding of enums. Return numeric value of …

cglzaguilar:
…enum instead of name. We will revisit this once the the system dependent on Vireo is fixed

### DIFF
--- a/source/core/TDCodecVia.cpp
+++ b/source/core/TDCodecVia.cpp
@@ -1738,10 +1738,10 @@ private:
 char TDViaFormatter::LocaleDefaultDecimalSeperator = '.';
 
 // the format const used in Formatter and Parser
-ViaFormatChars TDViaFormatter::formatVIA =  {"VIA",  '(',')','(',')',' ','\'', kViaFormat_NoFieldNames};
-ViaFormatChars TDViaFormatter::formatJSON = {"JSON", '[',']','{','}',',','\"', ViaFormat(kViaFormat_QuotedFieldNames | kViaFormat_SuppressInfNaN) };
-ViaFormatChars TDViaFormatter::formatJSONLVExt = {"JSON", '[',']','{','}',',','\"', ViaFormat(kViaFormat_QuotedFieldNames | kViaFormat_UseLongNameInfNaN) };
-ViaFormatChars TDViaFormatter::formatC =    {"C",    '{','}','{','}',',','\"', kViaFormat_NoFieldNames};
+ViaFormatChars TDViaFormatter::formatVIA =       { kVIAEncoding,  '(',')','(',')',' ','\'', kViaFormat_NoFieldNames};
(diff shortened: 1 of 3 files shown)


JackDunaway:
> efficiency is more important when communicating between WebVI and Vireos in the context of a single, already-compiled VI where the encoding cannot change

I think I understand what you're saying -- but would you check my understanding?

A "WebVI" started as G code, but was compiled to JS. This output JS from the developer's point of view is treated more like object code than source code (even though this representation is still "source code" in one sense).

Likewise, that "object code" needs to invoke and be invoked by Vireo acting as a runtime, which is also compiled to JS. It's this data transfer you're talking about, yes?

So then, what about when compiling application and Vireo runtime to WASM? Are interfaces then going to be more of a binary calling convention, instead of JSON, for efficiency?

> like the LvExt option for Inf/NaN support

This is a motivating factor for me caring about this issue about enums -- the fact that this option spits out RFC 7159 non-compliant code, it's caused a long tail of unexpected interoperability problems over the years. The root of the problem is that IEEE754 and JSON don't mix well anyway -- which is not our problem to solve, but we have to do something, and we're forced into choosing from a few suboptimal options.

But then, the sub-suboptimal decision to produce `{"val":NaN}` -- as opposed to `{"val":"NaN"}` -- or perhaps even `{"val":"0x7FFFFFFFFFFFFFFF"}` -- or any other encoding _that is at least still valid JSON_ -- I feel has a mistake that's been costly to live with. (To be fair, JS is one language chill parsing `{"val":NaN}` -- but JS is arguably a bit too chill on things like this.)

rajsite:
@spathiwa 
Yes, it seems reasonable to be able to distinguish between the FlattenToJSON primitive behaviors and the vireo.eggShell.readJSON behavior as vireo.eggShell.readJSON is intended to encode values in a format that generates valid JSON as output.

Which means the Vireo JSON parser / serializer should be able to handle the following formats:
FlattenToJSON (used on diagram)
FlattenToJSON LvEXT (used on diagram)
vireo.eggShell.readJSON (used by JS to communicate with the Vireo runtime)

We must be able to distinguish this uses to some extent already as readJSON for IEE754 Infinity in Vireo returns the valid JSON string `"Infinity"` and complex as `{"real": 123, "imaginary": 123}`, etc. Curious if you know how that is being distinguished right now and yes it's possible adding an additional option may simplify those paths.

@JackDunaway 
WASM shouldn't have any direct implication on how we approach communication / the apis between Vireo and JS. Effectively [WASM MVP](http://webassembly.org/docs/mvp/) should just encode the current vireo ouput in a binary format instead of a [text format](https://unpkg.com/vireo@5.3.1/dist/vireo.js) and run some things a bit faster (particularly 64-bit integer computations).

Using a JSON representation for values returned by the JS API for Vireo is definitely a very natural format for the JS environment and has got us a lot of lift so far.

PR #197, opened 2017-02-23
Make VI own a copy of its name string instead of using a substring,

spathiwa:
(The source string used to initialize the VI name could be volatile.)

### DIFF
--- a/source/include/VirtualInstrument.h
+++ b/source/include/VirtualInstrument.h
@@ -73,10 +73,10 @@ public:
             s.AliasAssignLen(_viName, IntIndex(strlen(ConstCStr(_viName))));
         return s;
     }
-    void SetVIName(const SubString *s)   {
-        IntIndex len = s->Length();
-        _viName = (Utf8Char*)malloc(len+1);
-        memcpy(_viName, s->Begin(), len);
+    void SetVIName(const SubString &s)   {
+        IntIndex len = s.Length();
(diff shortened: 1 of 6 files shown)


source/include/VirtualInstrument.h
```diff
@@ -62,12 +62,23 @@ public :
...
     TypedArray1D<VIClump>* Clumps()     { return _clumps; }
-    SubString VIName()                  { return _viName; }
-    void SetVIName(SubString s)         { _viName = s; }
+    SubString VIName()                  {
+        SubString s;
+        if (_viName)
+            s.AliasAssignLen(_viName, IntIndex(strlen(ConstCStr(_viName))));
+        return s;
+    }
+    void SetVIName(const SubString *s)   {
```
sanmut: Since SetVIName is not modifying "s", I would make this parameter a const SubString reference instead of pointer to follow the convention. It will also be clear at the call site that the argument being passed is not modified by the caller.

source/include/VirtualInstrument.h
```diff
@@ -62,12 +62,23 @@ public :
...
+        IntIndex len = s->Length();
+        _viName = (Utf8Char*)malloc(len+1);
```
sanmut: May be, you meant to do "new" since you are calling "delete" in the destructor ~VirtualInstrument().

PR #199, opened 2017-02-24
Code cleanup for Scan primitive

cglzaguilar:
Even though the diff, shows changes on the array section, I just indented the code, not changes on the array section were made.

### DIFF
--- a/source/core/NumericString.cpp
+++ b/source/core/NumericString.cpp
@@ -1232,6 +1232,124 @@ Boolean BelongtoCharSet(SubString* charSet, Utf8Char candidate) {
     }
     return false;
 }
+
+//----------------------------------------------------------------------------------------------------
+IntMax ScanIntBaseValues(char formatChar, char* beginPointer, char** endPointer)
+{
+    IntMax intValue = 0;
+    switch (formatChar) {
+        case 'x': {
+                intValue = strtoull(beginPointer, endPointer, 16);
+            }
+            break;
+        case 'd':
+        case 'u': {
+                intValue = strtoull(beginPointer, endPointer, 10);
+            }
+            break;
+        case 'b':
+        case 'B': {
+                intValue = strtoull(beginPointer, endPointer, 2);
+            }
+            break;
+        case 'o': {
+                intValue = strtoull(beginPointer, endPointer, 8);
+            }
+            break;
+    }
+    return intValue;
+}
+
+//----------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


source/core/NumericString.cpp
```diff
@@ -1232,6 +1232,124 @@ Boolean BelongtoCharSet(SubString* charSet, Utf8Char candidate) {
...
+                    char siPrefixesTable[] = { 'y', 'z', 'a', 'f', 'p', 'n','u', 'm', '0', 'k', 'M','G', 'T', 'P','E', 'Z', 'Y' };
+                    char prefix = **endPointer;
+                    int i = 0;
+                    for (; i < sizeof(siPrefixesTable); ++i)
+                        if (siPrefixesTable[i] == prefix)
+                            break;
+                    if (i < sizeof(siPrefixesTable)) {
+                        ++beginPointer;
+                        doubleValue *= pow(10.0, (i - 8) * 3);
+                    }
```
JackDunaway: This might be too pedantic? But sizeof a char might not always eval to 1, and the scope of `i` could be reduced to help brevity with that additional syntax. (Unless -- is C89 compatibility for variable declarations restrict to top of blocks a design goal?)

```C
for (int i = 0; i < sizeof(siPrefixesTable) / sizeof(siPrefixesTable[0]); ++i) {
  if (siPrefixesTable[i] == prefix) {
    ++beginPointer;
    doubleValue *= pow(10.0, (i - 8) * 3);
    break;
  }
}
```

spathiwa: Yes, that's too pedantic.   sizeof(char) will always be 1.  

Moving the declaration of i would invalidate the test of it after the loop to detect if the loop completed without finding the prefix so a default value can be provided.

JackDunaway: > Moving the declaration of i would invalidate the test of it after the loop to detect if the loop completed without finding the prefix so a default value can be provided

I think this is still handled properly by moving that test into the loop, since the default is to skip scaling, but ...

> Yes, that's too pedantic. sizeof(char) will always be 1.

... yes, you're right -- bits may differ, but sizeof(char) is 1. The more likely scenario is the table changing to another type to hold a different encoding, but again ... unlikely, still too pedantic.

sanmut: If we are concerned, a cleaner approach would be to declare a macro like the following in a header file and use it in the code:
#define ARRAY_SIZE(a) ( (sizeof(a) / sizeof((a)[0])) )

PR #200, opened 2017-02-27
Added ability to ScanString function to process named enum values

cglzaguilar:


source/core/NumericString.cpp
```diff
@@ -1351,6 +1351,31 @@ void DoubleScanString(StaticTypeAndData* argument, TypeRef argumentType, TempSta
 }
 
 //----------------------------------------------------------------------------------------------------
+Boolean EnumScanString(SubString* in, StaticTypeAndData* argument, TypeRef argumentType, char formatChar, char* beginPointer, char** endPointer)
+{
+    Boolean found = false;
+    if (formatChar == 's') {
+        for (IntIndex i = 0; i < argumentType->GetEnumItemCount(); i++) {
+            SubString itemName = argumentType->GetEnumItemName(i)->MakeSubStringAlias();
+            if (in->ComparePrefix(itemName.Begin(), itemName.Length()))
```
spathiwa: This will incorrectly take the first match instead of the longest match if the enum strings themselves have any common prefixes.  For example, scanning a string with an Enum16 ("red" "redorange" "redyellow") would match "red" first and so never match the other two.

spathiwa:
I think the new code will behave weirdly if one of the enums strings is empty (which is technically allowed).  LabVIEW desktop would never match such a string and just ignore it. Your code would match it if it didn't match anything else, but always return the first enum element (since it wouldn't overwrite longestMatchIndex since longestMatchLength would already be 0).

source/core/NumericString.cpp
```diff
@@ -1377,9 +1412,12 @@ Boolean TypedScanString(SubString* inputString, IntIndex* endToken, const Format
     
     switch (argumentType->BitEncoding()) {
         case kEncoding_UInt: 
-        case kEncoding_Enum:
             IntScanString(argument, argumentType, formatOptions->FormatChar, inpBegin, &endPointer);
             break;
+        case kEncoding_Enum:
+            if (!EnumScanString(&in, argument, argumentType, formatOptions->FormatChar, inpBegin, &endPointer))
```
sanmut: I wonder if EnumScanString() implementation should handle only the "if case". And call EnumScanString() if (formatChar == 's') and else call IntScanString() from here at this call site. I am not sure whether that would be the right thing to do?

cglzaguilar: Interesting, in general I prefer to have a single method that takes care of all the conditions, but I can see your point here, given that it will shown more clearly what is going on in this case structure. I'll consider changing the code in a future commit.

PR #201, opened 2017-02-27
Add EggShell_GetArrayLength

rajsite:


rajsite:
Renamed to Get1DArrayLength and fixed whitespace

@cglzaguilar Can you do a patch version bump for this so we can import?

PR #205, opened 2017-02-28
Fixed tautological-constant-out-of-range-compare warnings on compilation

cglzaguilar:


source/core/Math.cpp
```diff
@@ -278,6 +276,16 @@ using namespace std;
...
 		}
 
+#define DECLARE_VIREO_COMPARISON_PRIMITIVES(TYPE) \
+    DECLARE_VIREO_PRIMITIVE2( IsLT0##TYPE, TYPE, Boolean, (_Param(1) = _Param(0) <  0) ) \
+    DECLARE_VIREO_PRIMITIVE2( IsGE0##TYPE, TYPE, Boolean, (_Param(1) = _Param(0) >= 0) ) \
+    DECLARE_VIREO_COMPARISON_PRIMITIVES_BASE(TYPE)
+
+#define DECLARE_VIREO_COMPARISON_PRIMITIVES_BOOLEAN() \
+    DECLARE_VIREO_PRIMITIVE2( IsLT0Boolean, Boolean, Boolean, (_Param(1) = false) ) \
+    DECLARE_VIREO_PRIMITIVE2( IsGE0Boolean, Boolean, Boolean, (_Param(1) = true) ) \
```
sanmut: Did you mean to say "==" instead of "=" in these two lines?

source/core/Math.cpp
```diff
@@ -278,6 +276,16 @@ using namespace std;
 		return _NextInstruction();													\
 		}
 
+#define DECLARE_VIREO_COMPARISON_PRIMITIVES(TYPE) \
+    DECLARE_VIREO_PRIMITIVE2( IsLT0##TYPE, TYPE, Boolean, (_Param(1) = _Param(0) <  0) ) \
+    DECLARE_VIREO_PRIMITIVE2( IsGE0##TYPE, TYPE, Boolean, (_Param(1) = _Param(0) >= 0) ) \
```
sanmut: Why are we adding IsLT0 and IsGE0 but not IsLE0 and IsGT0? Do they also throw the same compiler warnings? If so, fixing partial compiler warnings for the same lines does not seem to be a good idea.

source/core/Math.cpp
sanmut: LabVIEW throws type errors when you wire a boolean to IsLT0, IsLE0, IsGT0, IsGE0. If we want to fix the warnings and refactor some code, I would rather do it such that Vireo throws a compiler error for these cases on boolean input. Thoughts?

source/core/Math.cpp
spathiwa: No, the IsLE0 and IsGT0 don't generate warnings.  The warnings are because Boolean can only be 0 or 1, and saying "if (b < 0)" when b can only be 0 or 1, or "if (b >= 0)", doesn't make sense.  
However, if "(b <= 0)" does make sense, and is sometimes true, sometimes false.

source/core/Math.cpp
spathiwa: He meant '='.
I don't have strong opinions on your second comment.  Vireo is not strictly G, and it's not strictly important if it allows things G wouldn't, provided DFIR filters such cases out before Vireo ever sees them when the source is a G VI.

source/core/Math.cpp
sanmut: ah, ok, I didn't understand the warnings correctly. This is fine then.

PR #206, opened 2017-03-01
Fix EggShell ReadValueString JSON to use quoted representations of "NaN", etc.

spathiwa:


source/core/TDCodecVia.cpp
```diff
@@ -1834,33 +1835,33 @@ void TDViaFormatter::FormatIEEE754(TypeRef type, void* pData)
...
-        } else {
-            pBuff = "nan";
-            len = 3;
-        }
-#else
         if (!suppressInfNaN) {
-            pBuff = _options._fmt.LongNameInfNaN() ?  "NaN" : "nan";
+            pBuff = _options._fmt.LongNameInfNaN() ?  "\"NaN\"" : "\"nan\"";
             len = 3;
+            if (quotedInfNaN)
```
cglzaguilar: I wonder if there is a way to produce cleaner code. Is it costly to add the quotes to pBuff later? Increasing len in one section of the if and then increasing the pointer to Buff in the else, required more attention to understand.

cglzaguilar: Just in my mind the code is more complicated that it needs to be.

spathiwa: I'd be glad to add comments, but I'd rather not make it less efficient for a slight improvement in readability.

cglzaguilar: Sound good to me, please add comments.

spathiwa: Done.

PR #207, opened 2017-03-01
Optimize EggShell_ReadValueString

spathiwa:
The FormatData call is inefficient for arrays by resizing the output
string one sub-element at a time.  Estimating the string size ahead of time
and preallocating the output cuts the time in half.

### DIFF
--- a/source/include/DataTypes.h
+++ b/source/include/DataTypes.h
@@ -143,10 +143,10 @@ typedef Int64   AQBlock8;
 //------------------------------------------------------------
 #ifdef VIREO_USING_ASSERTS
     #ifdef VIREO_MICRO
-        #define VIREO_ASSERT( _TEST_ ) VireoAssert_Hidden( _TEST_, __FILE__, __LINE__ );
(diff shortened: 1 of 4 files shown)


source/include/DataTypes.h
```diff
@@ -143,10 +143,10 @@ typedef Int64   AQBlock8;
 //------------------------------------------------------------
 #ifdef VIREO_USING_ASSERTS
     #ifdef VIREO_MICRO
-        #define VIREO_ASSERT( _TEST_ ) VireoAssert_Hidden( _TEST_, __FILE__, __LINE__ );
+        #define VIREO_ASSERT( _TEST_ ) if (!(_TEST_)) VireoAssert_Hidden( _TEST_, __FILE__, __LINE__ ); else;
```
cglzaguilar: Please elaborate why this change was necessary for my understanding, it seems both in the if statements do nothing.

spathiwa: Profiling indicated it was spending a lot of time in this function call, because the assert was called from a lot of short commonly-used functons, like Array Rank() and Length().
Adding the if test on the outside made the common case avoid the function call.
The 'else;' bit is to ensure that if VIREO_ASSERT is itself used inside an if, it doesn't cause a mismatched else clause in the caller.

PR #208, opened 2017-03-01
Store VI names in 'decoded' form in the type dictionary.

spathiwa:
Allow either %-code encoded or quoted strings for VI names in define
operations, enqueue arguments, or subVI invocations.

"My Test.vi" and My%20Test%2Evi are equivalent.


source/core/StringUtilities.cpp
```diff
@@ -1040,6 +1040,44 @@ IntIndex SubString::FindFirstMatch(SubString* searchString, IntIndex offset, Boo
...
+        _decodedStr = new Utf8Char[len+1];
+    else
+        _decodedStr = _buffer;
+    if (decode) {
+        SubString ss = s;
+        Utf8Char c;
+        Int32 value;
+        Utf8Char *pDest = _decodedStr;
+        while(ss.ReadRawChar(&c)) {
+            if (c == '%' && ss.ReadHex2(&value))
```
sanmut: Do we have to handle any special characters here? for instance, decoded double-quote or any other special character?

spathiwa: I did leave out one case, '+' maps to space, which I'll add.  Every other special character encodes to %xx.

test-it/CallChain.via
```diff
@@ -6,7 +6,7 @@ define(c0 dv(.Int32 5) )
 define(c1 dv(.Int32 1) )
 
 //---------------------------------------------------
-define(videf2 dv(.VirtualInstrument (
+define(videf2%20foo dv(.VirtualInstrument (
```
spathiwa: Yes, you can use double quotes here as well.
I'll add a couple of these special characters to the test.  There's no reason to test all of them because I only added decoding, not encoding, and decoding is uniform for all characters.

source/core/StringUtilities.cpp
sanmut: cool, thanks.

PR #211, opened 2017-03-01
Support for StringPickLine

sanmut:


source/core/String.cpp
```diff
@@ -784,6 +784,39 @@ VIREO_FUNCTION_SIGNATURE3(BranchIfGEString, InstructionCore, StringRef, StringRe
     }
 }
 
+//------------------------------------------------------------
+VIREO_FUNCTION_SIGNATURE4(StringPickLine, StringRef, StringRef, Int32, StringRef)
+{
+   SubString initString = _Param(0)->MakeSubStringAlias();
```
spathiwa: InitString should be an optional parameter so this might need a null check.
Also add a case to the test for init string = *

sanmut: Good catch, thanks. Both done.

PR #212, opened 2017-03-02
Fix QueueRefNumNamed intermittent test and GetObjPath EggShell VI name lookup bug

spathiwa:


source/core/Synchronization.cpp
```diff
@@ -607,19 +608,23 @@ RefNum QueueRefNumManager::LookupAlias(RefNum refnum) {
     }
     return refnum;
 }
-bool QueueRefNumManager::DisposeAlias(RefNum refnum)
+// If refnum is an alias, delete its refcount and if it reaches 1, dispose it. Otionally return its name.
```
cglzaguilar: Fix typo.

PR #215, opened 2017-03-08
Add CurrentBrowserFPS Function

rajsite:


rajsite:
@spathiwa Do we already have an unsupported on this platform type of error code?
If not suggestions of what it should be and where to put it?

spathiwa:
@rajsite The LabVIEW error code for "not supported on this platform" is 53 (mgNotSupportedErr). I don't think it exists in Vireo yet.

spathiwa:
There so far isn't a place for universal LabVIEW-specific error codes.  There is NIError in DataType,h, which are Vireo-specific, but the ones intended to be compatible with LabVIEW error codes are defined locally near the definition of the functions/prims which use/return them.   
We should probably add at least the common ones in a central header file, like kLVNoErr = 0, kLVArgErr = 1,  kLVOutOfMemoryErr=2, kLVNotSupportedErr=53 }

spathiwa:
I would suggest Datatype.h, below the definition of NIError.  We should probably audit the code and make sure those error codes are not being exposed to LabVIEW prim instructions, or merge/renumber NIError to match LabVIEW.

PR #216, opened 2017-03-09
Add complex polymorphic prims:

spathiwa:
Polar
ComplexToPolar
PolarToReOrIm
ReOrImToPolar
ComplexToReOrIm
ReOrImToComplex

Add complex support for ScanString.
Add more basic functionality tests to ComplexMathFunctions.via test.
(This is still kind of bare bones, more testing will come from importing via
files from complex gvi tests.)

Modify binop generic funciton emission to support prims with two outputs
by overloading the slot for the 'accumulator'. No 2-output prims have
an aggregates mode, so this was straight-forward.
(Swapped the snippet and accumulator fields in the instruction to make this easier.)


source/core/GenericFunctions.cpp
```diff
@@ -228,6 +233,10 @@ InstructionCore* EmitGenericBinOpInstruction(ClumpParseState* pInstructionBuilde
     if (savedOperation.CompareCStr("Split") || savedOperation.CompareCStr("Join")) {  // Split and Join are uniquely identified by source type rather than dest type
         goalType = sourceXType;
     }
+    if (argCount == 4 &&
```
cglzaguilar: Consider moving validation code early in the method. It seems we are doing unnecessary calculations if we are going to exit early. Also I was wondering about the "pInstructionBuilder->_argTypes[2]->BitEncoding()!=destType->BitEncoding()" check, since destType is actually pInstructionBuilder->_argTypes[2], but you fixed this is a later change.

test-it/results/ComplexMathFunctions.vtr
```diff
@@ -1,9 +1,12 @@
 ReOrImToComplex 2.000000 0.500000 => (2 0.5)
```
cglzaguilar: When there is an issue with tests I found the output to be more difficult to interpret than it needs to be. Please add a unique identifier at the beginning of each line, like 000, 001 and so forth, that way it is also easier to talk about a problem with somebody else.

Thanks to Sankara for previously suggesting this improvement in our tests output.

source/core/GenericFunctions.cpp
spathiwa: This would be optimizing the error case. A correct via program must make this test regardless.  Also, it is a via compile tom check, not a run time one.

test-it/results/ComplexMathFunctions.vtr
spathiwa: There are already unique strings at the beginning of each line identifying the prim and either an a or c suffix differentiating array or cluster cases. 
Adding numbers to each test case I think is a bad idea in the tests themselves because it complicates adding cases in the middle near other similar tests and is hard to maintain. 
However, along the same lines, I think the test DRIVER should output line numbers. That would solve this problem and let us update all tests easily at once.

source/core/Math.cpp
```diff
@@ -847,6 +847,14 @@ DECLARE_VIREO_PRIMITIVE2( Log2ComplexSingle, ComplexSingle, ComplexSingle, (_Par
 DECLARE_VIREO_PRIMITIVE2( ExpComplexSingle, ComplexSingle, ComplexSingle, (_Param(1) = exp(_Param(0)) ) )
 DECLARE_VIREO_PRIMITIVE3( PowComplexSingle, ComplexSingle, ComplexSingle, ComplexSingle, (_Param(2) = pow(_Param(0), _Param(1)) ) )
 DECLARE_VIREO_PRIMITIVE3( PolarComplexSingle, Single, Single, ComplexSingle, (_Param(2) = polar(_Param(0), _Param(1)) ) )
+DECLARE_VIREO_PRIMITIVE2( IsEQ0ComplexSingle, ComplexSingle, Boolean, (_Param(1) = _Param(0) == 0.0f ) )
+DECLARE_VIREO_PRIMITIVE2( IsNE0ComplexSingle, ComplexSingle, Boolean, (_Param(1) = _Param(0) != 0.0f ) )
+// The following are redundant but match LV prims and are needed so they can be polymorphic over arrays/clusters of complex
+DECLARE_VIREO_PRIMITIVE3( ComplexToPolarComplexSingle, ComplexSingle, Single, Single, _Param(1) = abs(_Param(0)); _Param(2) = arg(_Param(0)) )
+DECLARE_VIREO_PRIMITIVE4( PolarToReOrImSingle, Single, Single, Single, Single, ComplexSingle z = polar(_Param(0), _Param(1)); _Param(2) = z.real(); _Param(3) = z.imag(); )
```
sanmut: Should these oeprations like abs, polar (and any of the public APIs on Complex) handle Infinity, -Infinity, NAN, etc explicitly?

test-it/ComplexMathFunctions.via
```diff
@@ -1,21 +1,150 @@
 
-
 define(HelloComplex dv(.VirtualInstrument (
     Locals: c(
         //-------------------
         e(dv(.Double 2.0) dMagnitude)
         e(dv(.Double 0.5) dPhase)
+        e(dv(.Double 0.0) dReal)
```
sanmut: Should we add tests where input to the supported complex operations is special floating point numbers such as Infinity, -Infinity, NAN, etc?

source/core/Math.cpp
spathiwa: The complex support uses std:: complex from STL and already handles zero and Inf/NaN support correctly.

test-it/ComplexMathFunctions.via
spathiwa: Good point.  I'll add those when I get back.

cglzaguilar:
@spathiwa please add test cases requested by @sanmut in a later submission. I'm merging this one, so that I can finalize work on the DFIR side.

PR #217, opened 2017-03-14
Fix DST handling

spathiwa:


source/core/Date.cpp
```diff
@@ -176,23 +176,19 @@ namespace Vireo {
...
-#elif kVireoOS_emscripten
-        TempStackCString result;
-        result.AppendCStr(jsTimestampGetTimeZoneAbbr());
-        snprintf(timeZoneAbbr, sizeof(timeZoneAbbr), "%s", result.BeginCStr());
+#if (kVireoOS_linuxU || kVireoOS_macosxU || kVireoOS_emscripten)
+        time_t rawtime = timestamp.Integer() - kStdDT1970re1904;
+        struct tm timeinfo;
+        localtime_r(&rawtime, &timeinfo);
+        snprintf(timeZoneAbbr, sizeof(timeZoneAbbr), "%s", timeinfo.tm_zone);
+// #elif kVireoOS_emscripten variant deleted; localtime_r works correctly
```
cglzaguilar: Remove the comment, in the future is just going to cause confusion about what variant was deleted.

It was helpful for the code review.

source/core/Date.cpp
```diff
@@ -309,30 +305,23 @@ namespace Vireo {
...
-        TypeRef parseType = THREAD_TADM()->FindType(tsInt32Type);
-        Int32 minutes;
-        parser.ParseData(parseType, &minutes);
-        _systemLocaleTimeZone = -(minutes * kSecondsPerMinute);
-        // flipping the sign of the time zone since JS runtime will be positive
-        // for being behind UTC and negative for being ahead. ctime assumes
-        // negative for behind and postive for ahead. We attempt to match the ctime result.
-        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset
-#elif (kVireoOS_linuxU || kVireoOS_macosxU)
+// #if kVireoOS_emscripten formerly here which was using jsTimestampGetTimeZoneOffset has been deleted; the localtime_r emulation works correctly
```
cglzaguilar: Remove this comment, it refers to something that used to be and doesn't add value to the current code.

source/core/Date.cpp
```diff
@@ -309,30 +305,23 @@ namespace Vireo {
...
+        time_t now;
+        if (utcTime)
+            now = utcTime - kStdDT1970re1904;
+        else
+            now = time(NULL);
         localtime_r(&now, &tm);
         _systemLocaleTimeZone = int(tm.tm_gmtoff);
 #else
         // flipping the sign of the time zone
+        // FIXME -- if utcTime is non-zero, this should return timeZone bias based on that time (in GMT secs since epoch)
```
cglzaguilar: This is a todo, Please file a defect and add the defect id here in the comment.

source/core/Date.cpp
spathiwa: I actually disagree... I think comments about historical changes are useful as long as they don't clutter the code .  Leaving the entire former block of code intact but commented out would hurt readability and maintainability, but I think having a line indicating there was a change so someone knows to look back in git for differences in the event of a newly discovered bug is useful.

source/core/Date.cpp
spathiwa: A Rally defect or a git issue?  I think putting Rally defect numbers in github, when Rally is not accessible outside NI, is a bad idea.  I'll file a git issue.

cglzaguilar: Git issue is fine with me.

spathiwa: Done.

source/core/Date.cpp
cglzaguilar: Ok, on this one if you think there is value on that comment.

sanmut: I see both of your points. It is not a bad idea to put the current date against comments like this. After several months or a year of testing and shipping products and  having good runtime, it might be ok to delete the comment then.

source/core/Date.cpp
```diff
@@ -309,29 +307,25 @@ namespace Vireo {
...
-        // flipping the sign of the time zone since JS runtime will be positive
-        // for being behind UTC and negative for being ahead. ctime assumes
-        // negative for behind and postive for ahead. We attempt to match the ctime result.
-        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset
-#elif (kVireoOS_linuxU || kVireoOS_macosxU)
+// #if kVireoOS_emscripten formerly here which was using jsTimestampGetTimeZoneOffset has been deleted; the localtime_r emulation works correctly
+#if (kVireoOS_linuxU || kVireoOS_macosxU || kVireoOS_emscripten)
         struct tm tm;
-        time_t now = time(NULL);
+        time_t now;
```
sanmut: <nitpick> I would rename "now" to something general since it is used for time that is passed in the argument also.

source/include/Date.h
```diff
@@ -54,7 +54,7 @@ class Date {
         Date(Timestamp timestamp, Int32 timeZoneOffset); // this API is conceptually wrong; tzo should be computed from UTC date
         Date(Timestamp timestamp, bool isUTC = false);
         ~Date();
-        static Int32 getLocaletimeZone();
+        static Int32 getLocaletimeZone(Int64 utcTime = 0);
```
sanmut: Should we really need to support default value of 0 here? The one place the default value is being used seems to be "oid Format(SubString *format, Int32 count, StaticTypeAndData arguments[], StringRef buffer)" in NumericString.cpp. In that, if timezone is not 0, should we try to use again the passed timestamp from the argument? That may need a bit of refactoring.

If there is a need for this, I would add a second boolean "isUTC" argument as done in other places (instead of default value of 0) to make this clear and consistent at the callsites. If this function is used incorrectly without passing argument, I wonder whether some tests that use the exact epoch as timestamp (such that this is 0) may be failing?

spathiwa: This is actually different than the isUTC case in other places, because it's really DST rules for the local TZ in both cases.
However, you're right that it isn't needed.  I'll fix the only caller that passes 0, since that should also pass a timestamp.

source/core/Date.cpp
spathiwa: Will do.

source/core/NumericString.cpp
```diff
@@ -817,9 +817,11 @@ void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], Strin
                     case 't':
                     case 'T':
                     {
-                        Int32 tz = Date::getLocaletimeZone();
-                        if (fOptions.EngineerNotation || fOptions.FormatChar == 't') {
-                             tz = 0;
+                        Int32 tz = 0;
+                        if (fOptions.FormatChar == 'T' && !fOptions.EngineerNotation) {
```
sanmut: I am not sure if we want to add an extra check "argType->IsTimestamp()" (or assert/fail inside that about argType)?

source/core/Date.cpp
```diff
@@ -314,12 +314,8 @@ namespace Vireo {
 // #if kVireoOS_emscripten formerly here which was using jsTimestampGetTimeZoneOffset has been deleted; the localtime_r emulation works correctly
```
sanmut: I think we should delete all references of jsTimestampGetTimeZoneOffset from .cpp and especially .js  to make sure we don't accidentally mix and match relying on emscriptened native functionality and JS functionality and deal with errors due to differences.

PR #221, opened 2017-03-16
Implement a bunch of complex trig functions

spathiwa:
Cotangent, Sinc, ArcTan, ArcTan2, ArcSin, ArcCos, ArcCotanget,
ArcSecant, ArcCosecant.
(Some of these trig functions were 'working' for complex by acting
on them as if they were just clusters of two reals, and getting the
wrong answer.)
Rename Tan to Tangent for complex to match real version.
Also implement Sign, Incr/Decrement, Reciprocal, Scale2X
Scale2X accepts more numeric types for 'n' (exponent) arg.


spathiwa:
In order to get library support for asin, acos, and atan on complex numbers, I had to enable c++11 in the Travis build.  The support is technically part of c++11.   There's no way to enable it in C++ without turning on the rest of C++11 features as far as I can tell.    It works on Mac w/o this option.  I don't know the effect on Windows.
This may be controversial and warrant discussion.

source/core/GenericFunctions.cpp
```diff
@@ -1290,6 +1290,17 @@ VIREO_FUNCTION_SIGNATURET(VectorOpInternal, VectorOpInstruction)
                 *(Double *)scalarOpSnippet->_p2 = isIdentityOne;
             break;
         }
+        case kEncoding_Cluster:
+            if (array->ElementType()->IsComplex()) {
+                if (elementSize == 8) {
```
sanmut: Should this be sizeof(ComplexSingle) or sizeof(float)*2?

spathiwa: I think it should be sizeof(ComplexSingle) but I didn't do this because the type is not exposed in this file; it's local to Math.cpp where all of the actual complex support is and I didn't want to move it to a header file.  However, sizeof(float)*2 is an acceptable compromise.

.travis.yml
```diff
@@ -21,7 +21,7 @@ before_install:
 install:
 - npm install
 - pip install --user -r requirements.txt
-- if [ "$CXX" = "g++" ]; then export CXX="g++-4.9" CC="gcc-4.9"; fi
+- if [ "$CXX" = "g++" ]; then export CXX="g++-4.9" CC="g++-4.9 -std=c++11"; fi
```
sanmut: I don't know about these env with respect to emscripten. 

Are these used by emscripten or only native compilers? Regardless, we should keep these identical in our local machines (same OS) and travis. If so, should these changes go into make files also?

Also, if these are used by emscripten, does this increase size of the runtime with C++11?

Besides should we refrain from switching to "C++11" for portability (of Vireo) reasons?

.travis.yml
```diff
@@ -32,7 +32,7 @@ before_script:
 - nohup python -m httpbin.core --port 64526 &
 - while ! nc -q 1 127.0.0.1 64526 </dev/null; do sleep 2; done
 - curl http://127.0.0.1:64526/get?show_env=1
```
sanmut: Should we update make files instead?

.travis.yml
```diff
@@ -32,7 +32,7 @@ before_script:
...
-- make
+- make -e
```
sanmut: Can we update Makefiles instead of letting this override env variables?

source/core/TypeAndDataManager.cpp
```diff
@@ -874,7 +876,18 @@ Boolean TypeCommon::IsTimestamp()
     }
     return false;
 }
-
+//------------------------------------------------------------
+Boolean TypeCommon::IsComplex()
```
sanmut: Nice :+1:

.travis.yml
cglzaguilar: If I'm not mistaken Craig said we need C++11 in order to support asin, acos, and atan on complex numbers, are there other alternatives?

.travis.yml
spathiwa: The problem is that the behavior is already different per-platform, as the tools are different.  We use the same version of emscripten everywhere, but the native build uses native compilers.  The Travis build system's version of g++ we're using requires -std=c++11 for access to the complex trig functions we need, but other platforms don't appear to.  I did it this way because it was the easiest way to address a problem specific to the travis native build, but it could be done in the Makefile as well.

.travis.yml
spathiwa: The support we need is actually just in a header file guarded by #if __cplusplus >= 201103, but nothing in that code requires an underlying C++11 compiler feature, it's simply part of a newer stdlib standard.  We could extract the parts we need into Vireo and use it directly... if there aren't license issues with doing that.

PR #222, opened 2017-03-16
Remove http warnings

funkmaster2:
TA224253: The http client will no longer return warnings. Therefore I removed the code that evaluates forbidden headers and then merges in a warning if a header is forbidden when performing a http.addHeader call.

### DIFF
--- a/source/io/module_httpClient.js
+++ b/source/io/module_httpClient.js
@@ -44,7 +44,7 @@
 
         // WebVI Specific
         WEBVI_UNSUPPORTED_INPUT: 363798,
-        INVALID_HEADER: 363798,
+        INVALID_HEADER: 363798
     };
 
     var DEFAULT_TIMEOUT_MS = 10000;
@@ -71,7 +71,7 @@
 
         // Static Public Functions
         // None
-        
+
         // Prototype creation
         var child = HttpClient;
         var proto = child.prototype;
@@ -433,7 +433,6 @@
             var header = Module.eggShell.dataReadString(headerPointer);
             var value = Module.eggShell.dataReadString(valuePointer);
             httpClient.addHeader(header, value);
-
         };
 
         Module.httpClient.jsHttpClientRemoveHeader = function (handle, headerPointer, errorStatusPointer, errorCodePointer, errorSourcePointer) {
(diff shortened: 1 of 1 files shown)


funkmaster2:
@rajsite , when you get back can you take a look at these changes for removing warnings from the http client?

source/io/module_httpClient.js
```diff
@@ -71,45 +70,7 @@
         // None
 
         // Static Public Functions
-        // https://fetch.spec.whatwg.org/#forbidden-header-name
-        var forbiddenHeaderNames = [
```
sanmut: Since we have spent time finding the forbidden header names, I would leave a comment with an explanation and all these names.

source/io/module_httpClient.js
```diff
@@ -44,8 +44,7 @@
 
         // WebVI Specific
         WEBVI_UNSUPPORTED_INPUT: 363798,
-        INVALID_HEADER: 363798,
-        FORBIDDEN_HEADER_WARNING: 363798
+        INVALID_HEADER: 363798
```
sanmut: I would leave the comma at the end even though this is the last item (as a convention). That way, when items are added after the last item or the last item is removed, the diff won't show an extra line just because of a comma. Many groups have this convention for many programming languages.

funkmaster2: the eslint check complained when I had the comma in there.

/home/travis/build/ni/VireoSDK/source/io/module_httpClient.js
   47:31  error  Unexpected trailing comma                comma-dangle

I'm not sure if it would require changing all other structures in the file, or a change in the travis tests to allow trailing commas.

rajsite: Not a convention in our js code bases. Older browsers error on trailing commas. The linter enforces the right convention.

source/io/module_httpClient.js
rajsite: I wouldn't leave the names, it is liable to get outdated. The link to the spec is probably sufficient.

source/io/module_httpClient.js
rajsite: Also just for interest, trailing commas are not allowed in valid JSON data

PR #226, opened 2017-03-21
Fix MSVC build break

sanmut:


source/core/Math.cpp
```diff
@@ -12,6 +12,10 @@ SDG
  */
 
 #include <cstdlib>
+
+#ifdef _MSC_VER
```
spathiwa: I think this belongs in BuildConfig.h and not here.

sanmut: How about the following in BuildConfig.h (and use kVireoCompiler_MSVC here)?
#if defined(_MSC_VER)
	#define kVireoCompiler_MSVC 1
#endif
I think we want to distinguish OS vs Compiler.

source/include/BuildConfig.h
```diff
@@ -28,6 +28,10 @@ SDG
 	#define kVireoOS_windows 1
 #endif
 
+#ifdef _MSC_VER
```
cglzaguilar: Close to the bottom of the file, there is already a Vireo windows section, shouldn't be put it there?

 #elif (kVireoOS_windows)
 #if (_MSC_VER < 1900)		 #if (_MSC_VER < 1900)
     #define snprintf _snprintf		     #define snprintf _snprintf
 #endif		 #endif

sanmut: Will do.

PR #227, opened 2017-03-21
Fix to ReadJSON precision issue on floating point numbers

cglzaguilar:


source/core/TDCodecVia.cpp
```diff
@@ -1770,6 +1770,7 @@ TDViaFormatter::TDViaFormatter(StringRef string, Boolean quoteOnTopString, Int32
         _options._bEscapeStrings = false;
         _options._fmt = formatVIA;
     } else if (format->ComparePrefixCStr(formatJSON._name)) {
+        _options._precision = 17;
```
spathiwa: I think you should use 16 if it works correctly.
Notice a lot of the constants that had to change below introduce inaccuracy because 17 is slightly more accurate than representable. e.g.  2.8 becomes  2.7999999999999998.

cglzaguilar: 2.7999999999999998 is the value that is currently returned in CG for 2.8 in FlattenToJSON. I'll see if 16 returns the same value.

cglzaguilar: 16, causes the tests to fail. So 17 seems to be the correct number.

cglzaguilar: I talked to Craig and he's fine with the 17.

PR #228, opened 2017-03-22
Resize arrays eggshell

gleonoliva:
Here are my changes for exposing ResizeArray to eggshell.

I have a couple of questions about conventions for this repo. 

1. Do you think more validation is required for the rank and newLengths is required in `CEntryPoints.cpp`? Iwas just assuming that the actual call for resizing on line 229 would handle this.
2. I tried to differentiate the reason resizing failed with different error codes. Don't know if is ok or just use one.
3. Do you think more tests are needed?
4. Should I modify `CEntryPoints.h` with the function signature for this and `EggShell_GetArrayDimLength`, there might be more missing there. Is this relevant?


source/io/module_eggShell.js
```diff
@@ -128,6 +130,22 @@
             return EggShell_GetArrayDimLength(v_userShell, vi, path, dim);
         };
 
+        Module.eggShell.resizeArray = publicAPI.eggShell.resizeArray = function (vi, path, newDimensionSizes) {
+            var int32Byte = 4;
+            var length = newDimensionSizes.length;
+            var buffer = Module._malloc(length * int32Byte);
```
cglzaguilar: consider renaming length to rank and buffer to newLengths, to keep aligned with the names EggShell_ResizeArray is using

test-it/karma/publicapi/resizeArray.test.js
```diff
@@ -0,0 +1,76 @@
...
+
+    it('cannot resize fixed 2d arrays', function (done) {
+        var variableName = 'fixedArray2d';
+        runSlicesAsync(function (rawPrint, rawPrintError) {
+            expect(rawPrint).toBeNonEmptyString();
+            expect(rawPrintError).toBeEmptyString();
+
+            var resized = vireo.eggShell.resizeArray(viName, variableName, [3, 4]);
+
+            expect(resized).toBe(-2);
```
cglzaguilar: Add another tests when the returned value is -1, that is the array symbol didn't exist.

cglzaguilar:
Answer to your questions

1) Yes, ResizeDimensions should take care of this, if it doesn't we should fixed. Do you have test cases in mind?

2) More codes is better, let's give the user feedback as specific as we can.

3) Yes, once for error code -1, and probably we might need more tests for ResizeDimensions to handle incorrect parameters, like a Rank of -1 or stuff like that.

4) I didn't understand your question. Can you elaborate more?

gleonoliva:
Sure, the thing is that I didn't modify `CEntryPoints.h` is it necessary? Because I didn't see any errors.

source/core/CEntryPoints.cpp
```diff
@@ -210,6 +210,29 @@ VIREO_EXPORT Int32 EggShell_GetArrayDimLength(TypeManagerRef tm, const char* viN
...
+{
+    TypeManagerScope scope(tm);
+
+    SubString objectName(viName);
+    SubString path(eltName);
+    void *pData = null;
+
+    TypeRef actualType = tm->GetObjectElementAddressFromPath(&objectName, &path, &pData, true);
+    if (actualType == null || !actualType->IsArray()) {
+        return -1;
```
spathiwa: If you are going to return more than one error code, make them positive and use standard LabVIEW-style error codes.
kLVError_ArgError (=1)  for bad type/null array, and kLVError_MemFull (=2) for resize failure.
Add these errors to enum LVError in DataTypes.h.

I'll make a pass and modify the other prims which return local error codes which are not actually prim-specific to use these too.

gleonoliva: The thing is there are a bunch of places in this same file that return similar "error codes" but I don't see any constants. I can do it anyway with hopes that we improve error handling.

spathiwa: Most of the other places are actually intended to be returning sizes, though, and are encoding -1 as an 'error' size.  The return value there is not a pure error code.

PR #229, opened 2017-03-23
Add support for formatting of complex numbers

spathiwa:


source/core/Math.cpp
```diff
@@ -899,7 +899,7 @@ DECLARE_VIREO_PRIMITIVE3( Scale2XComplexSingleInt8, ComplexSingle, Int8, Complex
     _Param(2).real(Scale2X_SingleInt32(_Param(0).real(), Int32(_Param(1))));
     _Param(2).imag(Scale2X_SingleInt32(_Param(0).imag(), Int32(_Param(1))));
 } )
-DECLARE_VIREO_PRIMITIVE3( PolarComplexSingle, Single, Single, ComplexSingle, (_Param(2) = _Param(1) == 0.0 ? ComplexSingle(_Param(0),0.0) : _Param(0) >= 0.0 ? polar(_Param(0), _Param(1)) : polar(-_Param(0), _Param(1)+Single(M_PI))) )
+DECLARE_VIREO_PRIMITIVE3( PolarComplexSingle, Single, Single, ComplexSingle, (_Param(2) = _Param(1) == 0.0f ? ComplexSingle(_Param(0),0.0) : _Param(0) >= 0.0f ? polar(_Param(0), _Param(1)) : polar(-_Param(0), _Param(1)+Single(M_PI))) )
```
cglzaguilar: I'm curious if there are other places in the code where the same fix needs to be applied. Consider doing a global search for == 0.0 and evaluate if the same fix is required.

source/core/NumericString.cpp
```diff
@@ -206,7 +206,7 @@ void ReadPercentFormatOptions(SubString *format, FormatOptions *pOptions)
 }
 //---------------------------------------------------------------------------------------------
 void GenerateFinalNumeric (const FormatOptions* , char* , Int32* , TempStackCString* , Boolean );
-void RefactorLabviewNumeric(const FormatOptions* , char* , Int32* , Int32 , Int32 );
+void RefactorLVNumeric(const FormatOptions* , char* , Int32* , Int32 , Int32, Boolean );
```
cglzaguilar: Consider removing "LV" as well. As people start using Vireo for other purposes having references to LV will make less sense.

source/core/NumericString.cpp
```diff
@@ -498,6 +498,13 @@ void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], Strin
                 ErrFormatMissingNumArgs(format, count, arguments, buffer, tempString);
             }
 
+            char replacementString[2*kTempCStringLength]; // temp used for modifying string to support opts for sigDigits, engineering notation, etc.
+            Int32 complexArg = 0; // when formatting complex data, 0==real, 1==imag, 2==finished
```
cglzaguilar: Consider using an enumeration for those values, instead of having hardcode numbers.

source/core/NumericString.cpp
```diff
@@ -559,33 +582,49 @@ void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], Strin
                             if (precision < 0) {
                                 precision = 0;
                                 truncateSignificant = exponent + 1 - fOptions.Significant;
-                                // need truncate the integer part of the float because the sprintf doesnt do this for us.
+                                // need truncate the inxteger part of the float because the sprintf doesnt do this for us.
```
cglzaguilar: Is this an intentional typo? Please explain why or just fix it.

source/core/NumericString.cpp
```diff
@@ -1109,13 +1179,20 @@ void RefactorLabviewNumeric(const FormatOptions* formatOptions, char* bufferBegi
                 Int32 sizeOfExpoent = snprintf(tempNumber + baseIndex, kTempCStringLength-baseIndex, "E%+d", (int)exponent);
                 baseIndex += sizeOfExpoent;
             }
-            TempStackCString numberPart((Utf8Char*)tempNumber, baseIndex);
-            GenerateFinalNumeric(formatOptions, bufferBegin, pSize, &numberPart, negative);
+            tempNumber[baseIndex] = 0;
```
cglzaguilar: This code change seems to be identical to the previous change, now that we more than two lines of code doing same, consider creating a helper method, to be called in both places.

source/core/StringUtilities.cpp
```diff
@@ -118,25 +119,41 @@ Boolean SubString::CompareCStr(ConstCStr begin2) const
...
-        }
     }
     // Comparison is true if supplied string is the same length.
     return (*pBegin2 == 0);
 }
 //------------------------------------------------------------
 Boolean SubString::ComparePrefix(const Utf8Char* begin2, Int32 length2) const
 {
-    if (length2 > Length()) {
+    if (length2 > Length())
```
cglzaguilar: Can we make Compare and ComparePrefix to share code or are there performance implications? Also we have ComparePrefixIgnoreCase, while Compare used to take in a boolean for ignore case, and that is not the case for ComparePrefix.

source/core/Math.cpp
spathiwa: I did this just because sanmut requested it, but I did some testing and:
There is no difference at all in the code generated by emscripten based on whether you use an 'f' suffix for float literals.
On native, the compiler is smart enough to not emit  double if the constant is exactly representable as float, and automatically keeps it float.  So we never need to say 0.0f or 1.0f.

source/core/NumericString.cpp
spathiwa: We also have LV error codes which we must support, so I don't think we need to pretend there is no relationship to LabVIEW.   I did this rename because it wasn't capitalized correctly and just seemed too verbose.

source/core/NumericString.cpp
spathiwa: It's also an integer offset to GetSubElement, so I wasn't sure this was appropriate.  Hopefully the comment is sufficient.

test-it/ComplexConversions.via
```diff
@@ -0,0 +1,69 @@
...
+        IsEQ(local21 g8 local23)
+        Copy(local23 dataItem_NumberToOctalPass )
+        And(local11 local15 local30)
+        And(local19 local30 local30)
+        And(local23 local30 local30)
+        Copy(local30 local25 )
+        Copy(local25 dataItem_AllPass )
+        Printf ("000 All tests pass: '%s'.\n" dataItem_AllPass)
+        Printf ("010     Complex Format Pass: '%s'.\n" dataItem_ComplexFormatPass)
+		Printf ("020     Number to Hexadecimal Pass: '%s'.\n" dataItem_NumberToHexadecimalPass)
```
cglzaguilar: Align these statements with the previous ones.

source/core/NumericString.cpp
spathiwa: Oops.  Fixed.  I noticed this when I previewed by diff but glossed over it because I thought I had fixed the typo rather than introducing it.

source/core/NumericString.cpp
```diff
@@ -559,33 +582,49 @@ void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], Strin
...
+                            intDigits = exponent;
+                        // further processing done by shared code which calls RefactorLVNumeric below switch
                         argumentIndex++;
                     }
                     break;
                     case 'e': case 'E':
                     {
                         TypeRef argType = arguments[argumentIndex]._paramType;
-                        Double tempDouble = ReadDoubleFromMemory(argType,  arguments[argumentIndex]._pData);
+                        void *pData = arguments[argumentIndex]._pData;
```
spathiwa: I had already tried to do this but didn't like what I came up with because it accesses 5 different local variables which have to be passed, plus has a continue which needs to be converted to a return value.  ArgumentIndex and parseFinished really are specific to the context and don't make much sense when moved into a helper function, so only the argType and reload and pData really make sense to be factored out, and then it doesn't seem worth it.

source/core/NumericString.cpp
spathiwa: Good catch.  Done.

source/core/StringUtilities.cpp
spathiwa: We could; I didn't because of performance reasons.  Perhaps we could do it and then mark the shared function inline.

ComparePrefix needs a boolean argument for ignoreCase because is some cases the ignoreCase option is programmatic and not known at compile time.  This is even called directly from some prims.
I wanted a version in which ignoreCase IS known at compile time for performance, since this function is also very commonly called during parsing.
ComparePrefix is only used during parsing and we don't need a programmatic ignoreCase argument because there would be no callers.

test-it/ComplexConversions.via
spathiwa: Hey, you gave me that file that way.  :-)

source/core/StringUtilities.cpp
spathiwa: I made the several non-IC Compare functions call a common (inline) function.

source/core/NumericString.cpp
cglzaguilar: Ok, let's do this refactoring later then, if the structure of the code doesn't allow for clean code, it is a warning sign to me.

source/core/StringUtilities.cpp
```diff
@@ -71,14 +69,8 @@ Int32 SubString::DigitValue(Utf32Char codePoint, Int32 base)
...
-//------------------------------------------------------------
-Boolean SubString::Compare(const Utf8Char* begin2, IntIndex length2) const
-{
-    if (length2 != Length()) {
-        return false;
-    }
-    const Utf8Char* sCompare = _begin;
-    const Utf8Char* sEnd = _begin+length2;
+
+inline Boolean CompareString(const Utf8Char* sCompare, const Utf8Char* sEnd, const Utf8Char* begin2) {
```
spathiwa: I'm not sure what's confusing about it, but I added a comment explaining the parameters, and renamed begin2 to sCompare2.

I don't like renaming sCompare to begin1 because the variable is incremented as it steps through the array, so after one iteration it's no longer the beginning.   This was also true of begin2, which I don't like.
So instead I  added '1' to the first two args, and renamed the other string pointer to sCompare2.

source/core/NumericString.cpp
spathiwa: I actually went ahead and did this by moving it before the switch instead of making an extra function.
I didn't do this before because it requires extra checks for format code and that the argument is safe to access, but it's probably still worth it, so I did it.
During the process, I noticed an issue with error handling and fixed that too.

Issue #218, opened 2017-03-14
Windows DST support incomplete


spathiwa:
Date::getLocaletimeZone() needs to be extended on Windows to be able to set the timezoneoffset based on whether DST is in effect at a given/passed-in time, rather than the current time. 

This support has already been added for Linux/Mac/JS

spathiwa:
I think this win32 api might be helpful:
https://msdn.microsoft.com/en-us/library/ms724949(VS.85).aspx

sanmut:
https://github.com/ni/VireoSDK/commit/5c14df1a1061528ce8a975f2b1a15b0095f42527 (#230) fixes this issue. Closing.

PR #230, opened 2017-03-23
Fix build breaks on Windows and fix timezone, DST for all platforms

sanmut:


source/core/Date.cpp
```diff
@@ -311,14 +372,7 @@ namespace Vireo {
...
 #else
-        // Issue #218 TODO -  FIXME: if utcTime is non-zero, time zone bias should be based on utcTime passed in
-        // (in GMT secs since epoch), not current time.
-        // Possibly use SystemTimeToTzSpecificLocalTime
-        
-        // flipping the sign of the time zone
-        TIME_ZONE_INFORMATION timeZoneInfo;
-        GetTimeZoneInformation(&timeZoneInfo);
-        _systemLocaleTimeZone = -int((timeZoneInfo.Bias) * kSecondsPerMinute);
+        _systemLocaleTimeZone = GetTimeZoneOffsetSeconds(utcTime);
```
cglzaguilar: Glad you took care of this issue, once the pull request is in, please go ahead and close Issue #218

source/core/Date.cpp
```diff
@@ -180,27 +233,35 @@ namespace Vireo {
...
 // #elif kVireoOS_emscripten variant deleted (03/2017); localtime_r works correctly
 #elif kVireoOS_windows
-        // Issue #218 TODO - FIXME: return time zone info/dst based on timestamp.Integer(), not current time.
-        // Possibly use SystemTimeToTzSpecificLocalTime
         TIME_ZONE_INFORMATION timeZoneInfo;
-        int rc = GetTimeZoneInformation(&timeZoneInfo);
-        wchar_t *timeZoneName = L"TimeZoneUnknown";
-        if (rc == TIME_ZONE_ID_STANDARD)
+        int rc = GetTimeZoneInformationForYear(year, NULL, &timeZoneInfo);
+        char timeZoneName[kTempCStringLength] = "TODO-TimeZone";
```
spathiwa: "TODO" doesn't seem like a string that should be exposed to the user if an error.

source/core/Date.cpp
```diff
@@ -113,22 +167,21 @@ namespace Vireo {
...
             if (prev == ' ' && input[i] != ' ')
             {
-                output += input[i];
+                output[j] = input[i];
+                ++j;
             }
             prev = input[i];
         }
-        return output;
+        output[j] = 0;
```
spathiwa: I don't have objections to your C string refactor of this, but is this really the only way to get an abbreviated time zone on Windows? (I think Devin did this originally.)
Did you look for a different API that doesn't require use to hackishly abbreviate the string by looking for spaces?

source/core/Date.cpp
```diff
@@ -53,6 +53,60 @@ namespace Vireo {
 
     Int32 Date::_systemLocaleTimeZone = 0;
 
+#if kVireoOS_windows
+    // UnixTimeToFileTime and UnixTimeToSystemTime are from https ://support.microsoft.com/en-us/help/167296/how-to-convert-a-unix-time-t-to-a-win32-filetime-or-systemtime
+    void UnixTimeToFileTime(time_t t, LPFILETIME pft)
```
sanmut: Good point. Done.

source/core/Date.cpp
sanmut: No. LV's behavior seems to be locale specific. I haven't spent much time on this. I am not sure whether we want to abbreviate at all just to make the tests pass or make the tests platform specific and go with whatever the default is for the platform (don't abbreviate for Windows and use the abbreviated timezones for Linux/Mac). Let's make a final decision on that in the future, we can spend a bit more effort on this to clean this up.

source/core/Date.cpp
sanmut: Changing to "UnknownTimezone"

source/core/Date.cpp
sanmut: :-) Will close the issues after this gets merged. Thanks.

PR #232, opened 2017-03-24
Reduce number of HTTBin offline warnings during test

rajsite:


test-it/karma/utilities/testHelpers.httpBinHelpers.js
```diff
@@ -50,7 +51,10 @@
             throw new Error('queryHttpBinStatus must be called before makeTestPendingIfHttpBinOffline');
         }
         if (serverOnline === false) {
-            console.warn('HTTPBin Server offline, skipping HTTP tests');
+            if (alreadyWarned === false) {
+                console.warn('HTTPBin Server offline, skipping HTTP tests');
+            }
+            alreadyWarned = true;
```
cglzaguilar: Consider removing this inside the if. so it doesn't get executed over and over again.

rajsite:
Updated!

PR #233, opened 2017-03-27
Complex documentation

cglzaguilar:


cglzaguilar:
Everybody, please comment if you have any concern. Otherwise I'll rebase and merge this later today.

source/dox/Internals/API.md
```diff
@@ -0,0 +1,85 @@
+\defgroup VIAA_PI VI Assembly Application Programming Interface (API)
```
spathiwa: This looks like a typo of VIA_API

spathiwa:
It seems weird that every prim gets its own \defgroup.

source/dox/Internals/API.md
cglzaguilar: Fixed on latest file changes.

cglzaguilar:
Let's consider this an example of how the API might look like. There are several questions as to how best efficiently produce the documents and which tools to use. I'm open to suggestions.

PR #237, opened 2017-04-03
Updated errors returned by HTTP calls with call chain.

cglzaguilar:


source/core/ExecutionContext.cpp
```diff
@@ -228,31 +228,54 @@ VIREO_FUNCTION_SIGNATURE1(CallChain, StringRefArray1D*)
...
+    StringRefArray1D* callChain = (StringRefArray1D*)StringRefArray1D::New(itemType);
+    GetCallChainArray(callChain);
+    for (int i = 0; i < callChain->Length() - 1; i++) {        
+        stringRef->AppendStringRef(*callChain->BeginAt(i));
+        stringRef->AppendCStr("->");
+    }
+    if (callChain->Length() > 0) {
+        stringRef->AppendStringRef(*callChain->BeginAt(callChain->Length() - 1));
+    }
+    stringRef->AppendCStr("\r\n");
```
rajsite: I think that AppendCallChainString should not include the new line character. The newline is extra formatting that should  be moved to the AddCallChainToSourceIfErrorPresent function

rajsite: Talked with Carlos in person and agreed that a newline was not needed at all

cglzaguilar: Done, removed the new line character.

source/io/HttpClient.cpp
```diff
@@ -39,21 +39,49 @@ extern "C" {
 }
 #endif
 
+void AddCallChainToSourceIfErrorPresent(ErrorCluster *errorCluster, ConstCStr methodName)
```
rajsite: Would be helpful to add a comment showing some examples:

ie
If existing source is empty string then output is
"Primitive in MyVI->SubVI"

If existing source contains contents such as "<APPEND>\nCustom Error Information" then output is
"Primitive in MyVI->SubVI<APPEND>\nCustom Error Information"

source/core/ExecutionContext.cpp
spathiwa: AppendCallChainString() appears to have no callers in this commit.
Variable callChain is leaked and should be freed.

rajsite: AppendCallChainString is called in HttpClient.cpp, it is collapsed by default in the online viewer because it is a large diff

spathiwa: Ah, okay, sorry.  They really should make that easier to notice.

Please address the leak, though.  You only create the call chain array to stringify it, but then don't delete it.

cglzaguilar: Leak has been fixed.

cglzaguilar:
Leak has been fixed, newline has been removed and comments had been updated.

source/core/ExecutionContext.cpp
```diff
@@ -267,8 +267,8 @@ void AppendCallChainString(StringRef stringRef)
     if (callChain->Length() > 0) {
         stringRef->AppendStringRef(*callChain->BeginAt(callChain->Length() - 1));
     }
-    stringRef->AppendCStr("\r\n");
     itemType->ClearData(&callChain);
+    free(callChain);
```
spathiwa: I'm sorry, I missed the ClearData() call you already had here because a chain of inline comments had been added right before it in the diff review. 

The ClearData is sufficient, you should remove the extra free() (which is probably not causing extra issues because ClearData nulls callChain and free silently ignored null pointers).
Sorry for the confusion.

spathiwa: (PR approved pending removing the free I made you add.  😄 )

cglzaguilar:
Removed the free call as requested.

cglzaguilar:
Newline change and comments had been addressed from Milan's feedback.

Issue #153, opened 2017-01-10
Standardize on a test configuration format


rajsite:
Right now we have 2 formats for organizing tests represented in the repository

1. Tests organized by folder 
test-it/
test-it/Blacklist
test-it/x
test-it/ManualTests

2. Tests organized by test-it/testList.json configuration

Currently browser tests are configured by having karma serve test-it/*.via and jasmine disables specific tests in testSuite.js. This follows the conventions of karma and jasmine but it could be reconfigured to do either method proposed above.

Any preference for folder-based vs list-based configuration?

rajsite:
I would prefer a mixture.
Tests go into folders based on feature
Test sets go into testList.json based on platforms. Glob patterns used to prevent having to modify configuration for every new test added.

ie
test-it/core
test-it/filesystem
test-it/manual
test-it/http

```json
testList.json
{
  "native": [
    "test-it/core/*.via",
    "test-it/filesystem/*.via"
  ],
  "node": [
    "test-it/core/*.via",
    "!test-it/core/specificTestDisabled.via"
  ],
  "browser": [
    "test-it/core/*.via",
    "test-it/http/*.via"
  ]
}
```

cglzaguilar:
Sounds good, but I guess the core directory will contain most of the tests. It is worth it to subdivide core into arrays, strings and so for?

sanmut:
Is there a way to just tag tests such that a test can have multiple tags. As the test suite develops, I have a feeling there might be some overlap of categories (either folder based or list based configuration) for tests and putting the tests in one category/folder might be more misleading than not categorizing them? This may not be a big concern if the categories are broad and coarse. I tend to intuitively like tags/labels where this problem can be avoided.

rajsite:
Talked to @sanmut and it sounds like the json file proposed for picking tests out with paths, wildcards, and exceptions (ie !somefile.via) sounds good

The open question was the same as @cglzaguilar about how to organize tests. I'm going to propose organizing the tests corresponding to the src tree.

ie 
src/core
src/io

test-it/core
test-it/io
test-it/results/core
test-it/results/io
test-it/karma/jstests

@cglzaguilar and @sanmut any alternate proposals?

sanmut:
That sounds good to me (core and io) and following source tree.

spathiwa:
To me, 'core' is too coarse, and provides no benefit for organization because almost all of the Vireo language features are there.
Also, I prefer having a results folder inside each category, not at the top level.

[...]

PR #239, opened 2017-04-04
Rename vtr test suite

rajsite:
I'm calling the vireo tests that rely on .vtr results files to pass or fail 🎉 "The Vtr Test Suite" 🎉 

I moved the via files for the vtr test suite to the folder /test-it/via
I renamed the folder holding vtr files from results to /test-it/vtr

I updated test.js to handle the new paths and tested Native and JS builds.
I also updated The Karma Test Suite to handle the new paths.


spathiwa:
I'm fine with the camel case consistency, and ambivalent about renaming 'results' to 'vtr', but please explain the rationale for moving all the tests to a via subfolder?  I think that just makes it less convenient to run tests from the command line.  More typing before you can press tab to complete filenames = bad.

spathiwa:
If we're going to have folders for the tests, I'd rather have the folders serve some purpose, like grouping based on category (Array, String, Queue, Complex, HTTP, etc.)...  but I'm not super-excited about that either.  Having a more consistent naming scheme for the test filenames would be good, though, regardless of whether they're moved into folders.

rajsite:
> I'm fine with the camel case consistency

Not part of this pull request

> please explain the rationale for moving all the tests to a via subfolder?

The test-it directory has many files with different purposes mixed together. Trying to organize them.  For example some of the following types of files are loose while others are placed in a sub directory:

test runners:
test.js
run-tests.sh
run-tests.bat
HelloNode.js

configuration:
testList.json

fixtures used as data for tests:
TestDataFolder/

manual test:
ManualTests/

test code:
*.via
Blacklist/*.via
ManualTests/*.via
x/*.via

So the rationale is to start cleaning this up. Some minor benefits are:
* making it a tad easier to write scripts not having to filter out a bunch of noisy files
* it's easier to quickly inspect and compare the via and vtr folders
* moving closer step-by-step to having files better organized by purpose which makes it easier to maintain. I didn't even know we had run-tests.sh and run-tests.bat before this. 

> If we're going to have folders for the tests, I'd rather have the folders serve some purpose, like grouping based on category (Array, String, Queue, Complex, HTTP, etc.)... but I'm not super-excited about that either. Having a more consistent naming scheme for the test filenames would be good, though, regardless of whether they're moved into folders.

Not being discussed here. This is about placing the test files into a separate folder, not about the organization of those files in the folder. Picket fencing of that topic is encouraged in the following issue: https://github.com/ni/VireoSDK/issues/153

> I think that just makes it less convenient to run tests from the command line. More typing before you can press tab to complete filenames = bad.

my change makes it so that you have to hit three characters before you can autocomplete:
`v + i + tab + <first letter of test> + tab`

if the via folder is given a different name so they don't conflict then it would be just two characters:
`v + tab + <first letter of test> + tab`

or even just name the folder v so still two characters but can skip the first autocomplete, less reaching with the pinky for that extra tab:
`v + / + <first letter of test> + tab`

P.S. I would love a presidential_programmer twitter account:
More typing before autocomplete. Bad.
Build a Gate to keep out intermittent tests and make the Reviewers pay for it #FailingCITimes #FakeResults

spathiwa:
OK, I'm fine with it, but please leave the results directory instead of renaming it vtr, so it begins with a different letter.  (Results also seems more descriptive/less obscure than vtr).

cglzaguilar:
I prefer the name ExpectedResults, instead of Results. That is what actually the folder is intended to hold. Results is more like the output of some run, that is not the case here.

spathiwa:
Fine with me.

rajsite:
Closing in favor of https://github.com/ni/VireoSDK/pull/241

PR #241, opened 2017-04-04
Slightly improve VTR test suite organization

rajsite:
Moved .via files to the ViaTests folder
Moved .vtr files to the ExpectedResults folder
Made sure the test runner can use the new hard coded paths, ie:
`node test.js ViaTests\HelloWorld.via`


spathiwa:
Hey @rajsite, 

When test.js is invoked given a via filename path, can you please make it test for the file exactly as given before prepending ViaTest/ to it?  I frequently work on tests in the current directory while debugging or developing new features, and don't "productize" the test by making expected results until it's nearly done.

rajsite:
Can you want to open a separate issue with an example? I'm not sure exactly what the problem is.

PR #243, opened 2017-04-12
Add support for new API HttpCredentials

sanmut:


source/io/module_httpClient.js
```diff
@@ -216,8 +216,8 @@
             // withCredentials allows cookies (to be sent / set), HTTP Auth, and TLS Client certs when sending requests Cross Origin
             // Setting to false so communication to servers with Access-Control-Allow-Origin: *
             // See https://w3c.github.io/webappsec-cors-for-developers/#anonymous-requests-or-access-control-allow-origin
-            // TODO mraj: We need a way to expose this configuration to users
-            request.withCredentials = false;
+            // if the jsHttpClientSetCredentials was already called by the user, use that value. Otherwise, default is "false"
+            request.withCredentials = (requestData.withCredentials || false);
```
rajsite: Based on my other comments,, withCredentials should be a property on this class instance so it would be accessed by this._withCredentials. You also would not need the default hard coded here as it would be set in the constructor.

source/io/module_httpClient.js
```diff
@@ -573,7 +573,8 @@
                 method: method,
                 url: url,
                 xhrTimeout: xhrTimeout,
-                buffer: buffer
+                buffer: buffer,
+                withCredentials: httpClient.withCredentials
```
rajsite: withCredentials is a property of an HttpClient instance, it should not be passed in here

source/io/module_httpClient.js
```diff
@@ -591,6 +592,19 @@
...
+            var errorStatus = Module.eggShell.dataReadBoolean(errorStatusPointer);
+            if (errorStatus) {
+                return;
+            }
+
+            var httpClient = findhttpClientOrWriteError(handle, 'LabVIEWHTTPClient:Credentials', errorStatusPointer, errorCodePointer, errorSourcePointer);
+            if (httpClient === undefined) {
+                return;
+            }
+            httpClient.withCredentials = (isWithCredentials !== 0);
```
rajsite: It looks like this is just tacking on the withCredentials property to this object instance.

Instead it should be managed by the HttpClient class.

in the HttpClient constructor it should set this._withCredntials to false
On the prototype a setWithCredentials function should be added (As withCredentials is only used internally by the HTTPClient a getWithCredentials is not necessary so do not add that)

On this line we then invoke httpClient.setWithCredentials() instead of assigning directly to a property

source/io/HttpClient.cpp
```diff
@@ -446,6 +447,21 @@ VIREO_FUNCTION_SIGNATURE10(HttpClientPost, UInt32, StringRef, StringRef, StringR
 }
 
 //------------------------------------------------------------
+// handle(0), supply credentials(1), error cluster(2)
+VIREO_FUNCTION_SIGNATURE3(HttpClientSetCredentials, UInt32, Boolean, ErrorCluster)
+{
+#if kVireoOS_emscripten
+   jsHttpClientSetCredentials(
```
rajsite: You may need to sync, @cglzaguilar added a new pattern for error handling that should be used here

rajsite:
I also don't see any tests being added.

HttpBin echoes back in the response body the headers that appear in an http request. This is done to be able to test that the request including some headers.

When withCredentials is set to false the Access-Control-Allow-Credentials header should be missing / set to false and when set to true the Access-Control-Allow-Credentials header should be set to true 

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials

The following is an example test that parses a response body to make sure a header is included 
https://github.com/ni/VireoSDK/blob/13608e9ca8da862c88102a05e6402c0e29e7cf4d/test-it/karma/http/httpget.test.js#L300

sanmut:
I included files that I didn't mean to. Will create a new pull request.

source/io/module_httpClient.js
sanmut: Agreed. Will do.

source/io/module_httpClient.js
sanmut: Thanks. Will do.

source/io/HttpClient.cpp
sanmut: Good catch. Will do.

PR #244, opened 2017-04-12
Improvements to Http Error handling and test coverage

rajsite:
This set of changes kind of grew, sorry about that. This is the high level overview:

- Added http js tests to cover more error cases
- Made karma aware of testLists.json and added a browser target to testList.json
- Added VTR tests for HTTP client methods. Found that VTR tests for the actual HTTP methods cannot be made because test.js pumps Vireo synchronously. Left a todo in test.js describing that issue in test.js
- Added a custom matcher for Vireo which can be extended to add pretty diffs, etc, in the future for karma reports
- Modified Replace1D to support resizing to zero in 2293bb789e99eaaa818dc64b6049e76cd5dca238 Only had to modify one test as a result and after looking at the test with Carlos it looks like it was initially wrong anyway. Just want to make sure it looks sane to y'all @sanmut @spathiwa 
- Moved the fixtures from a synchronous loading pattern to an asynchronous pattern so we are less reliant on deprecated browser features (synchronous loading is deprecated in browsers because it blocks the UI thread waiting on a network request)
- Cleaned up some dead code


source/io/HttpClient.cpp
```diff
@@ -76,6 +78,11 @@ void GenerateNotSupportedOnPlatformError(ErrorCluster *errorCluster, ConstCStr m
 VIREO_FUNCTION_SIGNATURE6(HttpClientOpen, StringRef, StringRef, StringRef, UInt32, UInt32, ErrorCluster)
 {
 #if kVireoOS_emscripten
+    if (!_ParamPointer(0) || !_ParamPointer(1) || !_ParamPointer(2) || !_ParamPointer(3) || !_ParamPointer(4) || !_ParamPointer(5)) {
```
rajsite: Agreed the checks are too aggressive and should be fixed in the future to better align with the optional / required pattern

As this implementation improves on previous behavior (where no null checking is performed and memory could be corrupt), and does not impact tests, or current uses I am leaving TODOs and creating an issue to track the work

rajsite: https://github.com/ni/VireoSDK/issues/245

source/core/module_coreHelpers.js
```diff
@@ -134,6 +117,43 @@
...
+            }
+
+            var existingErrorCode = Module.eggShell.dataReadInt32(existingErrorCodePointer);
+            var existingWarning = existingErrorCode !== CODES.NO_ERROR;
+            var newWarning = newErrorCode !== CODES.NO_ERROR;
+            if (existingWarning) {
+                return;
+            }
+
+            if (newWarning) {
```
rajsite: done

source/core/module_coreHelpers.js
```diff
@@ -134,6 +117,43 @@
...
+            var newError = newErrorStatus;
+
+            if (existingError) {
+                return;
+            }
+
+            if (newError) {
+                Module.eggShell.dataWriteBoolean(existingErrorStatusPointer, newErrorStatus);
+                Module.eggShell.dataWriteInt32(existingErrorCodePointer, newErrorCode);
+                Module.eggShell.dataWriteString(exisitingErrorSourcePointer, newErrorSource);
```
rajsite: done

PR #246, opened 2017-04-18
Add support for new API HttpCredentials

sanmut:


sanmut:
Addressed feedback from @rajsite from the previous pull request. The first test could not be automated. I need to work on the second try of automating the test using cookie. I will create a separate PR For that. Need to start the submission process that depends on this, so would like to get this merged first.

source/io/module_httpClient.js
```diff
@@ -625,6 +629,19 @@
                 Module.eggShell.setOccurrenceAsync(occurrencePointer);
             });
         };
+
+        Module.httpClient.jsHttpClientSetCredentials = function (handle, isWithCredentials, errorStatusPointer, errorCodePointer, errorSourcePointer) {
+            var errorStatus = Module.eggShell.dataReadBoolean(errorStatusPointer);
```
rajsite: Rebase to latest, we check error status on C++ side now

source/io/module_httpClient.js
```diff
@@ -625,6 +629,19 @@
...
+            if (errorStatus) {
+                return;
+            }
+
+            var httpClient = findhttpClientOrWriteError(handle, 'LabVIEWHTTPClient:Credentials', errorStatusPointer, errorCodePointer, errorSourcePointer);
+            if (httpClient === undefined) {
+                return;
+            }
+            httpClient.setCredentials(isWithCredentials !== 0);
```
rajsite: There are several named values of 0 for this file, please pick a descriptive one:
https://github.com/ni/VireoSDK/blob/incoming/source/io/module_httpClient.js#L29

source/io/module_httpClient.js
rajsite: Sorry, looks like you do have the new pattern on the C++ side.

That means you can remove this errorStatus check on the js side altogether.

sanmut:
I like @rajsite 's recommendation of ConfigCORS. Updating the PR.

rajsite:
Since we are propagating the name change can we also propagate it to the HttpClient class?
Ie _includeCredentialsDuringCORS
And setIncludeCredentialsDuringCORS()

source/io/HttpClient.cpp
```diff
@@ -541,5 +569,6 @@ DEFINE_VIREO_BEGIN(HttpClient)
     DEFINE_VIREO_FUNCTION(HttpClientPut, "p(io(.UInt32) i(.String) i(.String) i(.String) i(.Int32) o(.String) o(.String) o(.UInt32) io(" ERROR_CLUST_TYPE_STRING ") s(.Occurrence))")
     DEFINE_VIREO_FUNCTION(HttpClientDelete, "p(io(.UInt32) i(.String) i(.String) i(.Int32) o(.String) o(.String) o(.UInt32) io(" ERROR_CLUST_TYPE_STRING ") s(.Occurrence))")
     DEFINE_VIREO_FUNCTION(HttpClientPost, "p(io(.UInt32) i(.String) i(.String) i(.String) i(.Int32) o(.String) o(.String) o(.UInt32) io(" ERROR_CLUST_TYPE_STRING ") s(.Occurrence))")
+    DEFINE_VIREO_FUNCTION(HttpClientConfigCORS, "p(io(.UInt32) i(.Boolean) io(" ERROR_CLUST_TYPE_STRING "))")
```
cglzaguilar: I noticed that Vireo definition includes a Boolean parameter. Are you also fixing the DFIR side to pass the appropriate variable?

cglzaguilar:
@sanmut are you adding tests on a later submission?

source/io/HttpClient.cpp
sanmut: I was originally intending to. But, after thinking about it again, I think I should just fix this to use numeric and handle that issue separately to keep changes simpler in both layers. I will fix this.

sanmut:
@cglzaguilar yes, as mentioned in the comments.... will add later.

source/io/HttpClient.cpp
sanmut: Done and updated PR. Addressed all feedback.

source/io/HttpClient.cpp
```diff
@@ -527,6 +528,33 @@ VIREO_FUNCTION_SIGNATURE10(HttpClientPost, UInt32, StringRef, StringRef, StringR
 }
 
 //------------------------------------------------------------
+// handle(0), includeCredentialsDuringCORS(1), error cluster(2)
+VIREO_FUNCTION_SIGNATURE3(HttpClientConfigCORS, UInt32, UInt32, ErrorCluster)
+{
+#if kVireoOS_emscripten
+   // TODO mraj these checks are too aggressive. Should allow unwired values for optional terminals and avoid checking types inserted by Vireo
+   // Timeout(4) null value is handled by js and occurrence(9) handled by vireo
```
rajsite: This comment is outdated and should be removed

PR #249, opened 2017-04-21
Fixes to FormatDateTimeString to handle optional parameters (unwired …

cglzaguilar:
…parameters, default values)


test-it/ViaTests/DateFormatString.via
```diff
@@ -44,6 +44,10 @@ define(HelloTime dv(.VirtualInstrument (
 	Printf("11:(%s)\n" s1)
 	FormatDateTimeString(s1 "%Z"  t4 true)
 	Printf("12:(%s)\n" s1)
+	FormatDateTimeString(s1 * t1 true)
+	Printf("13:(%s)\n" s1)
+	FormatDateTimeString(s1 * * true) // Output is going to be different everytime, so we don't print the result
+	FormatDateTimeString(s1 * * *) // Same here, we want to exercise the function with 'default' values
```
spathiwa: Instead of not printing, yYou can make the output print with "//" at the beginning of the line and the test drivers will ignore the difference, but you can still see the output if you run the test manually.
Some of the tests do this to print elapsed time for tests, which varies per run but doesn't cause failure.

cglzaguilar: Done, I update the test.

cglzaguilar:
@sanmut I don't know. I'll take a look.

cglzaguilar:
@sanmut I'm mostly fixing optional parameters behavior, not actual functionality.

cglzaguilar:
With 07/02/2014, I still see differences between chrome and firefox:

For firefox:

        13:(14 (=2014) 00 (=12 AM) 00 00 .500)
        14:(07/02/2014 12:00:00.500 AM)
        15:(Wed Wednesday July July 02 183 07 26 3 26)
        16:(07/02/2014 12:00:00 AM)
        17:(-05:00:00)
        18:(CST)

For chrome:

        13:(14 (=2014) 00 (=12 AM) 00 00 .500)
        14:(07/02/2014 12:00:00.500 AM)
        15:(Wed Wednesday July July 02 183 07 26 3 26)
        16:(07/02/2014 12:00:00 AM)
        17:(-05:00:00)
        18:(CDT)

Notice that timezone is different.

There is something else going on that needs to be investigated and will consider that independent of this submission.

PR #252, opened 2017-04-25
Publicapi coverage

rajsite:
- Lots of JS tests for readJSON and writeJSON
- Modified TDViaParser to support quoteInfNan to be symmetrical with TDViaFormatter
- Added default print and printError behaviors (can still be overwritten with vireo.eggShell.setPrintFunction and vireo.eggShell.setPrintErrorFunction)


rajsite:
Yea that would be valuable if we want to validate more environments like C++, Python, LabVIEW, etc are consuming the EggShell_ReadJSON and EggShell_WriteJSON apis correctly.

Seems like each environment may still want a known in memory representation against ie a LabVIEW cluster, a C++ struct, etc.

Maybe there is a good approach that doesn't require encoding the result in per platform representation:
1. call ReadJSON put it through your platform JSON parser
2. read the results file, put it through your platform JSON parser
3. recursively compare every field and ensure equality

This means the JSON parser needs to create a dynamic structure that can be reflected on, couldn't use a static one like LV without VI scripting or creating the in memory representation of the tests (ruining the whole point of separating the results).

The Write JSON tests are also more involved. They currently read to validate the intial value. Write a new value. Read the new value to validate the new value. Then rewrite the old value and revalidate reading the old value. The Write JSON test is then repeated for different values compared to the original value (ie with larger arrays or smaller arrays).

I think you propose an interesting idea but need to experiment more with what would make a valuable generic tester for the api. Might want it to cover less (not having tests for different array write sizes) or maybe even have more tests.

Something to keep in mind for the future, especially if we want to test those APIs in more language environments.

rajsite:
Updated the tests to pass in PhatomJS 👍

cglzaguilar:
@sanmut do you have any comments on this changeset?

test-it/karma/utilities/TestHelpers.VireoRunner.js
```diff
@@ -84,13 +89,29 @@
                     return result;
                 }
             };
+        },
+        toMatchIEEE754Number: function (util, customEqualityTesters) {
+            return {
+                compare: function (actual, expected) {
+                    var result = {};
+                    if (typeof actual === 'number' && typeof expected === 'number') {
+                        result.pass = Object.is(actual, expected);
```
sanmut: Does "Object.is" expect the exact same value for floating point numbers? If so, we should use that only for special floating point numbers (such as NaN, Inf, +0, etc) and then use some epsilon comparison for regular floating point numbers. I wonder if we can use toBeCloseTo matcher in Jasmine?

test-it/karma/utilities/TestHelpers.VireoRunner.js
```diff
@@ -3,10 +3,14 @@
     window.testHelpers = window.testHelpers || {};
 
     var rebootAndLoadVia = function (vireo, viaAbsolutePath) {
-        expect(viaAbsolutePath).toBeNonEmptyString();
+        // Jasmine Matchers library is not always ready in beforeAll so use jasmine core functions
+        expect(typeof viaAbsolutePath).toBe('string');
+        expect(viaAbsolutePath).not.toBe('');
```
rajsite: What is your concern?

test-it/karma/fixtures/publicapi/MultipleTypes.via
```diff
@@ -0,0 +1,90 @@
+// Autogenerated Vireo assembly file
```
rajsite: Leftover comment, this was a heavily hand modified file

test-it/karma/publicapi/GetArrayDimLength.Test.js
```diff
@@ -0,0 +1,49 @@
...
+    'use strict';
+    // Reference aliases
+    var Vireo = window.NationalInstruments.Vireo.Vireo;
+    var vireoRunner = window.testHelpers.vireoRunner;
+    var fixtures = window.testHelpers.fixtures;
+
+    var vireo;
+
+    var testsArrayDemoViaUrl = fixtures.convertToAbsoluteFromViaTestsDir('ArrayDemo.via');
+    var publicApiArrayTypesOptimizedViaUrl = fixtures.convertToAbsoluteFromFixturesDir('publicapi/ArrayTypesOptimized.via');
```
rajsite: 👍

test-it/karma/fixtures/publicapi/MultipleTypes.via
sanmut: My question was not about that. I was wondering whether you were planning to checkin GVI tests along with the other existing GVI tests after this goes in on the Vireo side?

test-it/karma/utilities/TestHelpers.VireoRunner.js
sanmut: In our JS tests, we are using toBeNonEmptyString and toBeEmptyString and I thought that was the convention we wanted to follow !!??

test-it/karma/utilities/TestHelpers.VireoRunner.js
```diff
@@ -84,13 +89,29 @@
                     return result;
                 }
             };
+        },
+        toMatchIEEE754Number: function (util, customEqualityTesters) {
```
rajsite: This is not a function declaration, this is a function expression.

function declaration:

```
(function () {

    function mydeclaredFunction() {
    };

}());
```

function expression assigned to var:

```
(function () {

    var myFunctionExpression = function () {
    };

}()))
;
```

function expression assigned to property of object:

```
(function () {

    var myObj = {
        myFunctionExpression: function () {
        }
    };

}());
```

> Is labeled function declaration allowed in 'strict' mode in all browser environments?

[ECMAScript permits function statements in exactly two places: directly within the list of statements that make up a program, and directly within the list of statements that make up the contents of a function body](http://whereswalden.com/2011/01/24/new-es5-strict-mode-requirement-function-statements-not-at-top-level-of-a-program-or-function-are-prohibited/)

> Is there a different way to do this without labeled function declaration?

This is not a function declaration / function statement. But I agree that if it was, we shouldn't use that form.

test-it/karma/utilities/TestHelpers.VireoRunner.js
rajsite: I think this is sufficient

rajsite: We want to use the most precise test we can. Some possibilities for this check from least specific to most specific:
isDefined
isString
isNonEmptyString
toMatch(/regexp/)
toEqual
toBe


In places where we can validate the exact string like the statusText is "OK" for HTTP that is as specific as we can get.
The cookies we know a subset match the cookie string.
The paths can be anything besides a non-empty string. We could maybe use (/.via/i) or something but that's unnecessarily strict.

rajsite: Oh. I misunderstood.

If you read the comment it says jasmine matchers are not available yet. So we can't use toBeNonEmptyString in this file

rajsite: That is a bug in the jasmine matchers library that provides all those extra function on top of jasmine's standard matchers. So far it only effects utility files like this, it has not been a problem in the test suites yet.

test-it/karma/fixtures/publicapi/MultipleTypes.via
rajsite: Carlos has done that as a separate task for the user story

test-it/karma/utilities/TestHelpers.VireoRunner.js
sanmut: ah, I missed the comment just above :-( . never mind.

test-it/karma/utilities/TestHelpers.VireoRunner.js
sanmut: oh cool, thanks for the explanation.

test-it/karma/publicapi/ReadJson.Test.js
```diff
@@ -0,0 +1,640 @@
...
+        it('String', function () {
+            var actualJSON = vireo.eggShell.readJSON(viName, 'dataItem_String');
+            var actual = JSON.parse(actualJSON);
+            expect(actual).toBe('Hello');
+        });
+
+        it('Double', function () {
+            var actualJSON = vireo.eggShell.readJSON(viName, 'dataItem_NumericDouble');
+            var actual = JSON.parse(actualJSON);
+            expect(actual).toBe(123.456);
```
rajsite: This isn't the result of a mathematical operation. We are writing an IEEE754 value to memory and reading it back. We should expect exact representation serializing IEEE754 values in and out.

test-it/karma/utilities/TestHelpers.VireoRunner.js
rajsite: These tests are validating exact representations of values can be serialized and deserialized from Vireo memory. We should not be using epsilons for comparisons during these tests.

rajsite:
> Do you have work items created to track the open issues listed here?

You can see above that issue #254 was created to track Craig's suggestion. Did any others need to be made?

rajsite:
There was a lot of discussion so asking for re-review from @sanmut to make sure all concerns were addressed

PR #253, opened 2017-04-25
Add test for Config CORS API

sanmut:


test-it/karma/utilities/TestHelpers.HttpBinHelpers.js
```diff
@@ -45,15 +45,15 @@
         request.send();
     };
 
-    var forceHttpBinQuery = function (done) {
+    var forceHttpBinQuery = function (done, relativePath) {
```
rajsite: Please put the callback at the end to match the convention of other test helpers.
`var forceHttpBinQuery = function (relativePath, done)`

PR #255, opened 2017-04-26
Format Into String / Scan From String with error I/O args

spathiwa:


source/core/NumericString.cpp
```diff
@@ -257,83 +257,33 @@ void DefaultFormatCode(Int32 count, StaticTypeAndData arguments[], TempStackCStr
...
-    buffer->AppendCStr(formatSpecifierBuf);
-    buffer->AppendCStr("' in '");
-    IntIndex length = format->Length()*2;
-    Utf8Char* formatUnEscaped = new Utf8Char[length];
-    IntIndex newLength = format->UnEscape(formatUnEscaped, length);
-    buffer->AppendUtf8Str(formatUnEscaped, newLength);
-    buffer->AppendCStr("'\n");
-    delete[] formatUnEscaped;
+// Format Into String/Scan From String (StringFormat/StringCan) error codes
+enum { kFormatArgErr=1, kFormatTypeMismatch = 81, kFormatCodeUnknown, kFormatTooManyFormatSpecs, kFormatTooFewFormatSpecs, kFormatScanFailed };
```
cglzaguilar: I'm curious as to how the error list was obtained and wonder if we didn't miss one.

source/core/NumericString.cpp
```diff
@@ -257,83 +257,33 @@ void DefaultFormatCode(Int32 count, StaticTypeAndData arguments[], TempStackCStr
...
+
+static Boolean SetFormatError(Int32 errCode, Int32 argNum, char formatCode, ErrorCluster *errPtr) {
```
cglzaguilar: Do we need to consider call chain information when generating the error?

source/core/NumericString.cpp
```diff
@@ -257,83 +257,33 @@ void DefaultFormatCode(Int32 count, StaticTypeAndData arguments[], TempStackCStr
...
+    if (errPtr) {
+        char argBuf[8];
+        snprintf(argBuf, sizeof(argBuf), "%d", argNum+1);
+        if (!errPtr->status)
+            errPtr->SetError(true, errCode, errCode == kFormatScanFailed ? "Scan From String" : "Format Into String");
+        if (errCode != kFormatTooManyFormatSpecs && errCode != kFormatTooFewFormatSpecs) {
```
cglzaguilar: If (error->status) was already set, do we still want to append all this error information? In the StringFormat and Scan function if the error->status on input is already on are we skipping any work?

source/core/NumericString.cpp
```diff
@@ -397,18 +347,18 @@ void TruncateLeadingZerosFromTimeString(StringRef buffer)
     }
 }
 
-void CreateMismatchedFormatSpecifierError(SubString* format, Int32 count, StaticTypeAndData* arguments, StringRef buffer, FormatOptions fOptions, Boolean* validFormatString, Boolean* parseFinished)
+void CreateMismatchedFormatSpecifierError(SubString* format, Int32 count, StaticTypeAndData* arguments, StringRef buffer, FormatOptions fOptions, Boolean* validFormatString, Boolean* parseFinished, ErrorCluster *errPtr)
```
cglzaguilar: Do we still need to pass buffer? It is not used in the new code and it is just resized to 0 length.

source/core/NumericString.cpp
```diff
@@ -1868,7 +1829,31 @@ VIREO_FUNCTION_SIGNATUREV(StringFormat, StringFormatStruct)
...
+struct StringFormatWithErrStruct : public VarArgInstruction
+{
+    _ParamDef(StringRef, StringOut);
+    _ParamDef(StringRef, StringFormat);
+    _ParamDef(ErrorCluster, ErrClust);
+    _ParamImmediateDef(StaticTypeAndData, argument1[1]);
+    NEXT_INSTRUCTION_METHODV()
+};
+
+VIREO_FUNCTION_SIGNATUREV(StringFormatWithErr, StringFormatWithErrStruct) {
```
cglzaguilar: The code appears to mostly a duplicate of StringFormat. Please create a function that both StringFormat and StringFormatWithErr call so they share code.

source/core/NumericString.cpp
```diff
@@ -2687,8 +2712,10 @@ VIREO_FUNCTION_SIGNATURE4(SpreadsheetStringtoArray, StringRef, StringRef, String
 DEFINE_VIREO_BEGIN(NumericString)
     DEFINE_VIREO_REQUIRE(Timestamp)
     DEFINE_VIREO_FUNCTION(StringFormatValue, "p(o(String) i(String) i(StaticTypeAndData))")
+    DEFINE_VIREO_FUNCTION_CUSTOM(StringFormat, StringFormatWithErr, "p(i(VarArgCount) o(String)   i(String) io(ErrorClust err) i(StaticTypeAndData))")
```
cglzaguilar: Remove support for old functions that didn't take an error cluster parameter. They are not very useful if you don't know what went wrong when something fails.

source/core/ExecutionContext.cpp
```diff
@@ -436,7 +436,7 @@ void ExecutionContext::IsrEnqueue(QueueElt* elt)
 DEFINE_VIREO_BEGIN(Execution)
     DEFINE_VIREO_REQUIRE(VirtualInstrument)
     DEFINE_VIREO_FUNCTION(FPSync, "p(i(String))")
-    DEFINE_VIREO_FUNCTION(CurrentBrowserFPS, "p(o(Double) io(" ERROR_CLUST_TYPE_STRING "))")
+    DEFINE_VIREO_FUNCTION(CurrentBrowserFPS, "p(o(Double) io(ErrorClust))")
```
cglzaguilar: Can we use ErrorCluster instead?

cglzaguilar:
@rajsite, please look at the formatting of the error code. We are using <Append>, but I wonder if everything else looks correct. Perhaps we should create a utility function that takes care of this.

source/core/NumericString.cpp
spathiwa: They are the only error codes generated by the NCG run-time code BytestreamDoFormat and ByteStreamDoScan.

source/core/NumericString.cpp
spathiwa: Yes, at least the leaf VI to match CG/NCG behavior.

source/core/NumericString.cpp
spathiwa: The top-level functions should not be called if an error comes in, so this helper function does not need to check for it.  String Format and Scan functions are supposed to do nothing and pass through the error if status is already set, and they do.

source/core/NumericString.cpp
spathiwa: There are several callers, all of which would want to zero Buffer, so might as well let it be common here.

source/core/NumericString.cpp
spathiwa: This code is hard to share because both functions take structure arguments, with a difference in an interior field.  If it were refactored, most of the code would still be pulling fields out of the structure to pass to a small common function.  Since the non-error I/O version of this is expected to be removed as soon as DFIR is refactored not to generate it, I think this is not worth doing.

source/core/NumericString.cpp
spathiwa: I plan to do this in a future commit, after DFIR supports the new error I/O versions.

source/core/ExecutionContext.cpp
spathiwa: Are you asking to rename the type itself in Vireo?
Are you already generating this type by name in DFIR anywhere?

source/core/NumericString.cpp
spathiwa: Fixed.

source/core/NumericString.cpp
spathiwa: Fixed.

spathiwa:
Carlos, I've commented on your requests and fixed the ones I agreed with that aren't intended to be done in a later PR.

source/core/NumericString.cpp
sanmut: NG provides the full call-chain by default in Desktop. I think WebVIs should match NG and not CG.

source/core/NumericString.cpp
```diff
@@ -257,83 +257,34 @@ void DefaultFormatCode(Int32 count, StaticTypeAndData arguments[], TempStackCStr
...
+            errPtr->source->AppendCStr(")");
+        }
+        errPtr->source->AppendCStr(" in ");
+        AppendCallChainString(errPtr->source);
+        if (formatCode && (errCode == kFormatCodeUnknown || errCode == kFormatTypeMismatch)) {
+            errPtr->source->AppendCStr("\n<APPEND>\n");
+            if (errCode == kFormatCodeUnknown) {
+                errPtr->source->AppendCStr("Invalid format specifier: ");
+            } else if (errCode == kFormatTypeMismatch) {
+                errPtr->source->AppendCStr("Format specifier type mismatch: ");
```
sanmut: We should put all the strings in a separate file to make localization easier in the future.

source/core/NumericString.cpp
```diff
@@ -257,83 +257,34 @@ void DefaultFormatCode(Int32 count, StaticTypeAndData arguments[], TempStackCStr
...
-    IntIndex length = format->Length()*2;
-    Utf8Char* formatUnEscaped = new Utf8Char[length];
-    IntIndex newLength = format->UnEscape(formatUnEscaped, length);
-    buffer->AppendUtf8Str(formatUnEscaped, newLength);
-    buffer->AppendCStr("'\n");
-    delete[] formatUnEscaped;
+// Format Into String/Scan From String (StringFormat/StringCan) error codes
+enum { kFormatArgErr=1, kFormatTypeMismatch = 81, kFormatCodeUnknown, kFormatTooManyFormatSpecs, kFormatTooFewFormatSpecs, kFormatScanFailed };
+
+static Boolean SetFormatError(Int32 errCode, Int32 argNum, char formatCode, ErrorCluster *errPtr) {
```
sanmut: Good to see multiple functions becoming one  :-) :+1:

source/core/NumericString.cpp
```diff
@@ -956,7 +909,8 @@ void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], Strin
     if (argumentIndex < count) {
         // make sure at least all of the args are used (even if out of order)
         if (usedArguments != count) {
-            ErrFormatExtraNumArgs(format, count - argumentIndex, arguments, buffer);
+            SetFormatError(kFormatTooFewFormatSpecs, count - argumentIndex, fOptions.FormatChar, errPtr);
```
sanmut: Not sure whether we would run into this. If number of arguments don't match number of format specifiers, it is a type-prop error in LabVIEW.

source/core/NumericString.cpp
sanmut: I am confused. Why are we adding a new Vireo instruction with ErrorStruct but why not modify the existing one? When the next version of Vireo is used to build LabVIEW, the DFIR should be modified to generate correctly. I think that's how we have been doing and I would rather not do work twice (second time to just delete functions). "npm link" makes it easier to build LV against local builds of Vireo. So, @cglzaguilar  can just develop DFIR side of this in parallel and get the DFIR part working and give even feedback.

source/core/NumericString.cpp
```diff
@@ -1986,15 +1972,55 @@ VIREO_FUNCTION_SIGNATUREV(StringScan, StringScanStruct)
...
+    _ParamDef(StringRef, StringRemaining);
+    _ParamDef(StringRef, StringFormat);
+    _ParamDef(UInt32, InitialPos);
+    _ParamDef(UInt32, OffsetPast);
+    _ParamDef(ErrorCluster, ErrClust);
+    _ParamImmediateDef(StaticTypeAndData, argument1[1]);
+    NEXT_INSTRUCTION_METHODV()
+};
+//------------------------------------------------------------
+VIREO_FUNCTION_SIGNATUREV(StringScanWithErr, StringScanStructWithErr)
```
sanmut: Same as before. We should just modify the existing function.

test-it/ViaTests/ScanFormatErr.via
```diff
@@ -0,0 +1,220 @@
+// Test scan from string
```
sanmut: Nice to see more tests :+1:

sanmut: We should create VI tests in Desktop for different error cases (with a subVI call) and validate those with this PR to make sure they all pass.

source/core/NumericString.cpp
spathiwa: Only if the format string is a constant, otherwise it's a run-time error.

source/core/NumericString.cpp
spathiwa: Uhm, okay, but that's a separate refactor of all strings in Vireo and not something I'm gonna do here.

source/core/NumericString.cpp
spathiwa: I guess you didn't read Carlos' comments or my replies.
Go do so.

source/core/NumericString.cpp
spathiwa: For some reason, github won't let me reply to your previous comment about call-chain, so I'll reply here:
It appears to be the convention that prims report a leaf-VI only call chain, and VIs report the full call chain.  Other prims in Vireo also only include leaf-VI callchain.

source/core/NumericString.cpp
spathiwa: I'm going to do this in a separate PR because I like to do things incrementally.
Deleting two prims is not "doing work twice", and I'm doing it, not you.

sanmut: As long as we do it "soon" and don't accrue tech debt, it is ok. Can you create a issue for this to track this and leave a TODO in the code against two prims?

source/core/NumericString.cpp
sanmut: You are right :-)

source/core/NumericString.cpp
spathiwa: Done.  Issue #256.

source/dox/Internals/API.cpp
```diff
@@ -0,0 +1,88 @@
+/**
```
rajsite: Why do we have API.md and API.cpp? Were these changes intended to go in API.md?

spathiwa: Oops.  This was just an  experiment because doxygen won't process directives in .md files.
I did not intend to add it to this PR, but it's also stupid that 'dox' is in the 'source' directory.
I did 'git add source' and forgot that was there.

source/core/NumericString.cpp
```diff
@@ -257,83 +257,34 @@ void DefaultFormatCode(Int32 count, StaticTypeAndData arguments[], TempStackCStr
...
+        errPtr->SetError(true, errCode, errCode == kFormatScanFailed ? "Scan From String" : "Format Into String");
+        if (errCode != kFormatTooManyFormatSpecs && errCode != kFormatTooFewFormatSpecs) {
+            errPtr->source->AppendCStr(" (arg ");
+            errPtr->source->AppendCStr(argBuf);
+            errPtr->source->AppendCStr(")");
+        }
+        errPtr->source->AppendCStr(" in ");
+        AppendCallChainString(errPtr->source);
+        if (formatCode && (errCode == kFormatCodeUnknown || errCode == kFormatTypeMismatch)) {
+            errPtr->source->AppendCStr("\n<APPEND>\n");
```
rajsite: Can we make a helper function for adding additional information to a source string? I imagine this will be used by multiple functions and it's possible it will change.

Also the format we had worked out only has a` \n` after the `<APPEND>` tag and not before:

https://github.com/ni/VireoSDK/blob/049bf0c3a2b3b9840d0a9ea67ffb9d2d2d07cb7c/source/io/module_httpClient.js#L57

spathiwa: Sure, I'll add one, but it'll really just be a helper for the APPEND tag, because having it take a string to add after the append would only cause us to always have to make an extra copy of the string if it needed any formatting.

spathiwa: Done.

source/core/ExecutionContext.cpp
rajsite: +1 for naming it ErrorCluster if we can

rajsite:
I saw that cpplint.py was marked executable :) did you give it a run? how bad was it?

spathiwa:
@rajsite Pretty bad; there were about 3700 warnings, but almost all of the issues were about line length (with limit set to Paul's recommendation of 120), space usage (at the of line, around braces, inside parens, before // comments, presence of tabs, etc.), and C-style tabs.

source/core/ExecutionContext.cpp
spathiwa: OK, I renamed it.  @cglzaguilar Please make sure to update DFIR if you are generating refs to this type.  I'm not sure if you are yet; all I've seen is replications of the whole cluster type.

spathiwa:
@cglzaguilar Timeout.

PR #258, opened 2017-05-01
Fix error code generation for JSON and queue prims

spathiwa:


source/include/DataTypes.h
```diff
@@ -132,14 +132,31 @@ typedef enum {
...
+    kJSONLV_StrictFieldNotFound = -375007,
+    kJSONLV_ClusterElemNotFound = -375006,
+    kJSONLV_TypeMismatch = -375005,
+    kJSONLV_InvalidPath = -375004,
+    kJSONLV_InvalidString = -375003,
+    // The rest are semantic analysis errors so can never be seen in LV-generated code: TODO (spathiwa) Implement for benefit of hand-written Vireo
+    kJSONLV_InvalidElementNameError = -375002,
+    kJSONLV_InvalidRootContainerError = -375001,
+    kJSONLV_UnsupportedTypeError = -375000
+} JSONLVError;
```
cglzaguilar: I noticed we have different naming conventions for errors, kNIError_, kLVError_, kJSONLV_ and kQueueArgErr. I suggest we use kLVError_.

source/core/TDCodecVia.cpp
```diff
@@ -2206,10 +2296,11 @@ VIREO_FUNCTION_SIGNATUREV(FlattenToJSON, FlattenToJSONParamBlock)
     } else {
         formatter.FormatData(arg[0]._paramType, arg[0]._pData);
     }
-    if (formatter.HasError()) {
+    Int32 errCode = formatter.GetError();
+    if (errCode) {
```
cglzaguilar: I don't think we are properly adding call chain information or the append string as in other calls.

source/core/TDCodecVia.cpp
```diff
@@ -2272,28 +2363,20 @@ VIREO_FUNCTION_SIGNATUREV(UnflattenFromJSON, UnflattenFromJSONParamBlock) // Str
...
-                    badNulls = true;
             }
             arg[0]._paramType->ClearData(buffer);
             delete[] buffer;
         }
     }
-    if (_ParamVarArgCount() > 7) {
-        Int32 code = badPath ? kJSONLV_InvalidPath : badNulls ? kJSONLV_TypeMismatch :  kJSONLV_InvalidString;
-        if (badPath || badNulls || log.HardErrorCount() > 0) {
+    if (_ParamVarArgCount() > 7) { // error I/O wired
```
cglzaguilar: I don't think we are properly adding call chain information or the append string as in other calls.

source/core/Synchronization.cpp
```diff
@@ -765,7 +765,7 @@ VIREO_FUNCTION_SIGNATURE6(QueueRef_Obtain, RefNumValType*, Int32, StringRef, Boo
         if (createdPtr)
             *createdPtr = false;
         if (errPtr)
-            errPtr->SetError(true, errCode, "ObtainQueue");
+            errPtr->SetError(true, errCode, "ObtainQueue", true);
```
cglzaguilar: Are we generating the Append string as in the FormatToString case?

source/include/TypeAndDataManager.h
```diff
@@ -1198,14 +1198,7 @@ struct ErrorCluster {
...
 
-    void SetError(Boolean s, Int32 c, ConstCStr str) {
-        status = s;
-        code = c;
-        if (source) {
-            source->Resize1D(0);
-            source->AppendCStr(str);
-        }
-    }
+    void SetError(Boolean s, Int32 c, ConstCStr str, Boolean appendCallChain = false);
```
spathiwa: Because sometimes you want to append other info before adding the call chain.
Changing the default might be reasonable though.

source/core/Synchronization.cpp
spathiwa: I don't understand the question re: something being handled in the SetError function.  Format Into String needs to add "(arg N)" between the prim name and the call chain, so it has to be slightly different.  But most prims can use SetError.

source/core/StringUtilities.cpp
```diff
@@ -902,14 +916,23 @@ Boolean SubString::ParseDouble(Double *pValue, Boolean suppressInfNaN /*= false*
             if (length >= 9 && ComparePrefixIgnoreCase(Utf8Ptr("inity"), 5))
                 _begin += 5;
         }
+        if (bParsed && suppressInfNaN) {
```
spathiwa: Nope.  Expand the elided code.  bParsed can be set to true again between the outer if and this test.

source/include/DataTypes.h
spathiwa: I agree.  But  I didn't name the JSON errors, I just moved them.  Are you suggesting I rename them as part of this PR?

source/core/TDCodecVia.cpp
spathiwa: Fixed.

source/include/TypeAndDataManager.h
spathiwa: Default value changed.

source/include/DataTypes.h
spathiwa: I went ahead and did this.

source/core/Math.cpp
```diff
@@ -294,8 +294,8 @@ DECLARE_SCALE2X_INTN_HELPER(Single)
     VIREO_FUNCTION_SIGNATURE7(InRangeAndCoerce##TYPE, TYPE, TYPE, TYPE, Boolean, Boolean, TYPE, Boolean) { \
```
spathiwa: Done, but I only touched this code to fix warnings, the PR otherwise has nothing to do with it.

source/core/TDCodecVia.cpp
spathiwa: Fixed.

source/core/StringUtilities.cpp
sanmut: ah never mind. That elision on the left side tricked me :-(

source/core/Synchronization.cpp
```diff
@@ -823,12 +823,12 @@ VIREO_FUNCTION_SIGNATURE4(QueueRef_Release, RefNumValType*, StringRef, TypedArra
         }
         else if (QueueRefNumManager::RefNumStorage().DisposeRefNum(refnumPtr->GetRefNum(), &queueRef) != kNIError_Success) {
             if (errPtr && !errPtr->status)
-                errPtr->SetError(true, kQueueArgErr, "ReleaseQueue");
+                errPtr->SetError(true, kQueueArgErr, "ReleaseQueue", true);
```
sanmut: Please remove all the 'true' at all the call-sites since the default is changed to 'true' now.

spathiwa: I don't think there's anything wrong with passing the parameter explicitly.

sanmut: Then, I would rather remove the "defaultness" of the argument.

source/include/DataTypes.h
cglzaguilar: Yes, I was suggesting to rename them.

spathiwa: As I said, I already did.

spathiwa:
Carlos, are you waiting on something else in particular before approving this?

cglzaguilar:
Yes, I have one last thing. Basically there is a function called SetFormatError, that is calling AddAppendixPreamble. I don't see AddAppendixPreamble being called for the Queue primitives, and the Unflatten\Flatten JSON primitives. I think we need too.

spathiwa:
@cglzaguilar  These prims have nothing to append.  That's for supplemental information, but these prims only have the prim name and callstack.

spathiwa:
@cglzaguilar approved the changes in person after discussion.

PR #261, opened 2017-05-05
Rely on shx shell expansion

rajsite:
Quote the strings to rely on shx internal shell expansion
https://github.com/shelljs/shelljs/issues/711#issuecomment-298490681


rajsite:
Reviewed with Carlos

PR #262, opened 2017-05-05
More format/scan/json error fixes and defect fixes for static VI resource limits and polymorphic prims in loops

spathiwa:


source/core/NumericString.cpp
```diff
@@ -224,11 +224,14 @@ void DefaultFormatCode(Int32 count, StaticTypeAndData arguments[], TempStackCStr
         }
         switch (argType->BitEncoding()) {
 
-        case kEncoding_UInt:
-        case kEncoding_Enum: {
+        case kEncoding_UInt: {
             buffer->AppendCStr("%u");
         }
         break;
+            case kEncoding_Enum: {
```
cglzaguilar: Remove an extra tab, so the new case aligns with others. We need to consider adding linting sooner rather than later so this is found automatically for us.

source/core/NumericString.cpp
```diff
@@ -1660,7 +1671,11 @@ Int32 FormatScan(SubString *input, SubString *format, Int32 argCount, StaticType
                         input->AliasAssign(input->Begin()+endPointer, input->End());
                         offsetPastScan += endPointer;
                     } else {
-                        SetFormatError(kFormatScanFailed, argumentIndex, fOptions.FormatChar, errPtr);
+                        TypeRef argType = pArg->_paramType;
```
cglzaguilar: On the previous case for 'b' and 'B' and think an error case is necessary.

source/core/NumericString.cpp
spathiwa: It lines up in the actual code and that line wasn't touched by me, so I didn't notice; it doesn't here because of the + added by the differ.
But since you ask, I'm going to go ahead and remove ALL the extraneous tabs in the code.

source/core/VirtualInstrument.cpp
```diff
@@ -291,11 +291,17 @@ void ClumpParseState::Construct(VIClump* clump, InstructionAllocator *cia, Int32
 
     _totalInstructionCount = 0;
     _totalInstructionPointerCount = 0;
-    
+
+    _argPointers.reserve(kClumpStatencrementSize);
```
cglzaguilar: Typo? Change it to kClumpStateIncrementSize?

source/core/NumericString.cpp
spathiwa: There are 60 files with whitespace problems, so I'm going to do this in a different PR.

source/core/VirtualInstrument.cpp
```diff
@@ -1083,42 +1090,44 @@ InstructionCore* ClumpParseState::EmitInstruction()
...
     if (_cia->IsCalculatePass())
         return instruction;
     
-    if (_recordNextInstructionAddress>=0) {
+    if (_perchIndexToRecordNextInstrAddr>=0) {
         // TODO support multiple perch patching
-        VIREO_ASSERT( _perches[_recordNextInstructionAddress] == kPerchBeingAlocated);
-        _perches[_recordNextInstructionAddress] = instruction;
-        _recordNextInstructionAddress = -1;
+        VIREO_ASSERT(_perches[_perchIndexToRecordNextInstrAddr] == kPerchBeingAlocated);
```
cglzaguilar: Fix type on kPerchBeingAlocated -> kPerchBeingAllocated

source/include/VirtualInstrument.h
```diff
@@ -246,17 +248,17 @@ class ClumpParseState
...
-    TypeRef         _argTypes[kMaxArguments];
-    
-    Int32           _argPatches[kMaxArguments];     // Arguments that need patching
+    std::vector<void*> _argPointers;
+    std::vector<TypeRef> _argTypes;
+
     Int32           _argPatchCount;
-    
-    PatchInfo       _patchInfos[kMaxPatchInfos];
+    std::vector<Int32> _argPatches;    // Arguments that need patching
```
cglzaguilar: Use PatchInfo instead of Int32?

source/core/NumericString.cpp
spathiwa: Never mind, went ahead and added it to this one because there were too many conflicts to resolve rebasing it against incoming.

source/core/VirtualInstrument.cpp
spathiwa: Oops.  Fixed.

source/core/VirtualInstrument.cpp
spathiwa: That one was Paul's, but ok! :)

source/include/VirtualInstrument.h
spathiwa: Nope, this is correct.  In addition to refactoring to use vectors, I moved some of the counts around to always be declared immediately before the vector they went with, so this might've caused the diff to confuse you.

spathiwa:
All changes requested by @cglzaguilar  done.

source/core/VirtualInstrument.cpp
spathiwa: Also, you have a typo in 'typo'.  😄

source/core/NumericString.cpp
spathiwa: Done, consolidated with other numeric cases.

source/include/VirtualInstrument.h
```diff
@@ -246,17 +248,17 @@ class ClumpParseState
     InstructionAllocator* _cia;
     
     Int32           _argCount;
-    void*           _argPointers[kMaxArguments];
-    TypeRef         _argTypes[kMaxArguments];
-    
-    Int32           _argPatches[kMaxArguments];     // Arguments that need patching
+    std::vector<void*> _argPointers;
```
sanmut: I have scheduled a meeting and discuss before using std::vector. I know there are a couple of  places vector is used already but before we proliferate the use of vector's we should talk. I think this commit is solving a separate defect by itself. Please this commit out of PR and submit it as a separate PR after we have a resolution on use of std::vector.

test-it/ViaTests/PolymorphicLoopBug.via
```diff
@@ -0,0 +1,119 @@
+// Autogenerated Vireo assembly file
```
sanmut: Does this test pass on Windows native?

test-it/ViaTests/NestedStructuresBug.via
```diff
@@ -0,0 +1,703 @@
+// Autogenerated Vireo assembly file
```
sanmut: Does this test pass with Windows native?

test-it/ViaTests/PolymorphicLoopBug.via
spathiwa: There's nothing about the bug fix that has any potential to be platform-specific, so it should, but I did not test it.   I'm not going to do extra work to build and test it manually when we're already working on a separate Windows CI.

test-it/ViaTests/NestedStructuresBug.via
spathiwa: See other comment.

source/include/VirtualInstrument.h
spathiwa: FYI, we already use not only stl::vector, but stl::map for the type dictionary itself (and the alternate #ifdef for support without this doesn't actually work), and vector is more light-weight than vector.

spathiwa: It is difficult to extricate this commit from the rest because of the other variable renames and spacing fixes made.  (I tried, and after the rebase failing 4 times requiring manual edits to fix up, I gave up).   It we decide not to use stl, it would likely be an an alternate implementation behind an #if/#else anyway, or our own implementation that mimics the STL API, so this code would remain.  Therefore, I'm inclined to keep it in this PR.

cglzaguilar: I'm OK on leaving this code for the pull request. If we decide we want to do something else, we can remove it later. I'm more concerned about the tests not passing on windows.

test-it/ViaTests/NestedStructuresBug.via
cglzaguilar: Yes, it does pass on WIndows native.

test-it/ViaTests/PolymorphicLoopBug.via
cglzaguilar: Yes, it does pass on Windows native.

PR #263, opened 2017-05-08
Fix win32 alignment type bug

spathiwa:


spathiwa:
This fixes a slew of win32 native test errors.

PR #265, opened 2017-05-08
Perform --ff-only merge to validate emsdk

rajsite:
Noticed in a failed build the following output

```
fatal: destination path 'emscripten-sdk' already exists and is not an empty directory.
remote: Counting objects: 5, done.
remote: Compressing objects: 100% (1/1), done.
remote: Total 5 (delta 4), reused 5 (delta 4), pack-reused 0
Unpacking objects: 100% (5/5), done.
From https://github.com/ni/VireoSDK
   5436b38..e5e25c3  incoming   -> origin/incoming
Merge branch 'incoming' of https://github.com/ni/VireoSDK into HEAD
# Please enter a commit message to explain why this merge is necessary,
# especially if it merges an updated upstream into a topic branch.
#
# Lines starting with '#' will be ignored, and an empty message aborts
# the commit.
~                                                                               
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

Realized the script for validating the cache was actually doing the pull against ni/VireoSDK and not ni-kismet/emscripten-sdk

Modified the script to actually validate emscripten-sdk and to do so by attempting a fast forward merge. Best case changes to emscripten-sdk are correctly picked up and worst case it fails and causes cache eviction.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ install:
 - npm install
 - pip install --user -r requirements.txt
 - if [ "$CXX" = "g++" ]; then export CXX="g++-4.9" CC="gcc-4.9"; fi
-- if ! git clone --depth 1 https://github.com/ni-kismet/emscripten-sdk.git; then git pull; fi
+- if ! git clone --depth 1 https://github.com/ni-kismet/emscripten-sdk.git; then cd emscripten-sdk && git remote update -p && git merge --ff-only origin/master && cd ..; fi
 - emscripten-sdk/emsdk activate --build=Release sdk-master-64bit
 - source emscripten-sdk/emsdk_env.sh
 - cat ~/.emscripten
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ The `esh` and `vireo.js` files will be put into the `./dist` directory.
 $ make native
 ```
 
-#### Build Javascript File
+#### Build JavaScript File
 ```shell
 $ make js
 ```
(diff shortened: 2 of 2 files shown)


rajsite:
Caching running how I expected, ready for review

spathiwa:
It seems to me like you just need the 'cd emscripten-sdk' part and could keep the 'git pull'.
I guess I may not understand what's going on here.  What possible local changes could there be to merge into?

rajsite:
Like the error in the description above if the git pull finds changes it interactively asks for a commit message blocking the build and times out after ten minutes. The fast forward merge should fail much quicker if there are inconsistencies that cause the cache to not be able to fast forward.

> I guess I may not understand what's going on here. What possible local changes could there be to merge into?

There shouldn't be any, but since these are coming out of the cache system I wanted git to verify the state of those cached files for me and make sure they have the latest changes (which is what the clone effectively used to do). If this fails for any reason, ie can't fast forward or another git error, build should fail out so the cache is evicted and we can re-examine the dependencies.

If the fast-forward merge is successful at pulling in new files, that's great, it means we updated that repo to use a new version and it was successfully pulled in. Github will update and store the new cache version at the end of the build.

rajsite:
removed @sanmut from review because he has more important things to tend to :)

PR #266, opened 2017-05-10
Fix many cpplint warnings.

spathiwa:


source/core/RefNum.cpp
```diff
@@ -69,7 +69,7 @@ RefNumStorageBase::RefNumHeaderAndData* RefNumStorageBase::CreateRefNumIndex(Ref
 {
     bool isNew;
     UInt32 index = static_cast<UInt32>(IndexFromRefNum(refnum));
-    RefNumHeaderAndData* rnp = CreateRefNumIndexT(index, isNew);
```
cglzaguilar: This seems to be a bug, do we have a test that covers it?

cglzaguilar: And my next topic, once cpplint is working, would be to focus on code coverage.

spathiwa: No, it wasn't a bug, lint recommended changing a non-const reference param to a pointer, so I did.

spathiwa: And this code is covered by any queue operation.

source/core/TypeAndDataReflection.cpp
```diff
@@ -27,13 +27,13 @@ namespace Vireo
 
 class DataReflectionVisitor : public TypeVisitor
 {
-public:
```
cglzaguilar: Do we want a single space or four spaces? if we want four spaces then we need to move most of the file contents to the right.

source/core/VirtualInstrument.cpp
```diff
@@ -34,7 +34,7 @@ NIError VirtualInstrument::Init(TypeManagerRef tm, Int32 clumpCount, TypeRef par
     _locals->SetElementType(localsType, false);
     _clumps->Resize1D(clumpCount);
     _lineNumberBase = lineNumberBase;
-    _clumpSource = clumpSource;
```
cglzaguilar: I'm confused, this seems something that the compiler should catch. Are we ignoring warnings?

source/core/TypeAndDataReflection.cpp
spathiwa: Google style guides and cpplint say private, protected, and public: labels should be indented by exactly one space, even though the recommended indentation for other code is 4 spaces.

source/core/VirtualInstrument.cpp
spathiwa: This change was necessary because of adding explicit to single arg constructors.
Before that, it appears the compiler was automatically handling it.   There were only 3 instances where this showed up.

source/core/RefNum.cpp
cglzaguilar: Excellent!

source/core/TypeAndDataReflection.cpp
cglzaguilar: I'm Ok with that recommendation.

source/core/VirtualInstrument.cpp
cglzaguilar: Sound good to me.

spathiwa:
I'm going to take a pass at fixing a couple of the warnings I turned off, and then put it in the CI.

cglzaguilar:
Ok to merge after ensuring the windows tests continue to pass.

source/core/VirtualInstrument.cpp
spathiwa: To clarify, it's because there was a constructor which took a SubString pointer, but no operator=.  The compiler was synthesizing an operator= out of the constructor, but with the explicit keyword added, it would no longer do so.

PR #267, opened 2017-05-10
Fix lint errors for TODO usage, avoiding using namespace,  and header…

spathiwa:
…/include-what-you-use

We now have zero lint errors except for the a couple of disabled ones.

The only lint errors still disabled are: enforcing open braces at end of
line, the format of the #ifdef include guard name, and C-style casts.
We also allow line to be up to 160 chars rather than 80.
See the comment at the top of make-it/cpplint/Makefile.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,8 @@ before_script:
 - nohup python -m httpbin.core --port 64526 &
 - while ! nc -q 1 127.0.0.1 64526 </dev/null; do sleep 2; done
 - curl http://127.0.0.1:64526/get?show_env=1
-- make
 - make lint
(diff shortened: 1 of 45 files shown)


.travis.yml
```diff
@@ -36,6 +36,7 @@ before_script:
 - while ! nc -q 1 127.0.0.1 64526 </dev/null; do sleep 2; done
 - curl http://127.0.0.1:64526/get?show_env=1
 - make
+- make lint
```
rajsite: maybe do the lint before `make` to fail quickly? Seems more likely someone will forget to lint before submitting that forget to build.

PR #273, opened 2017-05-15
Appveyor emsdk integration

rajsite:
Adding @spathiwa primarily for Makefile changes, see below for why mkdir is quoted.

The commits should be squashed with the following as the new commit comment

-----------

This set of changes enables the emscripten builds on appveyor in order to test against IE and Edge

Note:
The following tests were disabled in the browser due to failures:
DateFormatStringLocaleSpecific.via fails in Firefox on Windows (passes all others)
Round.via, Scale2X.via, and StringFormatComplex.via fails in IE11 and Edge (Edge run locally, can't run on CI). It looks like the failure is related to "-NaN" values on those browsers.

Note: AppVeyor is using emscripten 1.37.9, while Travis is using emscripten 1.36.0. Getting Travis on the same version will be in a separate commit.

For downloading the emsdk I tried using webclient as suggested in the AppVeyor docs but it would not show up in the directory it was run from:
(new-object net.webclient).DownloadFile('https://s3.amazonaws.com/mozilla-games/emscripten/releases/emsdk-1.35.0-portable-64bit.zip', '.\emsdk-1.35.0-portable-64bit.zip')

Ended up switching to Bits instead:
Start-BitsTransfer -Source https://s3.amazonaws.com/mozilla-games/emscripten/releases/emsdk-1.35.0-portable-64bit.zip -Destination .\emsdk-1.35.0-portable-64bit.zip

Modern versions of powershell have built-in archive extraction:
Expand-Archive .\emsdk-1.35.0-portable-64bit.zip -DestinationPath .\emscripten-sdk

Was getting the "/usr/bin/env: python2: No such file or directory" error running emcc which is bizarre because it is a Windows build.
I also get that error when using the Github Desktop command shell locally which adds a bunch of unix tools built for Windows to the path.
All Github issues I found were related to having the problem on Mac or Linux, but some also mentioned the issue being fixed in newer releases of Emscripten.
As newer releases of emsdk for Windows were out, switched to emsdk version 1.37.9 and the error stopped both locally in the Github Desktop command shell and on the AppVeyor CI.

Went back and forth on using the gnuwin32 utils. One issue was that mkdir didn't seem to work from Makefiles.
Turns out that cmd and powershell have their own mkdir shell commands that overrode the mkdir I added to the path.
Found that by quoting mkdir in the Makefile I can get it to evaluate the one on the path for Windows and Linux
"mkdir" -p $(OBJS)

I had trouble on AppVeyor using the "start" command from cmd as it looked like it still blocked the original shell. Instead I used Start-Process from powershell on AppVeyor and that seemed to work
ps: Start-Process npm "run httpbin" -PassThru
Note that Start-Process requires a process to run and takes a string of command-line arguments to pass instead of taking a commandline to run in a new process.

Turns out Windows Server 2016 does not support Microsoft Edge:
https://docs.microsoft.com/en-us/microsoft-edge/deploy/hardware-and-software-requirements


rajsite:
Updated with feedback from @cglzaguilar

make-it/Makefile
```diff
@@ -108,7 +108,7 @@ endif
 .DEFAULT_GOAL=help
 
 $(OUTPUT_DIR):
-	@mkdir -p $(OUTPUT_DIR)
+	"mkdir" -p $(OUTPUT_DIR)
```
spathiwa: Please add the @ back, and make a variable MKDIR = "mkdir" -p with a comment explaining that Windows needs the quote, then use that here.

make-it/Makefile
```diff
@@ -176,8 +176,7 @@ lARMv5:
 
 #========= Use makefile designed for emscripten builds
 vjs: $(OUTPUT_DIR)
-	# Just buid everything
-	make -f EmMakefile vjs
+	emmake make -f EmMakefile vjs
```
spathiwa: Why is emmake necessary?  This will fail on Mac, because emmake contains a #!/usr/bin/env python2, but the system python2 is just called python; and the version-mandled names have the minor version too, e.g. python2.7.

spathiwa:
There's a lot of testing/trial-and-error commits in this PR.  Can you squash some of them so the commits reflect only the major changes you ending up making?

make-it/Makefile
rajsite: The emmake is needed to work on Windows and it does not seem to effect the Linux builds.
I do not know exactly why it doesn't work on Windows without the emmake command...

I saw some posts on the emscripten forums about Mac and python2 name errors that were apparently fixed in newer versions of emsdk.

Can you try
emsdk update
emsdk install sdk-1.37.9-64bit

And see if you still have errors. If so I can move the emmake version to AppVeyor.yml or add a separate make target for windows

spathiwa: emsdk 1.37.9 seems to work (at least regarding  the ability to run emmake).

rajsite: Cool, that is the version on AppVeyor and the version I am trying to get Travis up to separately
👍

rajsite:
@spathiwa updated Makefiles, please review changes

PR #276, opened 2017-05-24
Fix to defect DE11926 ReshapeArray of Array of Error Clusters fails

cglzaguilar:
Vireo code to get next element from array iterator was working incorrectly.


source/core/Array.cpp
```diff
@@ -1351,44 +1351,52 @@ struct ArrayReshapeStruct : public VarArgInstruction
...
-            }
-            for (IntIndex i = 0; i < _indexDim; i++) {
-                _indexStack[i] = 0;
-                _indexDim = 0;
+            else {
+                increasingIndexes = false;
             }
         }
-        return (void*)_array->BeginAtND(_rank, _indexStack);
+
```
spathiwa: The Boolean variable seems unnecessary, and since this is used by run-time callbacks, I would try to optimize it out:

while (dimensionIndex < _rank) {
    _indexes[dimensionIndex]++;
    if (_indexes[dimensionIndex] >= _dimensionsLenths[dimensionIndex]) {
        _indexes[dimensionIndex] = 0;
        ++dimensionIndex;
        continue;
    }
}

source/core/Array.cpp
```diff
@@ -1351,44 +1351,52 @@ struct ArrayReshapeStruct : public VarArgInstruction
 class ArrayIterator
 {
  private:
-    ArrayDimensionVector  _indexStack;
-    IntIndex _indexDim;
-    IntIndex*  dimensions;
-    IntIndex  _rank;
     TypedArrayCoreRef _array;
+    IntIndex  _rank;
+    IntIndex*  _dimensionsLengths;
```
spathiwa: Please use _dimensionLengths (singular first word) to be less awkward and match similar variable usage elsewhere.

source/core/Array.cpp
```diff
@@ -1351,44 +1351,48 @@ struct ArrayReshapeStruct : public VarArgInstruction
...
-            if (_indexDim >= _rank) {
-                return NULL;
-            }
-            for (IntIndex i = 0; i < _indexDim; i++) {
-                _indexStack[i] = 0;
-                _indexDim = 0;
+        IntIndex dimensionIndex = 0;
+        while (dimensionIndex < _rank) {
+            _indexes[dimensionIndex]++;
+            if (_indexes[dimensionIndex] >= _dimensionLengths[dimensionIndex]) {
```
sanmut: Can we replace >= with ==?

source/core/Array.cpp
```diff
@@ -1351,44 +1351,48 @@ struct ArrayReshapeStruct : public VarArgInstruction
...
+                _indexes[dimensionIndex] = 0;
+                dimensionIndex++;
+            } else {
+                break;
             }
         }
-        return (void*)_array->BeginAtND(_rank, _indexStack);
+
+        if (dimensionIndex >= _rank) {
```
sanmut: Can  we replace >= with == ?

source/core/Array.cpp
spathiwa: I disagree.  The expression to check if an index is in bounds should be a form of the invariant "index < length".   This expression is the straight-forward inverse of that to check if it's out of bounds and it seems more rigorous and readable to me.   The two are equivalent in the context of this loop, but in general, why would you prefer a weaker test?

source/core/Array.cpp
spathiwa: Disagree, see reason above.

source/core/Array.cpp
sanmut: I see your point. The rigorous check suggests that it is possible in this code for ">" to happen when it actually is not. I prefer a check that is necessary for the algorithm. Something that is more than necessary -- for a defensive style programming -- can be handled via an explicit error or ASSERT in debug builds (whichever is appropriate on a case by case basis)  -- which will again explain the intent of the developer that this MUST not happen. When there is a real need for '>' to happen in the algorithm, then the assert can be removed and the condition can be modified -- which will make  the intent of the code clear. A more rigorous check actually allows for some unintentional code to creep in where the index goes past the length without being equal first and not being caught.

PR #277, opened 2017-05-31
Fix polymorphic array operations

sanmut:
Fixing polymorphic array operations that were failing when the dimensions of input arrays were not same.


spathiwa:
You should do the refactor to use iterators in a separate commit than the actual bug fix.
It is not obvious where the actual bug fix is.

JackDunaway:
> Also, be careful that the refactor did not affect run-time performance

Agree. Yet, @spathiwa, are you able to provide performance test to substantiate our concern, also helping @sanmut validate? Kind regards, -JRD

sanmut:
Addressed @spathiwa's feedback by splitting into multiple commits and optimizing the case -- where the dimension lengths are same -- a bit by just doing pointer arithmetic as done before . I am not concerned about the performance of using the Next iterator also much for the case of dimension lengths being different.

source/core/GenericFunctions.cpp
```diff
@@ -1977,18 +1985,18 @@ VIREO_FUNCTION_SIGNATURET(VectorVectorBinaryOp, AggregateBinOpInstruction)
...
         _PROGMEM_PTR(snippet, _function)(snippet);
-        srcArray1IterPtr = (AQBlock1 *)srcArray1Iter.Next();
-        srcArray2IterPtr = (AQBlock1 *)srcArray2Iter.Next();
-        destArray1IterPtr = (AQBlock1 *)destArray1Iter.Next();
+        srcArray1IterPtr = isInputArraysDimensionsSame ? (srcArray1IterPtr + elementSize1) : (AQBlock1 *)srcArray1Iter.Next();
+        srcArray2IterPtr = isInputArraysDimensionsSame ? (srcArray2IterPtr + elementSize2) : (AQBlock1 *)srcArray2Iter.Next();
+        destArray1IterPtr = isInputArraysDimensionsSame ? (destArray1IterPtr + elementSizeDest) : (AQBlock1 *)destArray1Iter.Next();
         if (destArray2)
-           destArray2IterPtr = (AQBlock1 *)destArray2Iter.Next();
+           destArray2IterPtr = isInputArraysDimensionsSame ? (destArray2IterPtr + elementSizeDest) : (AQBlock1 *)destArray2Iter.Next();
```
spathiwa: There's still a performance issue with testing isInputArraysDimensionsSame 4 different times for every element in the loop.
Better to test it once, and split out the 4 assignments into an if and else branch, avoiding ?: operator.
Actually, better to duplicate the whole while loop and test outside it.  This is the driver loop for all array/array binops, so performance is important.

sanmut: Done.

PR #278, opened 2017-05-31
Fixes to Array defects

cglzaguilar:
DE12861 Sort1DArray primitive incorrectly sorts NaN, -inf and inf
DE12894 Search1DArray primitive doesn't find NaN
DE12895 Threshold1DArray primitive doesn't work properly with Infinity, NaN and other cases


source/core/GenericFunctions.cpp
```diff
@@ -1782,12 +1783,32 @@ VIREO_FUNCTION_SIGNATURE1(IsLTAccumulator, void)
...
+    return _this;
+}
+//------------------------------------------------------------
+// Execute a snippet of binops until one of them returns true or the commutative pair returns true
+VIREO_FUNCTION_SIGNATURE1(IsLTSortAccumulator, void)
+{
+    BinaryCompareInstruction* binop = (BinaryCompareInstruction*)_ParamPointer(0);
+    if (DoLTBinayCompareAccumulator(binop)) {
+        return null;
+    }
```
spathiwa: I don't think you need to wrap this function.  Instead, revert this and use 
DEFINE_VIREO_FUNCTION_CUSTOM( IsLTSortAccumulator, IsLTAccumulator, "p(i(GenericBinOp))");
below so the LT and LTSort accumulator can be directly shared.

source/core/GenericFunctions.cpp
```diff
@@ -1733,9 +1731,29 @@ VIREO_FUNCTION_SIGNATURE1(IsEQAccumulator, void)
...
+    return _this;
+}
+//------------------------------------------------------------
+// Accumulators are used for elements of an array and elements of a cluster
+VIREO_FUNCTION_SIGNATURE1(IsEQSearchAccumulator, void)
+{
+    BinaryCompareInstruction* binop = (BinaryCompareInstruction*)_ParamPointer(0);
+    if (DoIsEQBinayCompareAccumulator(binop)) {
+        return null;
+    }
```
spathiwa: I don't think this wrap is necessary, see my other comment on LTSortAccumulator.

spathiwa: I meant to click request changes, not approve... there doesn't appear to be a way to correct it.

spathiwa:
This has nothing to do with this PR, but we really need to figure out the cause of that intermittent failure in QueueRefNumGetStatus.
The test does queue operations in parallel, with WaitMS calls to control timing of output.
I tried increasing the times between operations, which helped but did not make the problem go away.
@rajsite Do you know if there's anything about the Firefox run that would have less reliable timing than the other runs?  This failure is always on Firefox.

source/core/Array.cpp
```diff
@@ -1530,9 +1530,17 @@ VIREO_FUNCTION_SIGNATURE4(ArrayThreshold, Double, TypedArrayCoreRef, Double, Int
 {
```
sanmut: Please add comments for what each of the argument is

source/core/Math.cpp
```diff
@@ -275,8 +275,20 @@ DECLARE_SCALE2X_INTN_HELPER(Single)
 // Comparison
 #define DECLARE_VIREO_COMPARISON_PRIMITIVES_BASE(TYPE) \
     DECLARE_VIREO_PRIMITIVE3(IsLT##TYPE, TYPE, TYPE, Boolean, (_Param(2) = _Param(0) <  _Param(1))) \
+    DECLARE_VIREO_PRIMITIVE3(IsLTSort##TYPE, TYPE, TYPE, Boolean, \
```
sanmut: Please add one line comment that IsLTSort is needed only for sort to handle NANs.

source/core/Math.cpp
```diff
@@ -275,8 +275,20 @@ DECLARE_SCALE2X_INTN_HELPER(Single)
...
+        if (::isnan((double)_Param(1))) { \
```
sanmut: Please add (if they don't exist already) gvi and via tests using all the comparison operators comparing NaN's and Inf's and check those tests in along with this changeset.

source/core/Array.cpp
```diff
@@ -1526,6 +1526,8 @@ VIREO_FUNCTION_SIGNATURE4(ArrayInterpolate, void, TypedArrayCoreRef, StaticType,
     return _NextInstruction();
 }
 
+//----------------------------------------------------------------------
+// interpolated result(0), input array(1), threshold y(2), star index(3)
```
sanmut: Thanks for updating. typo:  "start index"

PR #279, opened 2017-06-01
Fix broken Windows release build

sanmut:


Vireo_VS/VireoCommandLine.vcxproj
```diff
@@ -96,7 +96,7 @@
     <UseDebugLibraries>false</UseDebugLibraries>
     <PlatformToolset>v140</PlatformToolset>
     <WholeProgramOptimization>true</WholeProgramOptimization>
-    <CharacterSet>Unicode</CharacterSet>
+    <CharacterSet>MultiByte</CharacterSet>
```
cglzaguilar: I think this would be my only concern, can you elaborate why we need to use Multibyte instead of Unicode.

spathiwa: Yes, I have this concern too.   Does this refer to the character set of the vcproj or of source code?   GitHub expects/uses UTF-8 encoding for source, correct, and only does client conversion for line endings?

sanmut: I was trying to match Release configuration with Debug configuration (except for optimization, debug info, etc, of course). I am not sure whether there was a reason why Debug was "MultiByte" to begin with, @spathiwa ? We don't seem to be using any multi-byte / Unicode strings yet so it does not matter now, I think ! If we want to go with "Unicode" or just choose "Not set", we should fix both Debug and Release builds.

spathiwa: I vote for choosing either Unicode or "not set" on both.

sanmut: I will change it to Unicode. It seems to matter only for which version of Win32 APIs are called. Since we don't expect anyone to be running Vireo in the future on any WinOS older than Win-NT, I think we can go with Unicode. I will fix both.

cglzaguilar: Unicode for both sounds good to me.

sanmut: Looks like I spoke too soon. Looks like there are some file system code that is already using mult-byte/ascii dependent Win32 APIs. If we want to go to Unicode, we have to fix those. I don't see a reason to put any effort to do that now. I have set the character set to "Not Set" in the VC project now. Again, this change will simply let us build both release and debug with similar configurations.

PR #281, opened 2017-06-05
Fix how UTF8 strings are read from memory

rajsite:
Several locations intended to read Vireo Strings from memory were using the Pointer_stringify function.

Pointer_stringify takes a length parameter which is ignored in the case of UTF8 strings. This causes errors as the function will read passed the end of the string. In addition, null characters contained within the Vireo String cause the string to be truncated at that character.

The sizedUtf8ArrayToJSString function was made by modifying the Pointer_stringify algorithm to not terminate on null characters and to take a length parameter.

Also validated that the remaining active Pointer_stringify use cases are being used on null terminated C strings exposed by the Vireo APIs.

### DIFF
--- a/source/core/module_coreHelpers.js
+++ b/source/core/module_coreHelpers.js
@@ -57,19 +57,20 @@
         };
 
         // Takes Vireo Strings that are UTF-8 encoded strings with known length and returns a JS string
+        // TODO mraj assumes valid UTF8 encoding https://github.com/ni/VireoSDK/issues/283
         Module.coreHelpers.sizedUtf8ArrayToJSString = function (u8Array, startIndex, length) {
             /* eslint-disable no-continue, no-plusplus, no-bitwise */
             var u0, u1, u2, u3, u4, u5;
             var idx = startIndex;
-            var lastVisitedIndex = startIndex + length;
-            lastVisitedIndex = lastVisitedIndex > u8Array.length ? u8Array.length : lastVisitedIndex;
+            var endIndex = startIndex + length;
+            endIndex = endIndex > u8Array.length ? u8Array.length : endIndex;
             var str = '';
             while (true) {
-                // For UTF8 byte structure, see http://en.wikipedia.org/wiki/UTF-8#Description and https://www.ietf.org/rfc/rfc2279.txt and https://tools.ietf.org/html/rfc3629
-                u0 = u8Array[idx++];
-                if (idx > lastVisitedIndex) {
+                if (idx >= endIndex) {
                     return str;
                 }
+                // For UTF8 byte structure, see http://en.wikipedia.org/wiki/UTF-8#Description and https://www.ietf.org/rfc/rfc2279.txt and https://tools.ietf.org/html/rfc3629
+                u0 = u8Array[idx++];
                 if (!(u0 & 0x80)) {
                     str += String.fromCharCode(u0);
                     continue;
--- a/source/core/module_coreHelpers.js
+++ b/source/core/module_coreHelpers.js
@@ -56,15 +56,22 @@
             fpSync = fn;
         };
 
-        Module.coreHelpers.utf8ArrayToStringWithNull = function (u8Array, idxIn) {
+        // TODO mraj when we manipulate strings in vireo do we always utf8 encode the value?
+        // TODO mraj the manual and textDecoder function may return different values in cases of corrupt utf8 or truncated multibyte utf8 character
+        // ie the character 128 would be 2 bytes in utf8
+        // Requires a length. Should NOT include a terminating null as part of length.
+        // Allows UTF8 arrays with internal null values.
+        var manualSizedUtf8ArrayToJSString = function (u8Array, startIndex, length) {
             /* eslint-disable no-continue, no-plusplus, no-bitwise */
             var u0, u1, u2, u3, u4, u5;
-            var idx = idxIn;
+            var idx = startIndex;
+            var lastVisitedIndex = startIndex + length;
+            lastVisitedIndex = lastVisitedIndex > u8Array.length ? u8Array.length : lastVisitedIndex;
             var str = '';
             while (true) {
                 // For UTF8 byte structure, see http://en.wikipedia.org/wiki/UTF-8#Description and https://www.ietf.org/rfc/rfc2279.txt and https://tools.ietf.org/html/rfc3629
                 u0 = u8Array[idx++];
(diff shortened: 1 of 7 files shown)


source/core/module_coreHelpers.js
```diff
@@ -56,15 +56,18 @@
...
         };
 
-        Module.coreHelpers.utf8ArrayToStringWithNull = function (u8Array, idxIn) {
+        // Takes Vireo Strings that are UTF-8 encoded strings with known length and returns a JS string
+        Module.coreHelpers.sizedUtf8ArrayToJSString = function (u8Array, startIndex, length) {
             /* eslint-disable no-continue, no-plusplus, no-bitwise */
             var u0, u1, u2, u3, u4, u5;
-            var idx = idxIn;
+            var idx = startIndex;
+            var lastVisitedIndex = startIndex + length;
```
spathiwa: This is weird, implying length is a length of the remaining string starting at index, rather than the whole string, even though the whole string is passed in.  That seems like a bad API.

test-it/karma/fixtures/publicapi/FPSyncUtf8.via
```diff
@@ -0,0 +1,17 @@
+define (c0 dv(.Double 7))
+define (c1 dv(.String 'Iñtërnâtiônàlizætiøn☃💩'))
+define (c2 dv(.Double 8))
```
spathiwa: Does this string have an embedded null to test that aspect of your change?

source/core/module_coreHelpers.js
rajsite: The function takes a UTF8Array of bytes, a start index, and a length (number of utf8 bytes). We are not passing in a string, but instead a UT8Array of bytes and it returns a JS string.

One caller happens to use the whole buffer so it passes a startindex of 0 and the length of the the whole buffer (vireo.loader.js)
Another caller actually has a starting index and length to extract a subset of the buffer (see module_eggShell.js)

Does seeing both usages make it clearer?
Would startIndex (inclusive) and endIndex (exclusive) make more sense than startIndex + length for the api?

spathiwa: Never mind.  It's clear enough, I think my confusion came from thinking in terms of C APIs, but in this case the UTF8Array knows its length, and doesn't need to be passed.

However, there are still some problematic corner cases:
The code will read one more byte than it should of u8Array. For example, if the array is empty, or you pass length of 0, it will still read the first element before returning.  It doesn't assign it to str, but it still reads past the end of the array.   Just make the test >= and move it before the array read.
Also, if you pass a length in a middle of a UTF8 code point, it will read more bytes than specified, including past the end of the array if the array isn't well-formed Unicode.

test-it/karma/publicapi/FPSync.Test.js
```diff
@@ -0,0 +1,67 @@
...
+                myDouble: myDouble,
+                fpSyncString: fpSyncString
+            });
+        };
+
+        vireo.coreHelpers.setFPSyncFunction(tracker);
+        var runSlicesAsync = vireoRunner.rebootAndLoadVia(vireo, publicApiFPSyncUtf8ViaUrl);
+
+        runSlicesAsync(function (rawPrint, rawPrintError) {
+            var fpSyncString = 'Iñtërnâtiônàlizætiøn\0☃💩';
```
sanmut: I wonder whether we should decode this expected string from UTF-8 encoded byte string (in hex) of the actual string in the .via file .

source/core/module_coreHelpers.js
```diff
@@ -56,17 +56,21 @@
             fpSync = fn;
         };
 
-        Module.coreHelpers.utf8ArrayToStringWithNull = function (u8Array, idxIn) {
+        // Takes Vireo Strings that are UTF-8 encoded strings with known length and returns a JS string
+        // TODO mraj assumes valid UTF8 encoding https://github.com/ni/VireoSDK/issues/283
```
sanmut: Can we use https://github.com/mathiasbynens/utf8.js that seems to throw an error on malformed UTF-8 strings?

rajsite: I think we need to step back and come up with a strategy for how to handle incorrectly encoded UTF-8 strings in Vireo.

Just erroring out might not be the best approach because LV semantics for strings just treat them as byte arrays. So if a user makes a wacky string that is invalid utf8 and wires it to a string indicator that shouldn't throw an exception.

test-it/karma/publicapi/FPSync.Test.js
rajsite: We could but I don't think it would be clearer.

The code point for 💩 is \x1F4A9

The JS string representation is "UTF-16" so each "code unit" is 16-bit. So that 💩 code point is broken into two code points ([with the surrogate pair algorithm](https://en.wikipedia.org/wiki/UTF-16#U.2B10000_to_U.2B10FFFF)) which is two "code units" or 4 bytes total hex:
\xD83D \x56489

The JS length of the string is two.

The Vireo string representation is "UTF-8" so each "code unit" is 8-bit. So that 💩 code point is converted into 4 "code units"  or 4 bytes hex: 
\xF0\x9F\x92\xA9

The Vireo length of the string is four.

So placing the hex would be more precise in the test but I don't think it would be clearer at demonstrating what is being tested.

source/core/module_coreHelpers.js
rajsite: Fixed the first case.

The second case of handling invalid UTF-8 code points I opened a separate issue that I'll add more detail to https://github.com/ni/VireoSDK/issues/283

One thing I wanted to point out is that as-is the algorithm shouldn't result in an error. Out of bounds access in JS results in the value of `undefined` and the code [as used from emscripten](https://github.com/rajsite/emscripten-sdk/blob/ee4476d49228259c27847d7a7794906368fa99db/emscripten/master/src/preamble.js#L549) would perform an `undefined & 63` which causes JS to coerce `undefined` to `0`.

And with strings in the middle of an array it may result in corrupt output reading a couple of extra bytes out of the array.

Not saying this is a good approach to handling invalid UTF-8 code points, but as is it shouldn't cause fatal exceptions until a consistent way for handling that case is determined.

test-it/karma/fixtures/publicapi/FPSyncUtf8.via
rajsite: Added an embedded null

PR #286, opened 2017-06-20
Fixes to defects

cglzaguilar:
DE13004 Instruction not generated for RoundToNearest, Floor, Ceil
DE13006 Instruction not generated for Scale2X


source/core/GenericFunctions.cpp
```diff
@@ -1667,10 +1668,6 @@ VIREO_FUNCTION_SIGNATURET(ClusterBinaryOp, AggregateBinOpInstruction)
             convertSnippet->_p1 = (AQBlock1*)(intptr_t(whichConvertArg) << 24);
             snippet = (Instruction4<AQBlock1, AQBlock1, AQBlock1, AQBlock1>*) next;
         }
```
spathiwa: This is wrong to fall through.  It will now handle convertArgs for every OTHER cluster elemenet.
It also now needs to save/restore the whichConvertArg (convertSnippet->_p1) for each element, not just the first.

test-it/ViaTests/Scale2XWithIntegers.via
```diff
@@ -0,0 +1,1208 @@
...
+        And(local463 local659 local659)
+        And(local494 local659 local659)
+        And(local525 local659 local659)
+        And(local557 local659 local659)
+        And(local580 local659 local659)
+        And(local625 local659 local659)
+        And(local657 local659 local659)
+        Copy(local659 dataItem_Pass)
+        Printf ("000 All tests pass: '%s'.\n" dataItem_Pass)
+		Printf ("010     Double test pass: '%s'.\n" dataItem_DoublePass)
```
sanmut: Indentation.

test-it/testList.json
```diff
@@ -8,6 +8,7 @@
                 "ListDirectory.via",
                 "Round.via",
                 "Scale2X.via",
+                "Scale2XWithIntegers.via",
```
sanmut: Why is this not in common?

cglzaguilar: If I remember correctly we put it here since it was failing on appveyor, you might ask Milan about it, meanwhile I'm just going to put it here, like the Scale2.via. We can move it later.

PR #291, opened 2017-06-28
Fixes to generic array operations.

cglzaguilar:
Most of the operations only worked with 1D arrays.
Fixes cover at least the following defects
DE13258 Vireo abort when calling transpose 2D Array
DE13026 Not operator doesn't work on arrays of two dimensions


spathiwa:
Looks good to me, although I only had time to give it a cursory look right now.

source/core/GenericFunctions.cpp
```diff
@@ -683,13 +683,18 @@ VIREO_FUNCTION_SIGNATURET(VectorMaxMinOp, AggregateMaxAndMinInstruction)
     IntIndex elementSizeY = srcArrayY->ElementType()->TopAQSize();
     IntIndex elementSizeMax = maxArray->ElementType()->TopAQSize();
     IntIndex elementSizeMin = minArray->ElementType()->TopAQSize();
+    IntIndex* newDimensionLengths = srcArrayX->DimensionLengths();
+    IntIndex rank = srcArrayX->Rank();
     IntIndex count = srcArrayX->Length();
-    if (srcArrayY->Length() < count)
+    if (srcArrayY->Length() < count) {
```
sanmut: I am not sure whether this is right. If the count is same but dimension length is different for the different dimensions, is this doing the right thing (as done by LabVIEW). I think we should write test cases for the different use cases as discussed in person.

spathiwa: I agree, this is wrong.  It actually needs to find the minimum of all the dimsizes individually and use the ArrayIterator to traverse the dims, like the VectorVectorBinaryOp does.
The max/min of a 4x2 array and a 2x4 array should produce a 2x2 array and ignore the extra elements in each array the other one doesn't have.

source/core/GenericFunctions.cpp
```diff
@@ -668,7 +668,36 @@ InstructionCore* EmitMaxMinEltsInstruction(ClumpParseState* pInstructionBuilder)
     }
     return pInstruction;
 }
+//------------------------------------------------------------
+bool GetMinimumArrayDimensions(TypedArrayCoreRef sourceArray1, TypedArrayCoreRef sourceArray2,
```
sanmut: I would make this as a static function only if used within this file.

source/core/GenericFunctions.cpp
```diff
@@ -2232,7 +2269,9 @@ VIREO_FUNCTION_SIGNATURET(VectorScalarBinaryOp, AggregateBinOpInstruction)
     IntIndex count = srcArray1->Length();
 
     // Resize output to size of input array
-    destArray->Resize1D(count);
+    IntIndex* newDimensionLengths = srcArray1->DimensionLengths();
+    IntIndex rank = srcArray1->Rank();
+    destArray->ResizeDimensions(rank, newDimensionLengths, true);
```
sanmut: destArray2 should be resized with newDimensionLengths as well, right?

source/core/GenericFunctions.cpp
```diff
@@ -2191,7 +2226,9 @@ VIREO_FUNCTION_SIGNATURET(ScalarVectorBinaryOp, AggregateBinOpInstruction)
...
```
sanmut: destArray2 should be resized with newDimensionLengths as well, right?

source/core/GenericFunctions.cpp
spathiwa: Looks like this was an oversight in the original as well, which means there probably isn't a test for this.
ReOrImToPolar or PolarToReOrIm on arrays should be a good test case for this,
and ComplexToReOrIm a test case for the un-op version.

source/core/GenericFunctions.cpp
sanmut: Please mention the list of primitives/nodes that will result in calling each of the different functions such as VectorScalar/VectorVector/Bin/UnOp just above each function. I think it might be easy as you have different tests now written and fresh in your memory.

spathiwa: That would be huge list, as most prims are either unops or binops, and which is which is pretty obvious, as is whether the arguments are scalar or vector deciding which of these wrappers is used.
What I think does needs to be documented are the exceptions and corner cases, such as unop/binops with two outputs.

Issue #283, opened 2017-06-06
Come-up with consistent approach for handling invalid UTF-8 strings


rajsite:
Currently the function used for reading strings from Vireo's memory assumes a valid UTF-8 format: https://github.com/ni/VireoSDK/blob/6a94624770d179ce14ab06e65731723ba1478f8c/source/core/module_coreHelpers.js#L60

As Vireo / LabVIEW strings are often treated as byte arrays the possibility for invalid UTF-8 string encodings is high.

I incorrectly believed that the string corruption from assuming valid code points would not cause internal errors: https://github.com/ni/VireoSDK/pull/281#discussion_r120449193

However I ended up finding a case where this does result in an error. For the LabVIEW string `\127` when it is read as JSON the string is encoded as `"\127"`

The algorithm causes the string to results in `"\crazystuff` causing the closing quote to be lost in corruption and resulting in invalid JSON. So while no error occurs in Vireos space, when the string is parsed in JS land an exception is thrown because it is invalid JSON.

Some discussion on Wikipedia: https://en.wikipedia.org/wiki/UTF-8#Invalid_byte_sequences

The wiki article describes "popular" replacement options:
1. Use the unicode replacement character
2. Use the low byte of the invalid code point range U+DC80-U+DCFF
3. Use code points U+0080–U+00FF with the same value as the byte
4. The Unicode code point for the character represented by the byte in CP1252

Some thoughts on the options:
1. After reading through the other popular options I think this will be the clearest. It makes it very easy to identify corrupt bytes and the number of corrupt bytes. It is also semantically appropriate. The replacement character is intended to replace unknown, unrecognized, or unrepresentable characters.
2. Does not work because we are going from UTF-8 to UTF-16 and the invalid code point range is utilized for UTF-16 surrogate pairs
3. This was my first thought at implementation but it results in a lot of collisions. ie if the user did intend to use codepoints U+0080-U+00FF in correctly formatted UTF-8 they would be indistinguishable from invalid bytes mapped to that range. 

    I think it's also possible to miss the corruption because there are valid and not too uncommon symbols in the range.

    As the string functions are used to read values from Vireo memory for controls and indicators, if the user (vireo integrator) believes invalid UTF-8 strings / arbitrary byte arrays may be present they should not try to read strings as JSON strings but as JSON arrays of numbers or using another API function. The benefit of readJSON on a string type is to aid in conversion from UTF-8 encoded strings to UTF-16 encoded strings.
[...]

PR #292, opened 2017-06-30
HTTP Binary support and safe string reads from Vireo

rajsite:
- Modifies the HTTP client to use TypedArrayViews / Blobs directly from Vireos memory to write / read instead of converting to a DOMString first
- Modifies the vireo string byte array to js string functions to validate utf-8 structure and emit unicode replacement codepoints for invalid utf-8 byte sequences (does not validate for utf-8 "safety" only byte structure)
- Starts integration of tagging for karma tests to reduce number of karma test runner configuration
- Updates eslint and creates package.json and package-lock.json for npm 5

Edit:
- Also fixes the Appveyor intermittent Chrome failures due to incorrectly configured timeout tests


rajsite:
Note: This change requires you to clear existing caches on TravisCI and to install npm > v5 on your system with `npm install -g npm@latest`.

The benefit of npm 5 is the package-lock.json file which guarantees reproducible builds on the CI. This is because prior to npm 5 and the package-lock.json file it was possible for the versions of subdependencies to change between any two runs of npm install

rajsite:
Did some manual testing and found a case where overlong UTF-8 representations of byte sequences allowed control characters to slip through. Added a fix for that and a lot more binary tests that verify decoding without error.

rajsite:
LVjs CI results: http://lvngtestexperts/Builds/Report/701198
Zero new full test failures. LabVIEW tests will be added on the TFS side.

source/core/String.cpp
```diff
@@ -96,34 +104,40 @@ void String::AppendEscapeEncoded(const Utf8Char* source, IntIndex len)
...
+            *ptr-- = '\\'; *ptr-- = '\\';
+            break;
         case '"':
-            *ptr--  = '\"';
-            *ptr-- = '\\';
-        break;
+            *ptr--  = '"'; *ptr-- = '\\';
+            break;
+        case '\x00': case '\x01': case '\x02': case '\x03': case '\x04': case '\x05': case '\x06': case '\x07':
+            *ptr-- = c + 48; *ptr-- = '0'; *ptr-- = '0'; *ptr-- = '0'; *ptr-- = 'u'; *ptr-- = '\\';
```
cglzaguilar: *ptr-- = c + 48;

Can we use named constants instead of 48? What is the meaning of 48?

rajsite: @cglzaguilar replaced them with the ASCII literal representation with I think does convey the behavior clearer

rajsite:
Added @gleonoliva to help review the JS changes

source/core/String.cpp
```diff
@@ -96,34 +104,40 @@ void String::AppendEscapeEncoded(const Utf8Char* source, IntIndex len)
...
-           break;
+            *ptr-- = '\\'; *ptr-- = '\\';
+            break;
         case '"':
-            *ptr--  = '\"';
-            *ptr-- = '\\';
-        break;
+            *ptr--  = '"'; *ptr-- = '\\';
+            break;
+        case '\x00': case '\x01': case '\x02': case '\x03': case '\x04': case '\x05': case '\x06': case '\x07':
```
spathiwa: I think this switch statement looks pretty ugly (like generated code), and it would be better to move this handling for all the \xDD cases into the default case, with explicit tests of (c & 0xF >= 10) for the 'A' vs '0' base character mapping, and (c & 0xF0 >= 0x10) for '0' vs '1' for 2nd to last digit.

spathiwa:
Defer to others for most of the JavaScript changes.

source/core/String.cpp
rajsite: @spathiwa updated to switch to conditionals for the ranges and moved them to the default case.

I prefer the ranges because it makes it easy to inspect coverage of the range x00 - x1F for the control characters over masking in the conditionals. I also want to preserve the symmetry of the conditionals in the look ahead section and the character insertion section.

Let me know if you feel strongly otherwise.

rajsite: This code is much easier to read outside the diff https://github.com/rajsite/VireoSDK/blob/efe8f4d6944bf64442388ab4e7c3609367bf2cf3/source/core/String.cpp#L129

rajsite:
@spathiwa and @cglzaguilar I implementing the formatting changes requested for String.cpp. Please approve if they look good to y'all :)

source/core/module_coreHelpers.js
```diff
@@ -75,33 +104,68 @@
                     str += String.fromCharCode(u0);
                     continue;
                 }
+
+                // Look ahead to validate the UTF-8 structure
+                if ((u0 & 0xE0) === 0xC0) {
```
gleonoliva: Seeing these hex values makes it hard to understand. At first I thought adding some comments, but perhaps declaring constants could help too.

source/core/module_coreHelpers.js
```diff
@@ -75,33 +104,68 @@
...
+                    if (idx >= endIndex || (u8Array[idx] & 0xC0) !== 0x80) {
+                        str += '\uFFFD';
+                        continue;
+                    }
+                } else if ((u0 & 0xF0) === 0xE0) {
```
gleonoliva: Maybe adding helper functions to end up with something like:
`if(bytesRequired(u0) == 1) { ... }`

source/core/module_coreHelpers.js
gleonoliva: Also it looks like inside this while(true) we are doing 3 different things. Do you think is possible to use methods for each thing? I think that would help understand better this code. Perhaps validate(), encode() and transformSurrogatePair()

spathiwa: If I may push back on that a little, these constants don't have any semantic meaning that would lend themselves to good named constants; they are defined in such a way that the encoding space for UTF-8 keeps more common characters with more compact encodings then less common ones.   If you are trying to work through this code to see how it encodes a particular byte array portion, having the hex constants directly in the code actually makes it easier to understand/debug, not harder, because there is no meaning you would add by changing the values to symbolic constants.

spathiwa: However, I think a link to the spec or resource you followed to implement this, or inline comments with example encodings, would be helpful.

rajsite: Yea talked about it with craig, memo, and jesse and agree that for now will add additional comments to help clarify the behavior of each region and expected byte structures.

I will also create a separate issue for breaking up the code into more components including benchmarking. Right now I'm worried that sub function calls may result in a performance penalty so we should make that change when we can benchmark for regressions as well. We should also investigate usage of the [JavaScript TextDecoder API](https://www.w3.org/TR/encoding/#dom-textdecoder) which may be faster but might only support valid utf-8 sequences and may have varying crossbrowser behavior

rajsite:
Awesome! @cglzaguilar can you rebase merge and version bump when you get a chance?

rajsite:
Fixed #283

PR #293, opened 2017-07-05
Fix crash when ReleaseQueue called with invalid Refnum.

sanmut:
Add a test to catch issues with invalid Refnums


sanmut:
@cglzaguilar appveyor failure is due to intermittent firefox issue and nothing wrong with the actual changes.

PR #298, opened 2017-07-17
First pass at CODEOWNERS file

rajsite:
Adding configuration for the Github CODEOWNERS feature. More info here: https://github.com/blog/2392-introducing-code-owners

I tried to guess at what people would be interested in owning.

The Pull Request is editable by reviewers. Please add or remove yourself to different parts and approve when you are satisifed.

Edit: If you are unable to edit directly just post a comment about what changes you want

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,15 @@
+# Each line is a file pattern followed by one or more owners.
+# More info: https://github.com/blog/2392-introducing-code-owners
+
+# These owners will be the default owners for everything in the repo.
+*                           @cglzaguilar
+
+# Owners of specific locations in the repo.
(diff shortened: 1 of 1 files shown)


rajsite:
GitHub CODEOWNERS is a bit different than TFS. If you just want to observe changes watch the repository.

To be a required code reviewer, add / remove from the CODEOWNERS. Also as described in the comment "Order is important. The last matching pattern has the most precedence."

So for example I am listed as an owner for *.js which is near the end of the CODEOWNERS file. So if a change comes in with only .js changes I'll be the only required reviewer added.

If we moved the catch-all line to be the last then Carlos would be the only reviewer for all files in the repo.

rajsite:
accidentally pushed on my incoming, closing to reset and moving to branch

PR #302, opened 2017-07-18
Merge errors

cglzaguilar:
Fix for defect DE12829. Implement Retain First Error Primitive


source/core/GenericFunctions.cpp
```diff
@@ -2437,6 +2437,84 @@ VIREO_FUNCTION_SIGNATURE4(ConvertEnum, StaticType, void, StaticType, void)
...
+// arguments: output cluster, then variable number of inputs that can be null, error cluster or a 1D array of error clusters
+struct MergeErrorsParamBlock : public VarArgInstruction
+{
+    _ParamDef(ErrorCluster, ErrorClusterOut);
+    _ParamImmediateDef(StaticTypeAndData, ErrorClusterInputs[1]);
+    NEXT_INSTRUCTION_METHODV()
+};
+
+static Boolean UpdateErrorWhenErrorOrWarningFound(ErrorCluster* errorClusterIn, ErrorCluster* errorClusterOut, Boolean findErrors) {
+    Boolean errorFound = errorClusterIn->status && findErrors;
```
sanmut: Please consider adding an inline method called "isError()" in ErrorCluster and use it everywhere.

source/core/GenericFunctions.cpp
```diff
@@ -2437,6 +2437,84 @@ VIREO_FUNCTION_SIGNATURE4(ConvertEnum, StaticType, void, StaticType, void)
...
+    Boolean warningFound = !errorClusterIn->status && errorClusterIn->code != 0 && !findErrors;
```
sanmut: Please consider adding an inline method called "isWarning()" in ErrorCluster and use it everywhere.

source/core/GenericFunctions.cpp
```diff
@@ -2437,6 +2437,84 @@ VIREO_FUNCTION_SIGNATURE4(ConvertEnum, StaticType, void, StaticType, void)
...
+    return false;
+}
+
+//------------------------------------------------------------
+// MergeErrors function for error clusters, supports multiple inputs
+VIREO_FUNCTION_SIGNATUREV(MergeErrors, MergeErrorsParamBlock)
+{
+    // Initialize output error cluster to not error
+    ErrorCluster* errorClusterOut = _ParamPointer(ErrorClusterOut);
+    errorClusterOut->SetError(false, 0, "", false);
```
sanmut: If there is neither error nor warning, I believe we should return the "source" string from the first input. May not be a big deal, but no need to break either.

spathiwa: Good catch.  The code should function as if the first error is in-place to the output.

source/core/GenericFunctions.cpp
spathiwa: Agree with suggestion, but I'd prefer hasError() and hasWarning(), because isError() could be misread as if it's querying the type rather than the data.

source/core/GenericFunctions.cpp
cglzaguilar: I did a check on this one by running a .gvi under desktop and the source string is not returned from the first input where there aren't errors or warnings.

spathiwa: Then that's a bug in desktop Merge Errors as well.  It does have this behavior in CG.

spathiwa: Just asked David, he agreed it's probably a bug and there should be a CAR.

sanmut: Looks like it works if the first input is not an array and does not work otherwise with .gvi on Desktop. -- which is definitely confusing. I will file David a CAR on this.

source/core/GenericFunctions.cpp
sanmut: Agree with spathiwa

cglzaguilar:
Closing to create a new one to trigger CI

PR #306, opened 2017-07-24
Add support for Timestamp operations and fix usage of cluster type

sanmut:


sanmut:
Without "Make IsA on type comparisons stricter." commit, multiplying a cluster with two doubles actually did a complex multiplication :-(

source/core/TypeAndDataManager.cpp
```diff
@@ -922,6 +924,20 @@ Boolean TypeCommon::IsComplex()
     }
     return false;
 }
+
+//------------------------------------------------------------
+Boolean TypeCommon::IsNamedClusterDataType() {
```
spathiwa: Please rename this to something like IsIntrinsicClusterDataType().
Via-code-defined clusters can have names, but that's not what this is checking.

source/core/TypeAndDataManager.cpp
```diff
@@ -922,6 +924,20 @@ Boolean TypeCommon::IsComplex()
...
+    TypeRef t = this;
+    while (t) {
+        if (t->Name().Compare(&TypeComplexDouble) || t->Name().Compare(&TypeComplexSingle) ||
+            t->Name().Compare(&TypeTimestamp)) {  // TODO(sanmut) Add Waveform types also
```
cglzaguilar: Please add a defect or an issue and reference it in your TODO

source/include/TypeDefiner.h
```diff
@@ -105,6 +105,9 @@ class TypeDefiner
 
 #else
 
+    // With non-unique names(AKA overloaded) only the C entry points are appended.
+    #define DEFINE_VIREO_FUNCTION_TYPED(_root_, _type_, _proto_)  DEFINE_VIREO_FUNCTION_CUSTOM(_root_, _root_##_type_, _proto_)
```
cglzaguilar: Is there something that can be done to avoid defining these functions in two places?

PR #307, opened 2017-07-24
Partial fix for defect DE12090. Large GET response causes JS error

cglzaguilar:
Allocating source array on heap to avoid stack overflow error.

### DIFF
--- a/source/io/module_eggShell.js
+++ b/source/io/module_eggShell.js
@@ -68,7 +68,6 @@
         var Data_GetStringBegin = Module.cwrap('Data_GetStringBegin', 'number', []);
         var Data_GetStringLength = Module.cwrap('Data_GetStringLength', 'number', []);
         var Data_WriteString = Module.cwrap('Data_WriteString', 'void', ['number', 'number', 'string', 'number']);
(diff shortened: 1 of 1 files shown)


rajsite:
The behavior in the code modified itself is well covered in existing tests.

Also this doesn't prevent out of memory errors, it just shifts it from a stack overflow to a heap out of memory. Which I don't believe we can test the difference without building the JS version with ASSERTIONS enabled. Currently we build and test the release configuration of the JS output, not debug or release with assertions.

We could try and invent a test that fails with a stack overflow and doesn't fail with a heap overflow but that seems finicky.

As to testing large http responses the largest I can get out of httpbin using the /bytes/n endpoint is 100KB and not large enough to fully stress this change.

To write a good test I think we need the following:

A way to serve a large file - Maybe as a karma fixture, but I would like to avoid checking in large files into GitHub. Maybe [generate them](https://www.npmjs.com/package/datagen) during test? Maybe leverage [git large file storage](https://git-lfs.github.com/)?

A way to detect stack overflow vs heap allocation failed - Maybe run the JS builds twice? One with normal release and once with release with assertions / debug?

---

I don't think we should block this right now on adding a test since this is covered by manual stress testing (which revealed the issue) but we should add it to a tech debt user story for when we try to improve out of memory behavior.

PR #311, opened 2017-07-26
Fix to defect DE13257. VI with correlation algorithm doesn't work as …

cglzaguilar:
…expected

Problem is related to loss of precision on absolute value for complex numbers when run on browsers.
Updated the calculation of the absolute value


source/core/Math.cpp
```diff
@@ -1043,7 +1046,10 @@ DECLARE_VIREO_PRIMITIVE2(IncrementComplexDouble, ComplexDouble, ComplexDouble, (
 DECLARE_VIREO_PRIMITIVE2(DecrementComplexDouble, ComplexDouble, ComplexDouble, (_Param(1) = _Param(0) - 1.0))
 DECLARE_VIREO_PRIMITIVE2(ReciprocalComplexDouble, ComplexDouble, ComplexDouble, (_Param(1) = 1.0 / _Param(0)))
 DECLARE_VIREO_PRIMITIVE2(SignComplexDouble, ComplexDouble, ComplexDouble, (_Param(1) = _Param(0) / abs(_Param(0)) ))
-DECLARE_VIREO_PRIMITIVE2(AbsoluteComplexDouble, ComplexDouble, Double, (_Param(1) = abs(_Param(0))))
+DECLARE_VIREO_PRIMITIVE2(AbsoluteComplexDouble, ComplexDouble, Double, {
+    ComplexDouble z = (ComplexDouble)_Param(0);
```
spathiwa: Please remove the unnecessary cast.

test-it/ViaTests/ComplexMathFunctions.via
```diff
@@ -25,30 +25,34 @@ define(HelloComplex dv(.VirtualInstrument (
         e(.ComplexSingle sCmplx)
         e(dv(.ComplexSingle (0.0 99.0)) s)
         e(dv(c(e(.ComplexSingle cd)) ((0.0 3.1415265))) c1)
+		e(dv(.ComplexDouble (-96.7324253138614551517 -265.620982121737029047)) cDouble)
```
sanmut: Indentation incorrect.

test-it/ExpectedResults/ComplexMathFunctions.vtr
```diff
@@ -40,3 +40,5 @@ Log (0 99) => (4.59512 1.5708)
 Log2 (0 99) => (6.62936 2.26618)
 Log10 (0 99) => (1.99564 0.682188)
 Log10 ((1.75517 0.958851) (1.62091 2.52441)) => ((0.30103 0.217147) (0.477121 0.434294))
+Absolute Double (-96.7324 -265.621) => 282.6865193998785912
```
sanmut: It might be flaky to expect a floating point number to this precision. May not make this test portable. I wonder whether we should do an epsilon comparison in .via against the expected value. Thoughts @cglzaguilar @spathiwa ?

sanmut:
Commit message "Updated the calculate the directly calculate the absolute value" is not worded clearly.

test-it/ExpectedResults/ComplexMathFunctions.vtr
spathiwa: @cglzaguilar How far off is the browser calculation using stdlib complex abs() directly?
This is a pretty straight-forward calculation on numbers with nearly the same exponent and mantissa range.  The expected value in the test is the same answer CG/NCG computes (using both methods).  While it's always a good idea to avoid exact comparison of floating point values, I wouldn't expect the answer to be off my more than one dbl machine epsilon (2.2e-16).
So I would say, if the answer is within +/- one epsilon, we might consider rolling this back and fixing the test only.  Otherwise this fix is necessary (but the test should still probably be fixed to do an epsilon comparison).

cglzaguilar: Good point. I'll do an epsilon comparison and print less decimals.

sanmut: The following code seems to produce identical results on native system. Could you tell us what they are on the browser? I agree with spathiwa that we should keep the original code unless there is a significant difference.

#include <iostream>
#include <string>
#include <complex>
#include <cmath>
#include <limits>

using namespace std;

int main()
{
  complex<double> c1(-96.7324253138614551517 -265.620982121737029047);
  double abs1 = abs(c1);
  double abs2 = sqrt(c1.real()*c1.real() + c1.imag()*c1.imag());
  cout.precision(std::numeric_limits<double>::max_digits10);
  cout << fixed << abs1 << endl << abs2 << endl;
}

cglzaguilar:
The expected value for (-96.7324 - 265.621i) is: 282.6865193998785912, but we're getting 282.6865193998785344 on the browser, 14 digit is different

When the result of abs is squared, expected value is 79911.66825041793345, but actual value is 79911.66825041790435 on the browser, 11 digit is different

This difference is causing the TestVI_Correlation.gvi test to fail.

cglzaguilar:
Closing this request and creating a new one to see if the AppVeyor issue goes away.

The test that is failing is:

[00:09:46] PhantomJS 2.1.1 (Windows 8 0.0.0) The Vireo VTR test suite can run test DHOStringToNumber FAILED
[00:09:46] 	RangeError: Maximum call stack size exceeded. in dist/vireo.js (line 7)
[00:09:46] 	RH@dist/vireo.js:7:1352

It seems unrelated to my change. Everything passes locally on my machine.

PR #313, opened 2017-07-27
Correlation fix

cglzaguilar:
This is the same pull request as this one:

Fix to defect DE13257. VI with correlation algorithm doesn't work as … #311

Making another pull request to see it the AppVeyor error goes away, which seems unrelated to my changes


rajsite:
Using a new Vireo instance in each run seems to increase test time froom about 12-13 min to 20 min

testtime

Is this something we are okay with? We expect to be able to re-use Vireo instances so this seems unnecessary to have to do.

Is this a problem that only manifests in PhantomJS? If so maybe we just disable PhantomJS since it is a deprecated runtime at this point according to the author.

Can you try sharing the Vireo instance again but instead remove the execution of PhantomJS from AppVeyor: https://github.com/ni/VireoSDK/blob/d374f2c3d3be1c17ec0b7be1fa49d94f5d38cb4f/.appveyor.yml#L39

If removing PhantomJS from AppVeyor works, I would rather do that.

cglzaguilar:
Closing this pull request as a new one that appears to have not problems with AppVeyor-CI has been created.

Issue #310, opened 2017-07-25
test-it/ViaTests/TestVI_ArrayInsertPartial.via fails


sanmut:
A recent change (within a month or so) seems to have made a failing case in the test test-it/ViaTests/TestVI_ArrayInsertPartial.via  to pass but actually caused an already passing case to fail.  Fix the bug and enable this test.

sanmut:
Following commit addressed this:
https://github.com/ni/VireoSDK/commit/e3a0d367a1df92cc9e7a150d146aad919e993c8d

PR #314, opened 2017-07-29
Fix Issue #310, ArrayInsertPartial test.

spathiwa:
ArrayInsertSubset was incorrectly limiting the dim length of
the inserted subarray to be no more than the length of the input
array along the inserted dimension.  All of the other dimensions
are correctly so-pinned (so the array grows along only one dimension), but
the inserted dimension should not be limited.
E.g.  Insert (by rows) into [[1,1][2,2]] a subarray of [3,3,3],[4,4,4],[5,5,5]]
The columns are pinned to 2 elements because the subarray has two many
columns to fit into the input array, but all 3 rows of subarray should
still be inserted into the input array, even though it has only 2 rows
to begin with, because the array grows along that dimension, but not
the others.

### DIFF
--- a/test-it/testList.json
+++ b/test-it/testList.json
@@ -281,6 +281,7 @@
                 "SubVIParameters.via",
                 "SubVISimple.via",
                 "TemplateBasics.via",
+		"TestVI_ArrayInsertPartial.via",
                 "TestVI_LargeArray.via",
                 "TestVI_VectorMaxMinOp.via",
                 "TestVI_VectorScalarBinaryOp.via",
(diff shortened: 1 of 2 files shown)


test-it/testList.json
```diff
@@ -281,6 +281,7 @@
                 "SubVIParameters.via",
                 "SubVISimple.via",
                 "TemplateBasics.via",
+		"TestVI_ArrayInsertPartial.via",
```
cglzaguilar: It seems there is an alignment issue on the new inserted element.

PR #319, opened 2017-08-08
Fix to defect DE13485

cglzaguilar:
DE13485 Max and Min primitive doesn't work with Scalars and Vectors
Enable Vector with Scalar and Scalar with Vector Max and Min operations


source/core/GenericFunctions.cpp
```diff
@@ -762,7 +773,88 @@ VIREO_FUNCTION_SIGNATURET(VectorMaxMinOp, AggregateMaxAndMinInstruction)
...
+        srcArrayYIterPtr += elementSizeY;
+        destArrayMaxIterPtr += elementSizeMax;
+        destArrayMinIterPtr += elementSizeMin;
+    }
+
+    return _NextInstruction();
+}
+//------------------------------------------------------------
+// This function is used by the "Max and Min" primitive when first input is array and second input is scalar
+VIREO_FUNCTION_SIGNATURET(VectorScalarMaxMinOp, AggregateMaxAndMinInstruction)
```
sanmut: I wonder we should just support only ScalarVectorMaxMinOp and not support VectorScalarMaxMinOp at all in Vireo; and let LabVIEW swap the inputs while emitting VIA such that the first is always Scalar and the second is Vector when one of the inputs is scalar and other is vector. This will avoid unnecessary code duplication here and make the runtime smaller. I think it will be an acceptable limitation for VIA that the first input must be scalar ans the second one must be vector when one of the inputs is scalar and the other is vector. Probably, same can be done for ScalarVectorBinaryOp too. Thoughts?

PR #320, opened 2017-08-14
Fix InRangeAndCoerce issues

sanmut:
* Fix the intrinsic cluster data type comparison issues
* Fix the crash due to optional parameter of the InRangeAndCoerce primitive in aggregates mode


source/core/TypeAndDataManager.cpp
```diff
@@ -778,14 +786,22 @@ Boolean TypeCommon::IsA(TypeRef otherType, Boolean compatibleStructure)
                 otherEncoding == kEncoding_Ascii || otherEncoding == kEncoding_Unicode) {
                 bMatch = TopAQSize() == otherType->TopAQSize();
             }
-        } else if (thisEncoding == kEncoding_Cluster && otherEncoding == kEncoding_Cluster &&
-                   !this->IsIntrinsicClusterDataType() && !otherType->IsIntrinsicClusterDataType()) {
-            if (this->SubElementCount() == otherType->SubElementCount()) {
-                for (Int32 i = 0; i < this->SubElementCount(); i++) {
-                    if (!this->GetSubElement(i)->CompareType(otherType->GetSubElement(i)))
-                        break;
+        } else if (thisEncoding == kEncoding_Cluster && otherEncoding == kEncoding_Cluster) {
```
sanmut: I see that this is growing quite a bit now that "TODO(PaulAustin): Consider merging this function with IsA function below." should be addressed sooner. I would like to do it in a separate PR and validate it independently as a release as it seems to require both refactoring and addressing functional differences and do not want to complicate this changeset that fixes a couple of isolated issues.

cglzaguilar: Ok with me, both functions seem to be similar in nature, but have subtle differences and are used in several places. It is a good idea to do the refactoring in another changeset.

source/core/Math.cpp
```diff
@@ -341,7 +341,9 @@ DECLARE_SCALE2X_INTN_HELPER(Single)
     VIREO_FUNCTION_SIGNATURE7(InRangeAndCoerce##TYPE, TYPE, TYPE, TYPE, Boolean, Boolean, TYPE, Boolean) { \
          /* Args:  x loLimit hiLimit includeLo includeHi coercedOut  inRangeOut */\
         VIVM_TRACE_FUNCTION(InRangeAndCoerce##TYPE)    \
-        _Param(5) = _Param(0) < _Param(1) ? _Param(1) : _Param(0) > _Param(2) ? _Param(2) : _Param(0); \
+        if (_ParamPointer(5)) { \
+            _Param(5) = _Param(0) < _Param(1) ? _Param(1) : _Param(0) > _Param(2) ? _Param(2) : _Param(0); \
+        } \
         _Param(6) = (_Param(0) > _Param(1) || (_Param(3) && _Param(0) == _Param(1)))    \
```
cglzaguilar: Do we need to check if _Param(6) is null? Why _Param(5) would be optional and not _Param(6). Please elaborate more on the scenario that requires this change. I wonder if we need to do this check for other primitives and not just InRangeAndCoerce.

source/core/TypeAndDataManager.cpp
sanmut: Agreed

source/core/Math.cpp
sanmut: Only _Param(5) is optional (wildcard) during CodeGen. _Param(6) is always used. It is this primitive specific for the CompareAggregates case.

cglzaguilar: I discussed this with Sankara and it is only this case in which we need to check for the output parameter, given how InRangeAndCoerce was generated for aggregates.

make-it/custom.mak
```diff
@@ -0,0 +1,9 @@
+# ********** IMPORTANT **********
+# All the contents of this file should be commented when checked in.
+# This file is ONLY for developers to turn on different flags for debugging
+# purposes in their local machines. Any changes to compiler flags that affect
+# distribution builds should be made in the actual Makefile's and not in this
+# custom.mak file.
+# *******************************
+# Turn the following flags on to debug access violation exceptions
+# EM_OPTFLAG = -s ASSERTIONS=2 -s SAFE_HEAP=1
```
rajsite: Can you also add an example for enabling debug builds?

EM_OPTFLAG = -g4

sanmut: Good idea. Thanks. Done.

PR #321, opened 2017-08-15
Fix to Defect DE13031 Instructions not generated for MaxAndMin

cglzaguilar:
Added code to consider cases when NaN was involved.


source/core/Math.cpp
```diff
@@ -327,13 +327,21 @@ DECLARE_SCALE2X_INTN_HELPER(Single)
     DECLARE_VIREO_PRIMITIVE2(IsGT0##TYPE, TYPE, Boolean, (_Param(1) = _Param(0) >  0)) \
     DECLARE_VIREO_PRIMITIVE2(IsNotANumPathRefnum##TYPE, TYPE, Boolean, (_Param(1) = ::isnan((double)_Param(0))) ) \
     DECLARE_VIREO_PRIMITIVE4(MaxAndMin##TYPE, TYPE, TYPE, TYPE, TYPE,    \
```
sanmut: Could you please add a one line comment about the arguments (input1, input2, max, min) for this and the one below? Let's just do that for every function/macro we modify to improve clarity.

cglzaguilar: Good idea. I tried adding // comments, but they don't seem to properly work inside the macro definition. I tried /* */ style comments and they work, but lint complains. I'll add the comments later on once I figure out a proper way to do it.

PR #322, opened 2017-08-15
Fix to defect DE12563 Stress VI with large block diagram crashes on run

cglzaguilar:
Removed hard code limits.


test-it/testList.json
```diff
@@ -31,7 +31,8 @@
         },
         "manual": {
             "tests": [
-                "StringFormatTime.via"
+                "StringFormatTime.via",
+                "TestVI_LargeBDTest.via"
```
cglzaguilar: Running "TestVI_LargeDBTest.via" on the framework causes Padding.via to fail on the second run. Padding.via fails with "Cannot enlarge memory arrays". I did some debugging and it appears that TestVI_LargeBDTest.via is not leaking. So something else is going on: memory fragmentation? Skipping for now.

sanmut: Could you please try to include the fix but exclude the test "TestVI_LargeBDTest.via" and see whether/how the tests fail.

cglzaguilar: Nothing fails when the test is included, but the fix is. My current changeset is adding the TestVI_LargeBDTest.via to the manual tests.

source/core/TypeTemplates.cpp
```diff
@@ -147,13 +147,14 @@ void TypeTemplateVisitor::VisitCluster(ClusterType* type)
     AggregateAlignmentCalculator* saveCalc = _alignmentCalculator;
     _alignmentCalculator = &calc;
 
-    TypeRef elementTypes[1000];   // TODO(PaulAustin): enforce limits or make them dynamic
     IntIndex subElementCount = type->SubElementCount();
+    TypeRef* elementTypes = new TypeRef[subElementCount];
```
sanmut: As discussed in person last week, why don't we make this a vector (as done above)?

source/core/TypeTemplates.cpp
```diff
@@ -180,13 +181,14 @@ void TypeTemplateVisitor::VisitEquivalence(EquivalenceType* type)
...
```
sanmut: why don't we make this a vector (as done above)?

source/core/TypeTemplates.cpp
cglzaguilar: Updated code to use vector instead.

source/core/TypeTemplates.cpp
cglzaguilar: Updated code to use vector instead.

test-it/testList.json
cglzaguilar: I just created a defect to look into the issue later on: DE14000 Vireo: Investigate why js test framework fails when TestVI_LargeBDTest.via is included.

PR #323, opened 2017-08-18
Fix InRangeAndCoerce defects and add tests

sanmut:


.gitignore
```diff
@@ -23,6 +23,7 @@ make-it/esh
 target-support/
 dist/
 custom.mak
+make-it/custom.mak
```
rajsite: If adding to the .gitignore then also delete custom.mak from the repo and add instructions to the README (preferred)

Otherwise don't include in .gitignore (also fine)

sanmut: You are right. Removed changes to .gitignore. Thanks.

source/core/GenericFunctions.cpp
```diff
@@ -679,36 +680,43 @@ InstructionCore* EmitMaxMinEltsInstruction(ClumpParseState* pInstructionBuilder)
...
 //------------------------------------------------------------
-static bool GetMinimumArrayDimensions(TypedArrayCoreRef sourceArray1, TypedArrayCoreRef sourceArray2,
+static bool GetMinimumArrayDimensions(std::vector<TypedArrayCoreRef> arrays,
     ArrayDimensionVector* newDimensionLengths, IntIndex* newRank)
 {
-    VIREO_ASSERT(sourceArray1 != null);
-    VIREO_ASSERT(sourceArray2 != null);
+    int arrayCount = arrays.size();
+    for (int i = 0; i < arrayCount; i++) {
+        VIREO_ASSERT(sourceArrays[i] != null);
```
cglzaguilar: I think sourceArrays is meant to be arrays. I wonder why you didn't get a compile error.

sanmut: Good catch. Fixed. We should try to turn on that and build.

PR #326, opened 2017-08-22
Content type default

rajsite:
Changes the Blob type from the default of empty string to application/x-www-form-urlencoded. This improves on the Coherent behavior of attempting to use a Content-Type of empty string and triggering a CORS preflight request.


cglzaguilar:
Check why Travis-ci is failing, before submitting.

PR #327, opened 2017-08-22
FloatToUInt Conversion Issue

gleonoliva:
Convert to unsigned integer converts negative floats to 0.


source/core/Math.cpp
```diff
@@ -413,6 +413,8 @@ DECLARE_SCALE2X_INTN_HELPER(Single)
             _Param(1) = numeric_limits<DEST>::max(); \
         } else if (::isinf(src)) { \
             _Param(1) = src < 0 ? numeric_limits<DEST>::min() : numeric_limits<DEST>::max(); \
+        } else if (src < numeric_limits<DEST>::min()) { \
+            _Param(1) = numeric_limits<DEST>::min(); \
```
sanmut: Do we need to test for src greater than DEST::max()?

sanmut: Please add the .via file and .vtr too.

sanmut:
Not a big deal. It is a good practice to to squash some of the commits if they are not providing great value as a separate commit. It is a good idea to have separate commits for different logical fixes, so forth.

README.md
```diff
@@ -21,6 +21,11 @@ Requirements:
 Setup:
 - `npm install`
 
+Setup (Windows):
```
cglzaguilar: For next time include unrelated changes to the fix in a different pull request. That makes them easier to understand.

test-it/testList.json
```diff
@@ -286,6 +286,7 @@
                 "TemplateBasics.via",
                 "TestVI_ArrayInsertPartial.via",
                 "TestVIArrayMaxAndMin.via",
+				"TestVIConversion.via",
```
cglzaguilar: Fix tab, other than looks good to me.

Issue #328, opened 2017-08-28
Clean-up build artifact output proposal


rajsite:
# Goals
1. Separate artifacts from source
2. Do not mix artifacts from different targets or builds
3. Include debug and release builds in GitHub Releases

# Supported Targets and Builds

Continue the conventions used for the GitHub Releases which follows the Clang / Rust platform triplets: https://forge.rust-lang.org/platform-support.html

## Targets
- [ ] 1686-pc-windows-msvc
- [ ] x86_64-unknown-linux-gnu
- [ ] asmjs-unknown-emscripten

## Builds
- [ ] debug
- [ ] release
- [ ] profile (asmjs-unknown-emscripten only unless the others have similar configs)

# Layout
Changing so that we are not mixing source build scripts with artifact output

`<vireo-sdk>/objs/<target>/<build>`
`<vireo-sdk>/dist/<target>/<build>`
`<vireo-sdk>/releases/esh_v<semver version>_<target>_<build>.<archive filetype>`

ex:
`<vireo-sdk>/dist/i686-pc-windows-msvc/esh.exe`
`<vireo-sdk>/releases/esh_v5.9.1_i686-pc-windows-msvc_debug.zip`

# Distributions

## GitHub Releases
Native releases will continue as is but with the full matrix above:
`esh_v<semver version>_<target>_<build>.<archive filetype>`

example:
`esh_v5.9.1_i686-pc-windows-msvc_debug.zip`

Emscripten releases will be bundled into the npm package and the default export will be release version of vireojs but the other builds will be included as well. TBD is how to expose the different builds in the JS environment.

PR #329, opened 2017-08-28
First pass adding vireo debug to package

rajsite:
Using approach discussed here: https://github.com/ni/VireoSDK/issues/328

This pull request covers only the emscripten part of the proposal


rajsite:
You may also notice that I changed from -Os for the build to -O3.

-Os is recommended by emscripten to shrink binary size at the cost of performance
-O3 is the recommended release setting by emscripten
https://kripken.github.io/emscripten-site/docs/tools_reference/emcc.html#emcc-o3

Running the JS benchmarks, -O3 gives on average 30% improvement on peek and poke tests into Vireo over -Os: https://docs.google.com/spreadsheets/d/1DJWLSjlsYiivtnl-os7fVtAL566Pcz034AdIvU7berY/edit?usp=sharing

The file size change is
-Os: 962 KB
-O3: 1095 KB
which is a 13% increase in file size or 133 KB increase which is pretty minor overall.

make-it/EmMakefile
```diff
@@ -50,8 +64,15 @@ help:
 #  Vireo has grown as well, so emscripten is getting better if the over all size has shrunk by about 20%
 
 EMCC= emcc
-EM_DBGFLAG = -g4
-EM_OPTFLAG = -Os # -Os -O3
+EM_OPTFLAG = -O3 #release config
+
+ifeq ($(BUILD), debug)
+   EM_OPTFLAG = -g4
```
cglzaguilar: Consider enabling flags already included in Custom.mak.

rajsite: Good catch

make-it/EmMakefile
```diff
@@ -50,19 +62,33 @@ help:
...
 
 EMCC= emcc
-EM_DBGFLAG = -g4
-EM_OPTFLAG = -Os # -Os -O3
+
+# release config
+EM_OPTFLAG = -Os
+
+ifeq ($(BUILD), debug)
+   EM_OPTFLAG = -g4 -s ASSERTIONS=2 -s SAFE_HEAP=1
```
sanmut: I think with this change , custom.mak should be deleted. It should not have been checked in originally.

make-it/EmMakefile
```diff
@@ -4,31 +4,43 @@
...
+VIREO_SDK=..
 
-OBJS=$(VIREO_SDK_DIR)/make-it/objs
-INCDIR=$(VIREO_SDK_DIR)/source/include
-BIN=$(VIREO_SDK_DIR)/bin
-TARGETDIR=$(VIREO_SDK_DIR)/dist
-CORESOURCEDIR=$(VIREO_SDK_DIR)/source/core
-IOSOURCEDIR=$(VIREO_SDK_DIR)/source/io
-UTSOURCEDIR=$(VIREO_SDK_DIR)/source/unittest
+TARGETS_SUPPORTED=asmjs-unknown-emscripten wasm32-unknown-emscripten
```
sanmut: We should add support for native builds also. Now, native builds create "Debug" and "Release" directories under "dist".

sanmut:
README.md should be updated with the new path to vireo.js as it is no longer under "dist" directly.

make-it/EmMakefile
rajsite: yep, as discussed in the summary: This pull request covers only the emscripten part of the proposal

Native will be updated by whoever wants to tackle it in the future. The proposal is here: https://github.com/ni/VireoSDK/issues/328

sanmut:
HelloNode.js should be updated? It has reference to old path for vireo.js.
                console.log('Error: ../dist/vireo.js not found (Maybe build it first?)');

sanmut:
It is likely beyond the scope of this change. I recommend adding a commandline  option for "test.js" to use "Debug" or "Release" version of vireo.js for running the VTR test suite.

sanmut:
Great effort. Thanks for doing this. Looks good.

make-it/EmMakefile
```diff
@@ -74,11 +74,6 @@ ifeq ($(BUILD), profile)
    EM_OPTFLAG = --profiling
 endif
 
-ifneq ($(wildcard custom.mak),)
```
sanmut: Sorry, may be, I was not clear. We should only remove the custom.mak file from source control. The references to it in the Makefile's should stay as-is before this commit. It is an internal build tool convention @spathiwa used in the make files. custom.mak is used locally by developers to override flags and build with that configuration as default. So, the convention is to use custom.mak the way it is used in Makefiles but not check in the file in source control but add the path of the custom.mak files to .gitignore so that it does not annoy the developers who use that. I broke that convention misremembering the usage by checking in the file. This change and the change below in make-it/Makefile should continue to reference custom.mak as before. Sorry for the confusion I created on this to begin with.

PR #330, opened 2017-08-28
Change optimization level

rajsite:
Changed from -Os for the build to -O3.

-Os is used by emscripten to shrink binary size at the cost of performance
-O3 is the recommended release setting by emscripten

https://kripken.github.io/emscripten-site/docs/tools_reference/emcc.html#emcc-o3

Running the JS benchmarks, -O3 gives on average 30% improvement on peek
and poke tests into Vireo over -Os:

https://docs.google.com/spreadsheets/d/1DJWLSjlsYiivtnl-os7fVtAL566Pcz034AdIvU7berY/edit?usp=sharing

The file size change is
-Os: 962 KB
-O3: 1095 KB
which is a 13% increase in file size or 133 KB increase which is pretty minor overall.

### DIFF
--- a/make-it/EmMakefile
+++ b/make-it/EmMakefile
@@ -64,7 +64,7 @@ help:
 EMCC= emcc
 
 # release config
-EM_OPTFLAG = -Os
+EM_OPTFLAG = -O3
 
 ifeq ($(BUILD), debug)
    EM_OPTFLAG = -g4 -s ASSERTIONS=2 -s SAFE_HEAP=1


rajsite:
As this is not an api change I believe this can be a minor revision from v6.0.0 to v6.1.0.

There seems to be disagreement about this so please state what you prefer.

sanmut:
I think we should test this change out of a release by itself, run the test suite on github side, use this on LV side and run the test suite. It may be a good idea to run some performance benchmark tests also. We should check this in if all that looks good. I don't see a reason to upgrade Major version though.

rajsite:
> I think we should test this change out of a release by itself, run the test suite on github side, use this on LV side and run the test suite.

Yea, we discussed this in person. Of course the GitHub tests will run and I'll queue an LVjs build before submission. This is our standard practice. Was there something else you were considering?

> It may be a good idea to run some performance benchmark tests also.

I ran the performance benchmark and linked results at the top. Was there something else you were considering?

> We should check this in if all that looks good. I don't see a reason to upgrade Major version though.

I'll take that as another vote for a minor revision.

rajsite:
As a reference it looks like this was the change that switched from -O3 to -Os initially: https://github.com/ni/VireoSDK/commit/c9882e771a26bdf840c118a298117008f1dff88c

Seems like at the time it was primarily motivated by file size.

PR #331, opened 2017-08-28
Fix to Defect DE14013. Uncaught Vireo / JS error (Instr. not generate…

cglzaguilar:
…d: Copy) when using Open+Read Tag from Skyline API

Root issue: Vireo is using more memory than originally requested when
emitting code for a copy. This behavior is random.

The EmitGenericCopyInstruction function is returning different memory
size values based on the alignment of the pointers, which are not always
the same in the requested and allocation phases. This behavior is random
and when present causes different memory sizes for the same instruction
causing a "copy instruction not generated" error.

Refactored code to not use CopyProcName, which provides the name of the
copy function to use and takes into consideration the pointer alignment
of source and destination, and use always CopyN. We loss some speed, but
gain on reliability.

### DIFF
--- a/source/core/GenericFunctions.cpp
+++ b/source/core/GenericFunctions.cpp
@@ -21,7 +21,10 @@ SDG
 
 namespace Vireo
 {
-
+//------------------------------------------------------------
+// Currently we are not calling this function, it intermittently returns different values 
+// between the first and second pass. We need to root cause the issue and reenable the code
+// This is tracked in github issue: https://github.com/ni/VireoSDK/issues/332
 //------------------------------------------------------------
 ConstCStr CopyProcName(void *pSource, void *pDest, Int32 aqSize)
 {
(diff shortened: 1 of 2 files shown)


spathiwa:
It seems to me if the issue is the line that checks source and dest pointer alignment and returns null, then we should investigate why the alignment is different, and whether the alignment check even makes sense on the vjs target; maybe it should only apply to native, if even there.  We don't currently support any backend targets where unaligned access would fault.
Except for this check, the Copy1, Copy2, etc. copy the same number of bytes as CopyN, but more efficiently, and I don't think we want to lose this performance

rajsite:
Just as a reference, the emscripten topic on memory alignment: https://kripken.github.io/emscripten-site/docs/porting/Debugging.html#memory-alignment-issues

> The Emscripten memory representation assumes loads and stores are aligned. Performing a normal load or store on an unaligned address can fail.

But it sounds like this behavior happens in native msvc builds as well as js.

sanmut:
spathiwa wrote "I don't think we want to lose this performance".
We agreed to check this in to unblock us from testing. This should be investigated further.

PR #334, opened 2017-08-31
Better fix for DE14013: Limit forced alignment to pointer-size in CopyProcName.

spathiwa:
A 32-byte cluster, e.g., is not allocated 32-bye aligned and does not
need to be, but this test was trying to enforce this, and if
the data pointer was reallocated between preflight and instruction
allocation and the unreasonably-expected alignment didn't match,
it would generate a CopyN instead of the Copy4, etc. that was
allocated space for in preflight, memory overwrite or crashes.

Add tests.  The original SkyLine test would not run in node.js due to HttpClient dependencies, but the problem still occurs if you replace the Http prim calls with stub subVIs with the same param signature.  So I did this and added the resulting test to testList.json.


source/core/GenericFunctions.cpp
```diff
@@ -29,7 +29,8 @@ namespace Vireo
 ConstCStr CopyProcName(void *pSource, void *pDest, Int32 aqSize)
 {
     // If the source or dest are not aligned to aqSize bytes, return null.
-    if (((uintptr_t)pSource % aqSize != 0) || ((uintptr_t)pDest % aqSize != 0))
+    Int32 alignment = aqSize > sizeof(void*) ? sizeof(void*) : aqSize;
```
cglzaguilar: Consider updating the comment to give a better explanation as to why we are doing this.

source/core/GenericFunctions.cpp
```diff
@@ -693,17 +697,19 @@ InstructionCore* EmitMaxMinEltsInstruction(ClumpParseState* pInstructionBuilder)
 static bool GetMinimumArrayDimensions(std::vector<TypedArrayCoreRef> arrays,
     ArrayDimensionVector* newDimensionLengths, IntIndex* newRank)
 {
-    int arrayCount = arrays.size();
-    for (int i = 0; i < arrayCount; i++) {
+    size_t arrayCount = arrays.size();
```
cglzaguilar: I think this refactor is unrelated to the main issue. Consider separating this code change in an independent pull request.

spathiwa: It is only a warning fix for unused code which I fixed while touching this file.
Seemed easier just to slip it in, as it's a no-op and shouldn't really need review.

source/core/GenericFunctions.cpp
spathiwa: Done.  Also called AQAlignment instead of computing alignment manually.

source/core/GenericFunctions.cpp
```diff
@@ -21,15 +21,14 @@ SDG
...
-    // If the source or dest are not aligned to aqSize bytes, return null.
-    Int32 alignment = aqSize > sizeof(void*) ? sizeof(void*) : aqSize;
+ConstCStr CopyProcName(void *pSource, void *pDest, TypeRef elemType)
+{
+    Int32 aqSize = elemType->TopAQSize();
+    // If the source or dest are not aligned to the alignment required by the Copy proc
+    // for this size, return null.
+    // (Just because an object is 4 bytes doesn't mean it's an integer and appropriately aligned;
+    // it could be an unaligned cluster of 4 Booleans).
+    Int32 alignment = elemType->TheTypeManager()->AQAlignment(aqSize);
```
sanmut: Good call to sue AQAlignment. Thanks.

test-it/ViaTests/CopyAlignmentError.via
```diff
@@ -0,0 +1,3586 @@
+// Autogenerated Vireo assembly file
```
sanmut: I would recommend adding a printf statement at the end of the first clump in the toplevel VI that just prints a string and check for that in the .vtr file. With the behavior differences between Debug and Release now and in the future and native vs JS (especially JS), there is a good chance that the Vireo may exit prematurely on error condition in the middle of executing/parsing a program without printing anything to console. A printf at the end is a safer bet, I think.

spathiwa: Good point.  I think we'd still notice the crash, but this is good just to be safe.
Done.

PR #336, opened 2017-09-01
Add support for converting error cluster to bool

sanmut:
Needed to make coercions work on many nodes on the diagram


sanmut:
> DFIR should just emit "error.status" whenever an error is used in a Boolean context.

I did consider that prior to doing this. However, the changes would be a bit involved without adding an additional instruction (like this approach does) -- which is more effort and riskier. We emit convert instruction on coercion node in DFIR graph. We have already other Convert prims in Vireo to convert Timestamps and enums. This follows the existing architecture. We can revisit later to optimize saving an instruction in the future.

spathiwa:
Disagree.  This is not a conversion, it is a cluster field access.  Other conversions need to be conversions because they actually convert to different types and need a dataspace allocation to store the result.  I think this prim fundamentally does not make sense in Vireo.  Please reconsider DFIR change, or more fully explain how this could introduce risk.

cglzaguilar:
@sanmut is DFIR already emitting a conversion instruction for the error cluster to boolean case?

sanmut:
> is DFIR already emitting a conversion instruction

Yes. Then, it is a Vireo error that this conversion instruction does not exist. Per discussion, I have created the issue for us to revisit later. 
https://github.com/ni/VireoSDK/issues/338

spathiwa:
Approved to go in for now, with issue filed.

PR #339, opened 2017-09-01
SubString ReadSubexpressionToken was reading past the end of its input.

spathiwa:


source/core/StringUtilities.cpp
```diff
@@ -669,7 +669,7 @@ TokenTraits SubString::ReadSubexpressionToken(SubString* token)
     } while (tt && (depth > 0));
 
     // Look for template parameters
-    while (*_begin == '<') {
+    while (_begin < _end && *_begin == '<') {
```
sanmut: Good catch

PR #341, opened 2017-09-06
Fix to CAR 663663 Booleans aren't stopping loops

cglzaguilar:
Created a instruction that will copy value and reset on a single Vireo
instruction


source/core/GenericFunctions.cpp
```diff
@@ -2676,6 +2686,9 @@ DEFINE_VIREO_BEGIN(Generics)
     DEFINE_VIREO_FUNCTION(CopyEnum4, "p(i(UInt32)  o(UInt32) i(UInt32))");
     DEFINE_VIREO_FUNCTION(CopyEnum8, "p(i(UInt64)  o(UInt64) i(UInt64))");
 
+    // Instruction to copy boolean value and reset it to another value
+    DEFINE_VIREO_FUNCTION(CopyAndReset, "p(io(Boolean from)  o(Boolean to) i(Boolean fromNewValue))");
```
spathiwa: 'fromNewValue' reads strangely.   I'd suggest renaming params to source, dest, and sourceNewValue or sourceResetValue.

source/core/GenericFunctions.cpp
```diff
@@ -2650,6 +2650,16 @@ VIREO_FUNCTION_SIGNATUREV(MergeErrors, MergeErrorsParamBlock)
 }
 
 //------------------------------------------------------------
+// We need this operation on a single vireo instruction so boolean latched buttons work correctly
+// from(0), to(1), fromNewValue(2)
```
sanmut: nitpick --- fromNewValue is a bit confusing to read. I would do "src", "dest", "newValueForSrc"---  nitpick.  I am ok to submit as is.

source/core/GenericFunctions.cpp
sanmut: Wow. Perfect coincidence on the comment :-)

PR #342, opened 2017-09-06
Fix bugs in Spreadsheet String to Array

spathiwa:
(for CAR 661959)
1) Fix a bug in SubString::FindFirstMatch which failed to find an exact
match at the tail end of a string (e.g. match 'u' in 'aeiou').
2) Fix matching of %s in Spreadsheet String to Array to not stop at
whitespace, since this prim has different meaning than Scan From String.
3) Add support for passing an array of delimiter strings
3) Intentionally replicate a minor bug in CG and desktop NCG that causes a
partial match of a delimiter string at the end of the input to ignore the
final delimiter when a complete match would occur later in the delim array.
(See final two cases in TestVI_MultipleDelimiters.via).
We should consider fixing this in both, but there's a danger of
breaking existing customer code in CG so it may be better to leave
this unaddressed and document it; parsing behavior would change
subtlely and return an extra default default element at the end of the
each row in this corner case.

Add tests.


test-it/ViaTests/TestVI_MultipleDelimiters.via
```diff
@@ -0,0 +1,116 @@
+// Autogenerated Vireo assembly file
```
cglzaguilar: Was it your intention to add this test? it seems unrelated to the change and I don't see it in the testList.json file.

source/core/NumericString.cpp
```diff
@@ -2593,15 +2646,28 @@ void ScanSpreadsheet(StringRef inputString, StringRef formatString, StringRef de
     }
 }
 //-------------------------------------------------------------------
-VIREO_FUNCTION_SIGNATURE4(SpreadsheetStringtoArray, StringRef, StringRef, StringRef, TypedArrayCoreRef)
+VIREO_FUNCTION_SIGNATURE4(SpreadsheetStringtoArray, StringRef, StringRef, TypedArrayCoreRef, TypedArrayCoreRef)
```
cglzaguilar: Shouldn't the fourth parameter still be a StringRef?

source/core/NumericString.cpp
```diff
@@ -2593,15 +2646,28 @@ void ScanSpreadsheet(StringRef inputString, StringRef formatString, StringRef de
...
+{
+    TypedArrayCoreRef outputArray = _Param(3);
+    if (_Param(2)->Length() ==0) {
```
cglzaguilar: I think this code can be further simplified by setting delimiter to \t or _Param(2).

source/core/NumericString.cpp
```diff
@@ -2408,14 +2408,65 @@ VIREO_FUNCTION_SIGNATURE4(ArraySpreadsheet, StringRef, StringRef, StringRef, Typ
...
+            }
+        }
+        *next = pastMatch;
+        return index;
+    }
+}
+
+// Wrapper around FormatScan:  %s for Spreadsheet Strings does not stop at whitespace the way it does
+// for the other format functions.
+static inline void FormatScanSS(SubString *elemString, SubString *format,  StaticTypeAndData *Value, Boolean formatIsString) {
```
cglzaguilar: Consider renaming to FormatScanForSpreadSheet. When looking at SS I was wondering about the meaning, until I figured out, but it is an unnecessary distraction when reading code.

test-it/ViaTests/TestVI_MultipleDelimiters.via
spathiwa: It is in the testList.json file, look again.

source/core/NumericString.cpp
spathiwa: Yes, I made this change before I made a separate function to handle the array case; I can back out of that.

source/core/NumericString.cpp
spathiwa: This isn't really an option because it would rely on construction of the STACK_VAR variable even in the case where it wouldn't be used.  Currently it's only used in the block where _Param(2) is zero length.

test-it/ViaTests/TestVI_MultipleDelimiters.via
spathiwa: And it's not unrelated, Spreadsheet to Array now takes an array of delimiters.

source/core/NumericString.cpp
spathiwa: Fixed.

source/core/NumericString.cpp
spathiwa: Done.

test-it/ViaTests/OctalTest.via
```diff
@@ -0,0 +1,107 @@
+// Autogenerated Vireo assembly file
```
cglzaguilar: Sorry, I was refering to this file, OctalTest.via

test-it/ViaTests/TestVI_MultipleDelimiters.via
cglzaguilar: Sorry, my intention was to refer to OctalTest.via, this one is OK.

test-it/ViaTests/OctalTest.via
cglzaguilar: It seems unrelated to the change and I don't see it in the testList.json file.

spathiwa: Oh, ok. I'll remove that.

source/core/NumericString.cpp
```diff
@@ -2596,12 +2649,25 @@ void ScanSpreadsheet(StringRef inputString, StringRef formatString, StringRef de
...
+    if (_Param(2)->Length() == 0) {
+        STACK_VAR(String, delimiter);
+        delimiter.Value->AppendCStr("\t");
+        ScanSpreadsheet(_Param(0), _Param(1), delimiter.Value, outputArray, false);
+    } else {
+        ScanSpreadsheet(_Param(0), _Param(1), _Param(2), outputArray, false);
+    }
+    return _NextInstruction();
+}
+//-------------------------------------------------------------------
```
cglzaguilar: Now that you're touching this function, please add a comment for the parameters

spathiwa: Done.

source/core/StringUtilities.cpp
```diff
@@ -1075,13 +1075,13 @@ void SubString::TrimQuotedString(TokenTraits tt)
...
     IntIndex searchStringLength = searchString->Length();
-    if (searchStringLength > Length())
+    if (Length() == 0 || searchStringLength > Length())
         return -1;
     if (offset < 0) {
         offset = 0;
     }
     const Utf8Char* pStart = _begin + offset;
-    const Utf8Char* pEnd = _end - searchStringLength;
+    const Utf8Char* pEnd = _end - searchStringLength+1;
```
sanmut: I am assuming the intent here is not "_end - (searchStringLength+1)". With associativity of + and - operators, this code is misleading to read with no spacing around +.

sanmut: Good catch. Looks like SearchSplitString and SearchAndReplaceString are affected by this change and may exhibit the same problem. I would recommend adding a couple of test cases that shows the problem without this fix to test-it\ViaTests\SearchSplitString.via. Carlos or I will be able to create a .via file with tests for SearchSplitString which may also be checked in.

spathiwa: Fixed spacing and added test case to SearchSplitString.via.
I do not think a "couple" of test cases are necessary.

spathiwa:
Is there anything still holding this up from being approved?

sanmut:
@spathiwa , Nothing from my side. I just left it unapproved until we determine that we can merge this and no outstanding showstopper needs to be fixed. Will check on that.

cglzaguilar:
I found a regression with the fix on FindFirstMatch while creating tests cases for SearchSplitString. Offsets that go beyond the input length size not longer work. See attached zip file. We need to either fix FindFirstMatch or SearchSplitString.

[SearchSplitStringRegression.zip](https://github.com/ni/VireoSDK/files/1341918/SearchSplitStringRegression.zip)

PR #343, opened 2017-09-13
Vireo fix for CAR 666144

spathiwa:
CAR description: WebVI 2D array controls do not show any data when the array
wired to them was built out of 1D arrays of different sizes.

(Vireo was crashing because ArrayConcatenate wasn't inflating non-flat
array default elements (e.g. strings) when filling in default elements
for uneven length input arrays.


sanmut:
Using BuildArray and IndexArray with QueueRefnum's still does not work. I wonder if we are missing a similar fix in other potential places? I have attached a test (TestVI_ArrayOfRefs.via) here. A simplified VIA file SimpleArrayOfRefs.via is sufficient to reproduce the crash.
[ArrayOfReferences.zip](https://github.com/ni/VireoSDK/files/1303249/ArrayOfReferences.zip)

cglzaguilar:
It seems that SimpleArrayOfRefs.via is not executing the same code path as the previous issue, since it is not dealing with adding arrays, just adding elements.

This might require additional code to be fixed, which is going to make this change riskier.

Any thoughts Craig?

spathiwa:
I don't think if affects the risk of the original change.  We should consider them separarely.

cglzaguilar:
All right then. Sankara, if we consider the new finding separately, are you ok on moving forward with this pull request? We already have a passing lvjs-ci, so I don't think there are issues with this change.

I would like to generate a new Vireo release and start the integration today to address the original problem of the CAR.

PR #344, opened 2017-09-15
CAR 666144 addendum for array of queues.

spathiwa:
There's two issue here:

Issue 1:
CopyData wasn't correctly implemented for RefNumValType.  It was
incorrectly called the base class wrapped type, which treated the refnum
data as an array (which it's not) and crashing.  (The refnum does hold
onto an array type to manage the refnum's queue, but the array is stored
out of line in the refnum manager, accessed via refnum cookie, and
not directly in the data).  This part of the fix is straight-forward,
was obviously wrong before, and is safe.

Issue 2:
ArrayConcatenate emits an array of booleans after the input element arguments
to identify whether each of inputs is an array or a scalar element.
EmitArrayConcatenateInstruction() calls CompareType with the output type
for each input type to decide this.  CompareType(A,B) ultimately will
return true if either IsA(A,B) or IsA(B,A), and since a refnum's type is
built out of an array type so it can remember the element type and
possible static size of the queue, IsA(refnum, array) winds up returning
true.  The fix for this is to shortcircuit the base-type loop in IsA and
make it stop when it encounters a refnum, considering it opaque.
(Probably PointerType should do this too.)

(The IsA change explicitly tests for refnums, so it should be limited in
scope.  But it's a bit more risky than the other array of non-flat data fix
because it touches a very commonly used function.)


source/include/TypeAndDataManager.h
```diff
@@ -951,8 +951,12 @@ class RefNumValType : public WrappedType
         *(RefNumValType**)pData = this;
         return kNIError_Success;
     }
-    virtual NIError ClearData(void* pData) { return kNIError_Success; }
-    virtual void*   Begin(PointerAccessEnum mode)       { return (void*)&_refnum; }
+    virtual NIError CopyData(const void* pData, void* pDataCopy)  {
+        *(RefNumValType**)pDataCopy = *(RefNumValType**)pData;
```
cglzaguilar: Just a question, here I assume we copy the pointer to the class, I wonder if we need to create another instance and then copy the data from one to the other, actually creating a duplicate. Does it make sense?

test-it/testList.json
```diff
@@ -288,6 +288,7 @@
                 "TemplateBasics.via",
                 "TestVI_ArrayInsertPartial.via",
 		"TestVI_ArrayNonFlatInflate.via",
+		"TestVI_ArrayOfRefs.via",
```
cglzaguilar: Do we need to fix a tab here? It seems the new line and the previous one are not aligned.

source/core/TypeAndDataManager.cpp
```diff
@@ -852,6 +852,8 @@ Boolean TypeCommon::IsA(const SubString *otherTypeName)
         if (t->Name().Compare(otherTypeName)) {  // ??? should this really consider unnamed types equal?? -CS
             return true;
         }
+        if (t->BitEncoding() == kEncoding_RefNum)
```
sanmut: Should we do this check instead similar to checking thisEncoding and otherEncoding for  kEncoding_Array and kEncoding_Cluster in the two functions below?

Boolean TypeCommon::IsA(TypeRef otherType, Boolean compatibleStructure)
Boolean TypeCommon::CompareType(TypeRef otherType)

This particular IsA() seems very generic and polluting this with specific encoded type does not look good to me.

test-it/testList.json
spathiwa: Fixed.

source/core/TypeAndDataManager.cpp
spathiwa: That other code is checking for intrinsics, and doesn't have  to deal with type inheritance.
I understand your concern, but unfortnately this check has to be here because this is the only place that it iterating through the type inheritance chain to look for a common base type, and that is the source of the problem.
I think we could improve this by adding a type predicate like "IsOpaqueReferencingType()" that could return true for refnums and possibly others such PointerType, and change this explicit check to use that.

source/include/TypeAndDataManager.h
spathiwa: I noticed this when I added CopyData too.  I think this probably warrants more investigation.
This works currently because of the way we generate Vireo from DFIR, where every call to Obtain_Queue has its own unique storage, but this implementation could have an aliasing problem.

source/core/TypeAndDataManager.cpp
spathiwa: Oh, and yes, isFlat = true is correct.  The data is an integer refnum value and the max queue size.

sanmut: Please add "IsOpaqueReferencingType()" and leave a TODO to check other types for PointerType. If possible, I would recommend adding tests that may fail now and keep them disabled too -- while this is fresh in the mind.

source/include/TypeAndDataManager.h
sanmut: If possible, I would recommend adding potential tests that may fail due to aliasing issues (and keep them disabled) -- while this is fresh in the mind.

source/core/TypeAndDataManager.cpp
spathiwa: Added this.
PointerType is internal and doesn't exist in via files so it's not possible to write a test for this, but I think because of this we never call IsA on PointerType either so it should not matter.  Mentioned this in the TODO.

source/include/TypeAndDataManager.h
spathiwa: I verified we can get weird aliasing behavior.  I'll work on making a test out of it.  I may wait to submit it until I have a fix as well.

spathiwa: Refactored refnum storage to avoid aliasing behavior so that queue refnums work inside arrays and clusters.

cglzaguilar:
Update: I'm currently doing CI validation on this pull request.

cglzaguilar:
LVjs CI validation passed with this pull request changes. Craig, please add the following tests to your pull request:
[TestVI_ArrayOfRefs.zip](https://github.com/ni/VireoSDK/files/1369708/TestVI_ArrayOfRefs.zip)

After that I'll merge this changeset.

PR #347, opened 2017-10-05
Refactor executeSlices to return a different result value allowing for more efficient scheduling, sleeping when nothing to do.

spathiwa:
A negative return value means there is still work to schedule, caller can recall executeSlices
immediately.
A zero return value means all clumps are finished with nothing waiting
on timers; VI is done, don't call again unless you load a new via file.
A positive return value is a time delay in milliseconds before the next
clump needs to be waken up; the caller can use an OS-specific sleep call
or schedule itself asynchronously after a delay before calling
executeSlices again.

Change native caller (esh) main to handle the new result and sleep when
appropriate, using a new Platform.Timer API.

Change VJS callers to handle new result:
  Karma TestHelpers.VireoRunner was already scheduling executeSlices
asynchronously, it just needed to be modifed to use the given timeout
instead of 0.
  test.js has been refactored so the test driver is asynchronous and
can sleep when executeSlices tells it to.
  The manual http tests have been modified to check that the executeSlices
return value is != 0, rather than > 0, and to use the returned delay
in their setTimeout call iff they were already asynchronous.

There are no new tests added here rather than the 'meta'-test that all
the test runners continue to run their tests correctly.
This is a pure refactor such that no via code changes functionality,
it just uses CPU more efficiently.  To test manually, run tests
which have Waits or blocking calls, and monitor CPU during the run
to verify CPU does not go to 100%/#processors.

### DIFF
--- a/source/include/ExecutionContext.h
+++ b/source/include/ExecutionContext.h
@@ -63,8 +63,13 @@ enum ExecutionState
     kExecutionState_None = 0,
     kExecutionState_ClumpsInRunQueue = 0x01,
     kExecutionState_ClumpsWaitingOnTime = 0x02,
-    kExecutionState_ClumpsWaitingOnQueues = 0x04,
-    kExecutionState_ClumpsWaitingOnISRs = 0x08,
+};
+
+enum ExecSlicesResult {
+    kExecSlices_ClumpsWaiting = -2,
+    kExecSlices_ClumpsInRunQueue = -1,
+    kExecSlices_ClumpsFinished = 0,
+    // ... or positive value indicating clumps waiting for specific time (in ms)
 };
 
 // Each thread can have at most one ExecutionContext (ECs). ExecutionContexts can work
@@ -114,7 +119,7 @@ class ExecutionContext
     ECONTEXT    void            ExecuteFunction(FunctionClump* fclump);  // Run a simple function to completion.
 
     // Run the concurrent execution system for a short period of time
-    ECONTEXT    ExecutionState  ExecuteSlices(Int32 numSlices, PlatformTickType tickCount);
+    ECONTEXT    Int32 /*ExecSlicesResult*/ ExecuteSlices(Int32 numSlices, PlatformTickType tickCount);
     ECONTEXT    InstructionCore* SuspendRunningQueueElt(InstructionCore* whereToWakeUp);
     ECONTEXT    InstructionCore* Stop();
     ECONTEXT    void            ClearBreakout() { _breakoutCount = 0; }
--- a/test-it/test.js
+++ b/test-it/test.js
@@ -197,7 +197,7 @@
             }
         }
         var hrstart = process.hrtime();
-        tester(testName, function(newResults) {
+        tester(testName, function (newResults) {
             var hrend = process.hrtime(hrstart);
             var msec = hrend[1] / 1000000;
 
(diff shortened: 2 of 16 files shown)


rajsite:
@spathiwa 

Please do not add spurious eslint configuration rules because you disagree with the configuration. It pollutes the code base and mixes style for no reason.

Also the language is not appreciated; please don't create commit messages like the following again:

> Disable no negated condition eslint rule. It's retarded.

There is more detail about the eslint rule here: https://eslint.org/docs/rules/no-negated-condition

Please remove the per file eslint disables that were added and follow the style configuration. The linting is of course not a hard and fast rule and where it can severally [harm code readability](https://github.com/ni/VireoSDK/blob/d7a0a00dfeb75b61be3df7893e374a98266c0844/source/io/module_eggShell.js#L29) exceptions [are made](https://github.com/ni/VireoSDK/blob/9d356e2b8df5e4a6a04d3421e0e2073e726e813c/source/core/module_coreHelpers.js#L99) but the tone of your message does not seem to make this the case.

If you have a technical reason we can discuss it further (if you have a personal preference, frankly, we have both already wasted too much time discussing it).

Frankly

You may also find it valuable to run the linters locally in [your development environment](https://medium.com/@hpux/vim-and-eslint-16fa08cc580f) or as part of your workflow with either [`npm run lint`](https://github.com/ni/VireoSDK/blob/d7a0a00dfeb75b61be3df7893e374a98266c0844/package.json#L30) to just run linting or [`npm run test`](https://github.com/ni/VireoSDK/blob/d7a0a00dfeb75b61be3df7893e374a98266c0844/package.json#L29) to lint and test in Chrome. This can help shorten the round-trip time from the CI.

Thank you for your understanding.

source/include/ExecutionContext.h
```diff
@@ -63,8 +63,13 @@ enum ExecutionState
     kExecutionState_None = 0,
     kExecutionState_ClumpsInRunQueue = 0x01,
     kExecutionState_ClumpsWaitingOnTime = 0x02,
-    kExecutionState_ClumpsWaitingOnQueues = 0x04,
-    kExecutionState_ClumpsWaitingOnISRs = 0x08,
+};
+
+enum ExecSlicesResult {
```
rajsite: I don't have a lot of C++ experience, but can you define ExecSlicesResult as a union of an enum and an Int32 so that ExecuteSlices can actually return a value of that type?

Otherwise it seems strange to have the enum defined if it is not actually used anywhere, maybe it should just be commented out?

test-it/test.js
```diff
@@ -197,47 +198,71 @@
...
+            testOutput = testOutput + text + '\n';
+        });
 
-        vireo.stdout = '';
         if (viaCode !== '' && vireo.loadVia(viaCode) === 0) {
             // TODO mraj because this is running synchronously we cannot do tests that rely on asynchronous results like http
-            while (vireo.executeSlices(1000000)) {
-                // Nothing here
-            }
+            // TODO spathiwa I think we can now...
```
rajsite: lol almost, we would need to add an XHR polyfill to the global scope in node similar to the following: https://github.com/rajsite/vireo-node-eightball/blob/f7bc85b065b9446b45fd02f38b1554dd79f3949d/server.js#L30

source/include/ExecutionContext.h
spathiwa: I hear you, but unfortunately, there's no good solution to this in C++.  Normally you might create a small wrapper class for it, and make the known-values predicates, but that would be overkill here and may not be interoperable with JS as a return value.

rajsite:
> I fully planned to revisit it, I thought, before anyone even saw it, because you told me you wouldn't look at it until you came back!

I miss understood, I assumed since the pull request was open it was intended for review. Might be valuable to state in a comment that it is still in progress or close and open a new pull request when ready to review.

> As far as my comment, I call things like I see them. Are you seriously offended by this term as applied to inanimate code? I could change it to moronic or just plain stupid if you would prefer, but I'm not censoring

1. Yes I do find it inappropriate. Code is made by people and I think it is important to be respectful of people's work. If there is a concern it should be made in an informative manner that helps someone learn something new.

2. The comments make for poor commit notes:

   > Disable no negated condition eslint rule. It's retarded.

   Was not useful in describing why the rule was not valuable to us. The following commit was far more valuable:

   > Disable no negated condition eslint rule. 
   > It's a semantic rule, not a style rule, and could actively hurt code readability by forcing the inversion of if/else statements into a less logical order to silence the warning.

3. This development is occurring in a public forum where we are representatives of National Instruments and should behave as such. If the Vireo project grows into a larger community I would hope we can create an environment that is welcoming and inviting to all users.

Changing the global setting is a preferable change over placing inline disables throughout the code if we have decided to follow that style. 👍

PR #348, opened 2017-10-09
 Refactor executeSlices to return a different result value allowing for more efficient scheduling, sleeping when nothing to do.

spathiwa:


spathiwa:
(Btw, I  just rebased the new incoming with the array queue fixes into this and re-pushed it; sorry if that interfered with any reviews in progress.)

source/io/module_eggShell.js
```diff
@@ -373,6 +373,14 @@
         };
 
         Module.eggShell.executeSlices = publicAPI.eggShell.executeSlices = function (slices) {
+            // compatibility non-async entry point for callers which do not expect negative return values
+            var result = EggShell_ExecuteSlices(v_userShell, slices);
+            return (result !== 0) ? 1 : 0;
+        };
+
+        Module.eggShell.executeSlicesAsync = publicAPI.eggShell.executeSlicesAsync = function (slices) {
```
rajsite: I don't think we need to maintain both functions at this time. Our only client is desktop and I think it is more important at this time to build up a clean api.

Users of the npm js version are very likely to be using package management and I think it is sufficient to create a major version bump to signify the API change and include the change in the description.

If there is consensus with @cglzaguilar and @sanmut to preserve both endpoints then I think we need a different name. Usually the Async version of a JS function is expected to either take a callback or return a promise that resolves asynchronously to the current stack. This function runs synchronously and returns a value synchronously that is consumed on the same stack. A better name might be "executeSlicesAsNecessary" or "executeSlicesUntilWait". We should also add a deprecated comment to the executeSlices function and add a run once comment log that states the function is deprecated (use a flag to print a warning on first call).

source/include/Synchronization.h
```diff
@@ -69,12 +69,15 @@ class OccurrenceCore : public ObservableCore
...
 
+const Int32 kMaxExecWakeUpTime = 10000;  // (milliseconds).  10 seconds.
+
 //------------------------------------------------------------
 //! Timer object that clumps can wait on.
 class Timer : public ObservableCore
 {
  public:
     Boolean AnythingWaiting()                   { return _observerList != null; }
+    IntMax NextWakeUpTime()                     { return _observerList != null ? _observerList->_info : 0; }
```
cglzaguilar: By looking at the comment where _info is declared, _info contains state information, but in this case it appears we expect it to contain time information, is this correct?

source/io/module_eggShell.js
cglzaguilar: I don't think we need to keep both endpoints if there is not need.

cglzaguilar:
Note: LVjs-CI passed with current changes.

source/io/module_eggShell.js
spathiwa: I'm fine with executeSlicesUntilWait or similar.  But I was envisioning this change to be *very* temporary;  @cglzaguilar had issues running this on the CI using the modified executeSlices which can return negative values, and I wanted to see if the issues were solely due to the negative return values or not.   If so, it would be due to code which tested if the return value > 0 rather than !== 0.  There's no reason for such callers to have made this assumption, so I wouldn't be surprised if they were only written this way to avoid triggering that terrible eslint rule.

Assuming this code passes the CI, then rather than submitting this and introducing the fork, if you can go ahead and fix all executeSlices in the TFS code to test for non-zero instead of > 0  first (which is still compatible with the previous incarnation of executeSlices and more correct), I can just leave the single entry point.

source/include/Synchronization.h
spathiwa: Yes.  The _info field is used differently by different inherited classes of ObservableCore. 
For Timers, it's always a wake up timer value.  (For Occurrences, it's the next occurrence 'count'.  I didn't look at all of them.)

source/core/ExecutionContext.cpp
```diff
@@ -311,7 +310,7 @@ InstructionCore* ExecutionContext::SuspendRunningQueueElt(InstructionCore* nextI
     }
 }
 //------------------------------------------------------------
-ExecutionState ExecutionContext::ExecuteSlices(Int32 numSlices, PlatformTickType tickCount)
+Int32 /*ExecSlicesResult*/ ExecutionContext::ExecuteSlices(Int32 numSlices, PlatformTickType tickCount)
```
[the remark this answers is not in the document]
spathiwa: Added comments.

source/io/module_eggShell.js
rajsite: The TFS code has one usage of executeSlices to update in `niEggShell.js`:

```
retVal = this._vireo.eggShell.executeSlices(1000);
...
executionState.finished = (retVal > 0 ? false : true);
```

which can be updated as part of the integration of the new version of Vireo into TFS. So yes please remove the additional entry point that was added from this file and all the other JS files that were updated to use it.

In summary:
1. We will have just the `executeSlices` function with the new behavior
2. When the package is updated in TFS, that changeset will include the change for how the return value of executeSlices is used

rajsite: We just talked about it but I'm also okay with having the function name updated to executeSlicesUntilWait. The new name describes the behavior more precisely and having the name change is a more noticeable breaking change for users updating the package which is a win-win to me.

sanmut:
This is great work. I think we should just add an automated test to make sure this feature does not regress and this is one of those that are really hard to know if broken.
- For native Windows, either add command line arguments to esh or use esh-test and that will spit the CPU time used by the process when it exits (last line in main()) . For Windows, [GetProcessTimes](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683223(v=vs.85).aspx) can be used lpUserTime should give the CPU time. There are also performance counter APIs that may be used as well. Write one or two simple VIA programs that will only call WaitMilliseconds for 1 second and 2 seconds. Executing either program should finish with CPU time less than 100 ms with the fix, but will take more than 1 second and 2 seconds respectively without the fix.
- For Linux flavors, I think the above mechanism should work with calling /proc/<CurrentPID>/stat.
- For JS, I am not sure whether performance.now() can be used before and after calling executeSlicesUntilWait and until all slices/clumps are executed and return. @rajsite  would know better.

source/core/ExecutionContext.cpp
```diff
@@ -383,21 +385,42 @@ ExecutionState ExecutionContext::ExecuteSlices(Int32 numSlices, PlatformTickType
...
+    ExecutionState state = kExecutionState_None;
+    Int32 reply = kExecSlices_ClumpsFinished;
     if (!_runQueue.IsEmpty()) {
-        reply = (ExecutionState) (reply | kExecutionState_ClumpsInRunQueue);
+        state = kExecutionState_ClumpsInRunQueue;
+        reply = kExecSlices_ClumpsInRunQueue;
     }
     if (_timer.AnythingWaiting()) {
-        reply = (ExecutionState) (reply | kExecutionState_ClumpsWaitingOnTime);
+        state = (ExecutionState) (reply | kExecutionState_ClumpsWaitingOnTime);
```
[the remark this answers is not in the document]
spathiwa: The right hand side should be 'state'; thanks for catching that, I've fixed it.
But I left state a bit set intentionally; that's why I made a new enum for return value.
You can have both clumps executing currently and clumps waiting on time simultaneously and this allows us to represent that.

source/include/ExecutionContext.h
```diff
@@ -63,8 +63,13 @@ enum ExecutionState
     kExecutionState_None = 0,
     kExecutionState_ClumpsInRunQueue = 0x01,
     kExecutionState_ClumpsWaitingOnTime = 0x02,
-    kExecutionState_ClumpsWaitingOnQueues = 0x04,
-    kExecutionState_ClumpsWaitingOnISRs = 0x08,
+};
```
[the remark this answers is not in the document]
spathiwa: They are bitwise still, explained above.

source/core/ExecutionContext.cpp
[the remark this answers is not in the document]
sanmut: hmm, "state"and "ExecutionContext._state" seem to be only assigned and never used after this PR. Should we just remove it altogether?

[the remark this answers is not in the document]
spathiwa: You're right.  Code which could check this flag instead check the _runQueue and _timer subelements directly.  Might as well kill it, we can add it back if there becomes a need.

rajsite:
> This is great work. I think we should just add an automated test to make sure this feature does not regress and this is one of those that are really hard to know if broken.
> 
> For native Windows, either add command line arguments to esh or use esh-test and that will spit the CPU time used by the process when it exits (last line in main()) . For Windows, GetProcessTimes can be used lpUserTime should give the CPU time. There are also performance counter APIs that may be used as well. Write one or two simple VIA programs that will only call WaitMilliseconds for 1 second and 2 seconds. Executing either program should finish with CPU time less than 100 ms with the fix, but will take more than 1 second and 2 seconds respectively without the fix.
> For Linux flavors, I think the above mechanism should work with calling /proc//stat.
> For JS, I am not sure whether performance.now() can be used before and after calling executeSlicesUntilWait and until all slices/clumps are executed and return. @rajsite would know better.
> If this cannot be done in this PR, that's fine. But, we should track this as github issue / user story as a requirement for completion of this feature.

What if in ExecuteSlices we instead keep track of the last returned returned ExecutionState in the user shell along with a timestamp? With that information ExecuteSlices can determine if it is being called too frequently and print a warning.

This would be helpful in general for Vireo integrators to be warned that they can pump the runtime more efficiently.

It would probably need a check to see if the last value was invalidated, ie a Set_Occurence was called by the HTTP nodes between calls to ExecuteSlices so a warning is not necessary.

source/io/module_eggShell.js
```diff
@@ -372,7 +372,9 @@
             return result;
         };
 
-        Module.eggShell.executeSlices = publicAPI.eggShell.executeSlices = function (slices) {
+        Module.eggShell.executeSlicesUntilWait = publicAPI.eggShell.executeSlicesUntilWait = function (slices) {
+            // returns -1 if should be called again ASAP, 0 if nothing to run, or positive value N if okay
```
rajsite: Should the comment be "returns < 0 if should be called again ASAP"?

spathiwa: Changed.

test-it/esh.js
```diff
@@ -0,0 +1,58 @@
...
+        vireo.loadVia(text);
+    } catch (e) {
+        console.log('Usage: ' + command + ' [file.via]...');
+        if (arg.substring(0, 1) !== '-') {
+            console.log('Cannot open ' + arg);
+        }
+        process.exit(1);
+    }
+
+    var execVireo = function () {
```
rajsite: Could make this a bit cleaner using setImmediate in node:

```
 (function executeVireo () {
    var state = vireo.executeSlicesUntilWait(100000);
    if (state > 0) {
        setTimeout(executeVireo, state);
    } else if (state < 0) {
        setImmediate(executeVireo); // Or even just calling executeVireo();
    }
}());

```

source/io/module_eggShell.js
spathiwa: Ugh, I hate how I can't reply directly to your non-review comment

> What if in ExecuteSlices we instead keep track of the last returned returned ExecutionState in the user shell along with a timestamp? With that information ExecuteSlices can determine if it is being called too frequently and print a warning.

> This would be helpful in general for Vireo integrators to be warned that they can pump the runtime more efficiently.

> It would probably need a check to see if the last value was invalidated, ie a Set_Occurence was called by the HTTP nodes between calls to ExecuteSlices so a warning is not necessary.

I like this idea but I don't think I want to try to do it in this PR.  I think I would make the check only if the run loop didn't have anything to do for multiple calls in a row, and the previous return value was > 0.  (A call to Set_Ocurrence would schedule something, so the run loop would run and reset the counter.)

test-it/esh.js
spathiwa: Agreed; done.

rajsite:
I think I may have found one issue. If I put a console log to print `timeUntilNextClump` here:

https://github.com/spathiwa/VireoSDK/blob/bc7836083a81e396d23d4b97ca0c8bc5527dc61f/test-it/karma/utilities/TestHelpers.VireoRunner.js#L47

And then focus the following test by changing `it` to `fit`: 

https://github.com/spathiwa/VireoSDK/blob/bc7836083a81e396d23d4b97ca0c8bc5527dc61f/test-it/karma/longrun/Timeout.Test.js#L102

Then execute the test using npm `run karma -- --browsers Chrome`

I can see that the function is getting called very frequently and timeUntilNextClump is always zero. From previous discussion I think I would expect kMaxExecWakeUpTime instead of zero.

The following is the via code that the test starts with before writing new values: https://github.com/spathiwa/VireoSDK/blob/bc7836083a81e396d23d4b97ca0c8bc5527dc61f/test-it/karma/fixtures/http/GetMethod.via

spathiwa:
@rajsite I reproduced this and figured out the issue.
Unspecified time limit sleeps do delay for a  maxint time period, as we discussed before.  (It's actually an Int64, so it's 0x7fffffffffffffff).
The problem is that this is being added to the current time and OVERFLOWING the time value.  The stored absolute timestamp becomes negative, which then gets interpreted as a time already past when compared to the current time in the event loop, and ExecuteSlices returns 0 ("immediately").

Just changing the max constant to something smaller but still large works around the problem, but is probably not the best solution.

I'll work out what to do about it tomorrow...  Probably make the PlatformTickType unsigned and always do subtractions instead of inequality tests when doing  time comparisons.

source/core/ExecutionContext.cpp
```diff
@@ -383,21 +384,37 @@ ExecutionState ExecutionContext::ExecuteSlices(Int32 numSlices, PlatformTickType
...
 #ifdef VIREO_SINGLE_GLOBAL_CONTEXT
     // TODO(PaulAustin): check global memory manager for allocation errors
 #else
     if (THREAD_TADM()->_totalAllocationFailures > 0) {
-        reply = kExecutionState_None;
+        reply = kExecSlices_ClumpsFinished;
     }
 #endif
-    _state = reply;
     return reply;
```
rajsite: Do you think it is worth asserting that if reply is < 0 that it is either -1 or -2?

spathiwa: Well, not really.   I'm actually contemplating removing the distinction between those two values, I'm not sure there is a situation where a client would really need to distinguish between them, especially in JS where you can't wait < 4 ms, much less < 1 ms.

rajsite: Could it be helpful for with native embedders like raspberrypi? If so we could leave it as is. I don't feel strongly either way.

spathiwa: What do you think about scrapping -2, and swapping the meaning of -1 and 0?
This way the return value doesn't feel like it has disjointed meaning between the timeout and the other cases.
0 really means no timeout, call me again immediately.
-1 is a traditional error or EOF code, meaning don't call me again.  Maybe could scrap the enum altogether.

PR #350, opened 2017-11-01
Vireo early wake-up and API refactors

rajsite:
# JS API Changes

- Added `vireo.eggShell.maxExecWakeUpTime()` which returns the `kMaxExecWakeUpTime` value. Used by tests to parameterize expectations and may be valuable to advanced Vireo integrators (but just used for testing so far).

- Added `vireo.eggShell.setExecuteSlicesWakeUpCallback()` for a user to listen for wake-up events

- Added `vireo.eggShell.executeSlicesUntilClumpsFinished()` which runs Vireo async to completion

- Modified `vireo.eggShell.loadVia()` to throw on error instead of return a status code (almost no one except the karma and test.js test runners checked the status code amongst all the usages in Travis and TFS)

- Modified `vireo.httpClient.abortAllRunningRequests()` which was a per instance vireo method but incorrectly affected all running instances

- Added `vireo.httpClient.setXMLHttpRequestImplementation()` which allows a user to provide their own XHR Level 2 implementation per vireo instance. This is valuable in node.js to allow the user to provide an XHR implementation and during tests to simulate certain behaviors (ie quick network resolution, simulate network errors)

- Modified the Vireo loader script to provide a warning suggesting the need to compile Vireo when an exception occurs loading vireo.js

- Modified the default print behavior to perform console.log instead of console.debug. This is a better default as pretty much every user in TFS and Travis was overriding the default print to use console.log. It is also a better default as node.js does not support console.debug and Chrome 58+ started hiding console.debug messages.

- Fixed usage of HttpClient methods on node.js that include a body. The optimization was introduced to use Blobs and ArrayBuffers for XHR requests instead of converting to DOMStrings and was incompatible with node.js making not possible to use the HttpClient in node.js even with a hacked global polyfill. This change and added tests will help prevent breaking node.js in the future.

# Test Infrastructure Changes

- Modified test.js and esh.js to use the xhr2 node library to enable the HttpClient in the node.js environment

- Added two example HttpClient VTR tests and a new test group called httpbin to testList.json. This test group requires HttpBin to run locally and was enabled on both AppVeyor and Travis

# Version Recommendation

I think this should be a minor version bump. Most of the API changes are API additions or are modifications on uncommon code paths (it is uncommon to have good error checking for loadVia).

### DIFF
--- a/make-it/EmMakefile
+++ b/make-it/EmMakefile
@@ -98,6 +98,7 @@ EM_WRAP = --pre-js $(PREJS) --post-js $(POSTJS)
 
 EM_EXPORTS = -s EXPORTED_FUNCTIONS="[\
     '_Vireo_Version', \
+	'_Vireo_MaxExecWakeUpTime', \
     '_EggShell_Create', \
     '_EggShell_REPL',\
     '_EggShell_ExecuteSlices',\
--- a/make-it/EmMakefile
+++ b/make-it/EmMakefile
@@ -98,7 +98,7 @@ EM_WRAP = --pre-js $(PREJS) --post-js $(POSTJS)
 
 EM_EXPORTS = -s EXPORTED_FUNCTIONS="[\
     '_Vireo_Version', \
-	'_Vireo_MaxExecWakeUpTime', \
+    '_Vireo_MaxExecWakeUpTime', \
     '_EggShell_Create', \
     '_EggShell_REPL',\
     '_EggShell_ExecuteSlices',\
--- a/source/core/ExecutionContext.cpp
+++ b/source/core/ExecutionContext.cpp
@@ -312,12 +312,11 @@ InstructionCore* ExecutionContext::SuspendRunningQueueElt(InstructionCore* nextI
 // ExecuteSlices - execute instructions in run queue repeatedly (numSlices at a time before breaking out and checking
 // timers), until all clumps are finished or tickCount time is reached.
 // See enum ExecSlicesResult for explanation of return values, or comments below where result is set.
-Int32 /*ExecSlicesResult*/ ExecutionContext::ExecuteSlices(Int32 numSlices, PlatformTickType tickCount)
+Int32 /*ExecSlicesResult*/ ExecutionContext::ExecuteSlices(Int32 numSlices, Int32 millisecondsToRun)
 {
     VIREO_ASSERT((_runningQueueElt == null))
-
     PlatformTickType currentTime  = gPlatform.Timer.TickCount();
-    PlatformTickType breakOutTime = currentTime + tickCount;
+    PlatformTickType breakOutTime = currentTime + gPlatform.Timer.MicrosecondsToTickCount(millisecondsToRun * 1000);
 
     _timer.QuickCheckTimers(currentTime);
 
--- a/source/io/module_eggShell.js
+++ b/source/io/module_eggShell.js
@@ -454,21 +454,15 @@
         // If a callback (stdout, stderr) is provided, it will be run asynchronously to completion
         Module.eggShell.executeSlicesUntilClumpsFinished = publicAPI.eggShell.executeSlicesUntilClumpsFinished = function (callback) {
             var MAXIMUM_VIREO_EXECUTION_TIME_MS = 4;
-            var printText = '';
-            var printTextErr = '';
-            var timerToken;
 
-            var origPrint = Module.print;
-            var origPrintErr = Module.printErr;
+            var timerToken;
             var origExecuteSlicesWakeupCallback = Module.eggShell.executeSlicesWakeupCallback;
 
             var vireoFinished = function () {
-                Module.print = origPrint;
-                Module.printErr = origPrintErr;
                 Module.eggShell.executeSlicesWakeupCallback = origExecuteSlicesWakeupCallback;
 
                 if (typeof callback === 'function') {
-                    callback(printText, printTextErr);
+                    callback();
                 }
             };
 
@@ -493,16 +487,6 @@
                     timerToken = undefined;
                     setTimeout(vireoFinished, 0);
                 }
-            };
-
-            Module.print = function (text) {
-                printText += text + '\n';
-                origPrint(text);
-            };
-
-            Module.printErr = function (textErr) {
-                printTextErr += textErr + '\n';
-                origPrintErr(textErr);
             };
 
             Module.eggShell.executeSlicesWakeupCallback = function () {
--- a/source/io/module_eggShell.js
+++ b/source/io/module_eggShell.js
@@ -415,6 +415,7 @@
         };
 
         Module.eggShell.executeSlicesUntilWait = publicAPI.eggShell.executeSlicesUntilWait = function (slices) {
+            // Returns an ExecSlicesResult:
             // returns < 0 if should be called again ASAP, 0 if nothing to run, or positive value N if okay
             // to delay up to N milliseconds before calling again
             return EggShell_ExecuteSlices(v_userShell, slices);
@@ -451,7 +452,7 @@
         // Runs synchronously for a maximum of 4ms at a time to cooperate with most browser execution environments
         // A good starting point for most vireo uses but can be copied and modified as needed
         // callback (stdout, stderr)
-        Module.eggShell.executeSlicesToCompletion = publicAPI.eggShell.executeSlicesToCompletion = function (callback) {
+        Module.eggShell.executeSlicesUntilClumpsFinished = publicAPI.eggShell.executeSlicesUntilClumpsFinished = function (callback) {
             var printText = '';
             var printTextErr = '';
             var timerToken;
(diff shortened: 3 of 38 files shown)


source/io/module_eggShell.js
```diff
@@ -361,29 +386,133 @@
...
             var result = Module._EggShell_REPL(v_userShell, viaTextPointer, viaTextLength);
             Module._free(viaTextPointer);
-            return result;
+            Module.print = origPrint;
+            Module.printErr = origPrintErr;
+
+            if (result !== 0) {
+                throw new Error('Loading VIA failed for the following reason: ' + niErrorEnum[result] +
+                    ' (error code: ' + result + ')' +
+                    ' (stdout: ' + printText + ')' +
```
spathiwa: What happens to the original output?   Will all the output and errout be printed twice, once during loading an then again when the exception is thrown, since the wrappers also forward to the original print functions?

source/io/module_eggShell.js
```diff
@@ -361,29 +386,133 @@
...
+                var startTime = performanceNow();
+
+                do {
+                    execSlicesResult = Module.eggShell.executeSlicesUntilWait(100000);
+                    elapsedTime = performanceNow() - startTime;
+
+                    if (execSlicesResult >= 0) {
+                        break;
+                    }
+                } while (elapsedTime < MAXIMUM_VIREO_EXECUTION_TIME_MS);
```
spathiwa: The C entry point for ExecuteSlices has a maximum execute time parameter which is not exposed to the JS wrapper.   Maybe instead of duplicating that functionality here, we should just expose the parameter to JS.

test-it/esh.js
```diff
@@ -2,54 +2,28 @@
...
-    (function executeVireo () {
-        var timeDelay = vireo.executeSlicesUntilWait(100000);
-        if (timeDelay > 0) {
-            setTimeout(executeVireo, timeDelay);
-        } else if (timeDelay < 0) {
-            setImmediate(executeVireo);
-        }
-    }());
+    var vireo = new Vireo();
+    vireo.httpClient.setXMLHttpRequestImplementation(xhr2);
```
spathiwa: Yay!

spathiwa: Approved with minor comments.

source/io/module_eggShell.js
rajsite: They will only be "printed" once by the stdio functions by passing to the print functions. I imagine longer term we may want errors reported by the runtime to go down a different path than stdio / stderr for C / JS embedders. It makes sense for a standalone command line tool but not so much for an embedded library. Even now it seems strange that vireo errors are printed to stdio and not stderr.

Since it is very possible stdio is ignored (ie no via from the editor performs stdio) capturing the stdio that occurs only on loadvia and including it in the exception is valuable context.

So to answer the question if you leave the default print functions and do not perform exception handling and let the exception propagate forward then you will likely see the messages twice. I think the better long term approach is to not put Vireo runtime messages on stderr and leave stderr as something user code can print to.

source/io/module_eggShell.js
rajsite: Oh yea forgot about that. And if I understand it right the current hardcoded value in ExecutionSlices is 20 ticks and from TickCount ticks are microseconds so it's effectively unused as is.

Yea I'll try exposing it and see if it behaves as expected. I'll leave the review open for now for other feedback since that will be a pretty small scope change on top of this.

source/io/module_eggShell.js
spathiwa: I agree, Vireo run-time errors should go to stderr.

make-it/EmMakefile
```diff
@@ -98,6 +98,7 @@ EM_WRAP = --pre-js $(PREJS) --post-js $(POSTJS)
 
 EM_EXPORTS = -s EXPORTED_FUNCTIONS="[\
     '_Vireo_Version', \
+	'_Vireo_MaxExecWakeUpTime', \
```
cglzaguilar: Fix space, it appears not aligned with the other lines on my view.

rajsite: 👍

source/io/module_eggShell.js
rajsite: Worked well! Exposed to the JS side and deleted unnecessary code.

spathiwa: Hurray!

source/core/ExecutionContext.cpp
```diff
@@ -312,12 +312,11 @@ InstructionCore* ExecutionContext::SuspendRunningQueueElt(InstructionCore* nextI
 // ExecuteSlices - execute instructions in run queue repeatedly (numSlices at a time before breaking out and checking
 // timers), until all clumps are finished or tickCount time is reached.
 // See enum ExecSlicesResult for explanation of return values, or comments below where result is set.
-Int32 /*ExecSlicesResult*/ ExecutionContext::ExecuteSlices(Int32 numSlices, PlatformTickType tickCount)
+Int32 /*ExecSlicesResult*/ ExecutionContext::ExecuteSlices(Int32 numSlices, Int32 millisecondsToRun)
```
sanmut: I would rather keep "PlatformTickType tickCount" as is. For some of the embedded devices it is foreseeable to call ExecuteSlices with time resolution smaller than milliseconds. I would rather fix the C EntryPoint EggShell_ExecuteSlices() to call tm->TheExecutionContext()->ExecuteSlices() after converting from the input millisecondsToRun to tickCount.
Also, currently, there are callsites of ExecutionContext::ExecuteSlices() call that assume the second argument is nanoseconds and not milliseconds.

spathiwa: I respectfully disagree; I am in favor of this change.   While I think it's fine to have a platform-specific tickcount internally, I don't think it's good to have it in a public API.  I think standardizing on milliseconds is cleaner, and I don't think there's a practical reason for a target to need to run for less than 1 ms on any real target Vireo is likely to support; Vireo is not a practical real-time target, and RT targets are the only ones where I would see such support as actually useful.

source/io/module_eggShell.js
```diff
@@ -361,21 +386,91 @@
...
-            return EggShell_ExecuteSlices(v_userShell, slices);
+        // executeSlicesUntilWait
+        // numSlices (optional): The minimum number of slice sets to run before checking if maxTimeMS has passed.
+        //    The larger the value the less overhead for execution and the quicker the diagram progresses
+        //    One slice set corresponds to 10 slices and at a minimum one slice set executes per invocation
+        // millisecondsToRun (optional): The amount of time in milliseconds vireo can execute slice sets before vireo saves state and returns
+        // return value (type ExecSlicesResult):
+        //     returns < 0 if should be called again ASAP, 0 if nothing to run, or positive value N if okay
+        //     to delay up to N milliseconds before calling again
+        Module.eggShell.executeSlicesUntilWait = publicAPI.eggShell.executeSlicesUntilWait = function (numSlices, millisecondsToRun) {
```
sanmut: If the "numSlices" and "millisecondsToRun" are optional, do we not need to check whether the values are valid and if not use pre-defined default values before calling into EggShell_ExecuteSlices()?

source/core/ExecutionContext.cpp
sanmut: hmm, I thought Paul was experimenting with microblaze and Vireo. Do you all know what the scope of that research was?

sanmut: @spathiwa  wrote ", I don't think it's good to have it in a public API. I think standardizing on milliseconds is cleaner," 
>> Yes, I am also ok with public API (in CEntryPoints.cpp) to have millisecond resolution as mentioned in my comment above. I am also saying let's keep internal as TickCount. I think we actually agree.

spathiwa: He was. I do not know the scope; perhaps if he is listening he can chime in.
But if there is a need to have a higher-resolution time period for this argument, I would favor adding a separate entry point, which *explicitly* takes nanoseconds rather than PlatformTickCount type.  And that's easy enough to do whenever we need it in the future, it shouldn't block this change.

source/io/module_eggShell.js
rajsite: We don't "need" to because asm.js does a  `<number> | 0` to coerce it to Int32 semantics in the js runtime.

You can see it in the generated output at the _EggShell_ExecuteSlices function in https://unpkg.com/vireo@7.0.0/dist/asmjs-unknown-emscripten/debug/vireo.js

reproduced below (this is an older version that took only two parameters while it now takes 3):
```
function _EggShell_ExecuteSlices($0, $1) {
 $0 = $0 | 0; //@line 11101
 $1 = $1 | 0; //@line 11102
 var $2 = 0, $3 = 0, $4 = 0, $5 = 0, $6 = 0, $7 = 0, $8 = 0, $9 = 0, label = 0, sp = 0; //@line 11103
 sp = STACKTOP; //@line 11104
 STACKTOP = STACKTOP + 16 | 0; //@line 11105
 // ...
}
```

And the value of the arguments if not provided is `undefined` and the value of `undefined | 0` is the numeric value `0`. The behavior of not passing an argument is subtly described in the comments as `at a minimum one slice set executes per invocation`.

I agree that it is a subtle behavior so I'm not opposed to putting the checks in. My only concern was that executeSlices is on a fast path and I want to reduce overhead on a fast path but it could definitely be a premature optimization. It is very likely run times will optimize the path anyway and the type checks may even assist that.

source/core/ExecutionContext.cpp
rajsite: One thing I want to keep consistent is that ExecuteSlices takes a time to run and returns a time to wait and those units and types are the same for both.

I also would not like to use PlatformTickType because both the unit and type vary across platforms which makes for a messy public api.

We should also keep one entry point that uses a type of Int32 because asm.js has very poor performance handling and manipulating Int64 types (won't be the case for wasm, but something we have to deal with now).

As long as we use a unit that can represent at least a couple of orders of magnitude above kMaxExecWakeUpTime and fits in an Int32, I'm good.

sanmut: Is my understanding right about "EggShell_ExecuteSlices defined in CEntryPoints.cpp is public API. ExecutionContext::ExecuteSlices defined in ExecutionContext.cpp is not a public API."? If so, I am not asking about changing EggShell_ExecuteSlices defined in CEntryPoints.cpp. If not, I would like to understand this better. With the current change, we are calling ExecutionContext::ExecuteSlices with 10 million milliseconds in one place -- which does not sound right.

cglzaguilar: I think we need to have this discussion on person to better understand the consequences, meanwhile Milan is blocked, and I would like Milan to continue, so I asked him to submit his changes. We can revisit later and adjust as necessary.

rajsite: Talked to @sanmut and agreed:
* the time unit in and out should be consistent so we won't revert the change
* longer term the main.cpp should be consuming the CEntryPoints that other embedders would use as well
* We can add entry points to the public api in the future for higher resolution timing and change the internal representation as needed

PR #357, opened 2017-12-20
Fix to handle data values for optional arguments.

sanmut:
The existing code in ClumpParseState::ResolveActualArgument created a null
pointer for optional arguments for flat types but created a new value for
non-flat types. This was incorrect because the value for this optional
argument (though not explicit in the VIA code) is reused for subsequent
calls. (Almost) All the primitives explicitly handle correctly whether an
optional parameter is null by checking the pointer and using the correct
default value if null regardless of whether the type of the argument is
flat or not. Thus this method is fixed to treat flat and non-flat types
same and not allocate data value for optional arguments.

Add a test (WildcardArguments.via)

Some primitives such as StringFormat, FormatDateTimeString are the
exceptions that did not handle the optional arguments and they are fixed
to do so.

A memory overrun in StringScan is fixed by using AliasAssignCStrLen.

Fix InlineConstants.via to not use first argument as optional argument any
more as it is not anymore.

### DIFF
--- a/source/core/VirtualInstrument.cpp
+++ b/source/core/VirtualInstrument.cpp
@@ -531,18 +531,8 @@ void ClumpParseState::ResolveActualArgument(SubString* argument, void** ppData, 
     // See if actual argument is '*' meaning use the default/unwired behavior.
     if (argument->CompareCStr(tsWildCard)) {
         _actualArgumentType = FormalParameterType();
-        if (!_actualArgumentType->IsFlat()) {
-            // Define a DefaultValue type. As a DV it will never merge to another instance.
-            // Since it has no name it cannot be looked up, but it will be freed once the TADM/ExecutionContext is freed up.
-            DefaultValueType *cdt = DefaultValueType::New(_clump->TheTypeManager(), _actualArgumentType, false);
-            cdt = cdt->FinalizeDVT();
-            if (needsAddress) {
-                *ppData = (AQBlock1*)cdt->Begin(kPARead);  // * passed as a param means null
-            }
-        } else {
-            // For flat data, the call instruction logic for VIs will initialize the callee parameter
-            // to the default value. For native instructions a null parameter value is passed.
-        }
+        // Regardless of whether the optional argument is flat or not, do not allocate data value
+        // and use null-pointer and let the prims handle the optional arguments correctly
         _argumentState = kArgumentResolvedToDefault;
         return;
     }
--- a/test-it/ViaTests/InlineConstants.via
+++ b/test-it/ViaTests/InlineConstants.via
@@ -21,7 +21,7 @@ define(HelloWorld dv(.VirtualInstrument (
         it need to be escaped. How.
 */
 
-        StringToUpper(* str)
+        StringToUpper(str str)
         Printf("passing default value <%s>\n" str)
         StringToUpper("Hello, World" str)
         Printf("passing inline double quoted constant value <%s>\n" str)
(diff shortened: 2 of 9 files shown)


test-it/ViaTests/WildcardArguments.via
```diff
@@ -0,0 +1,54 @@
...
+           Printf("String=[%s] Scanned Value=[%d]\n" g0 local_1)
+           Printf("String=[%s] Scanned Value=[%d]\n" g1 local_2)
+
+/* Clump Ended. */    )
+)))
+
+//---------------------------------------------------
+define (SubVI1 dv(.VirtualInstrument (
+    Params: c(  //param block
+        i(dv(.String '')dataItem_InputString)
```
jattasNI: Can you make sure we have tests for both flat and non-flat data where the default value is different than the "default default"?

jattasNI:
I don't spend much time in the VireoSDK codebase. Is it normal for the CIs to fail?

sanmut:
Not generally. This particular failure is not an intermittent issue but it is just broke now. Fix is on the way. It was my bad to make a trivial change and not run the test suite locally.

test-it/ViaTests/WildcardArguments.via
sanmut: Done. Added a new test case for flat datatype too.

source/core/VirtualInstrument.cpp
```diff
@@ -531,18 +531,8 @@ void ClumpParseState::ResolveActualArgument(SubString* argument, void** ppData,
...
-        if (!_actualArgumentType->IsFlat()) {
-            // Define a DefaultValue type. As a DV it will never merge to another instance.
-            // Since it has no name it cannot be looked up, but it will be freed once the TADM/ExecutionContext is freed up.
-            DefaultValueType *cdt = DefaultValueType::New(_clump->TheTypeManager(), _actualArgumentType, false);
-            cdt = cdt->FinalizeDVT();
-            if (needsAddress) {
-                *ppData = (AQBlock1*)cdt->Begin(kPARead);  // * passed as a param means null
-            }
-        } else {
-            // For flat data, the call instruction logic for VIs will initialize the callee parameter
```
rajsite: So if I understand (which I may not), this change is making it so that for **native instructions** (ie primitives) when a * parameter is passed that both Flat and Non-Flat representations result in a null parameter value.

Based on this comment it sounds like for **VI calls** for both Flat and Non-Flat data a DefaultValue is created for * parameters. So it sounds like for VI calls we should always be expecting non-null parameters for both Flat and Non-Flat when * is used.

If that interpretation is correct then do we have a test that is verifying that behavior for VI calls? Just based on the comment it sounds like we would now only be creating default values for Flat types when a VI is invoked with * as the logic for Non-Flat types was just removed.

test-it/ViaTests/InlineConstants.via
```diff
@@ -21,7 +21,7 @@ define(HelloWorld dv(.VirtualInstrument (
         it need to be escaped. How.
 */
 
-        StringToUpper(* str)
+        StringToUpper(str str)
```
rajsite: Are we modifying this because we are choosing to not update StringToUpper to handle null values at the moment?

rajsite: Sounds like this concern will be addressed as part of https://github.com/ni/VireoSDK/issues/358

source/core/VirtualInstrument.cpp
sanmut: I have added a test to exercise different SubVI call behavior with parameter types. With and without this fix, the behavior is the same. Discussed with Milan and the comments might be outdated. Did find a couple of non-showstopper issues and filed them:
https://github.com/ni/VireoSDK/issues/360
https://github.com/ni/VireoSDK/issues/359
Milan is fine to submit this now.

sanmut: If @cglzaguilar has feedback, we will address it later in a separate PR later.

PR #368, opened 2018-01-17
CAR 682384: Fix to handling parse errors within ParseArrayData

segaljared:
Fix for CAR 682384: WebVI hang executing Unflatten from JSON

ParseArrayData doesn't break out of its own parse loop when there is an error parsing elements. This can result in an infinite loop within the array's parse loop. If the parse of the element has an error, then we'll now break the loop and bubble the error up so it is returned by the ParseArrayData call.


source/include/TDCodecVia.h
```diff
@@ -35,7 +35,8 @@ enum ViaFormat {
     kViaFormat_UseLongNameInfNaN = 8,  // mask,  clear == use inf,nan, set == use Infinity/NaN
     kViaFormat_SuppressInfNaN = 16,    // use neither,
     kViaFormat_JSONStrictValidation = 32,
-    kViaFormat_QuoteInfNanNames = 64
+    kViaFormat_QuoteInfNanNames = 64,
```
sanmut: We should try to use hex format since it is a bit mask for clarity.

source/core/TDCodecVia.cpp
```diff
@@ -936,8 +936,11 @@ Int32 TDViaParser::ParseArrayData(TypedArrayCoreRef pArray, void* pFirstEltInSli
                     // For nested dimensions just parse the next inner dimension using the array type.
                     subErr = ParseArrayData(pArray, pElement, level + 1);
                 }
-                if (subErr && !errCode)
+                if (subErr && !errCode) {
                     errCode = subErr;
+                    if (Fmt().StopArrayParseOnFirstError())
```
cglzaguilar: I think we need to break out in general and not just for the JSON scenario. My concern is that the hang may still be there with bad VIA array code and we should fix that case too. I'm fine with this fix giving that we don't have much time to exercise it. Jared, please open an issue in Github to revisit this fix later on.

PR #370, opened 2018-01-19
Fixing ArrayDelete issue

segaljared:
ArrayDelete crashes when called on certain 2D arrays. This occurs when the length passed into the instruction is longer than the dimension along which you are deleting. For example, deleting 3 "elements" (rows in this case) from a 3 x 2 array starting at row index 1. Currently, ArrayDelete will attempt to create an output array of size 3 x -1, which causes a crash when we call arrayOut->ResizeDimensions.

The fix is to account for the lengths of each dimension when computing endIndex and deletedPortionLength.


source/core/Array.cpp
```diff
@@ -1211,8 +1211,11 @@ VIREO_FUNCTION_SIGNATURE7(ArrayDeleteND, TypedArrayCoreRef, StaticType, void,
 
     IntIndex startIndex = offset > 0 ? offset : 0;
```
sanmut: Looking at the CG prim documentation, it looks like we may have to compare the absolute value of offset (if offset is negative) against the length and make some decisions.  Specifically, the following two scenarios:

1) If you wire a value to length and a negative value to the index, deleted portion is an array with an outer dimension size of the value of length minus the absolute value of the index. 
2) If the index is a negative value and the value of length is smaller than the absolute value of the index, deleted portion is empty.
I would recommend writing test cases for these in the VI test and ensure Vireo produces similar results as Desktop. This may be a corner case but it is likely easier to fix it now than come back to this later.

source/core/Array.cpp
```diff
@@ -1310,7 +1313,9 @@ VIREO_FUNCTION_SIGNATURE6(ArrayDelete, TypedArrayCoreRef, StaticType, void, Type
     TypedArrayCoreRef arrayIn = _Param(3);
     TypeRef deletedPartType = _ParamPointer(1);
     IntIndex length = (_ParamPointer(4) == NULL)? 1 : _Param(4);
-    IntIndex offset = (_ParamPointer(5) == NULL)? arrayIn->Length() - 1 : _Param(5);;
+    IntIndex offset = (_ParamPointer(5) == NULL)? arrayIn->Length() - length : _Param(5);
+    if (offset > arrayIn->Length())
+        offset = arrayIn->Length();
 
     IntIndex startIndex = offset > 0? offset : 0;
```
sanmut: Same as the comment mentioned against ArrayDeleteND.

source/core/Array.cpp
sanmut: I suggest just emitting ArrayDeleteND version from DFIR for both 1D and multi-dim arrays and run the CI test suite and see if everything passes. If so, we may be able to deprecate ArrayDelete and keep only ArrayDeleteND. Again, this might be easier to do now while we are at it -- if it already works. If we're out of time, we can revisit this later.

source/core/Array.cpp
```diff
@@ -1211,8 +1211,11 @@ VIREO_FUNCTION_SIGNATURE7(ArrayDeleteND, TypedArrayCoreRef, StaticType, void,
 
     IntIndex startIndex = offset > 0 ? offset : 0;
     IntIndex endIndex = offset +
-        deletedPortionLength > arrayIn->GetLength(numberOfDimensions-1-dimensionToDelete) ?
+        deletedPortionLength > arrayIn->GetLength(numberOfDimensions-1-dimensionToDelete) - startIndex ?
         arrayIn->GetLength(numberOfDimensions-1-dimensionToDelete) : offset + deletedPortionLength;
```
cglzaguilar: It appears the code is adding offset twice in the false clause of the conditional operator (?). I think that section is incorrect and it should just use "deletedPortionLength" and not "offset + deletedPortionLenght". I wonder if we have tests that exercise this scenario.

cglzaguilar: Also shouldn't "arrayIn->GetLength(numberOfDimensions-1-dimensionToDelete)" in the true clause of the conditional operator(?) also use - startIndex? Otherwise endIndex would be ending beyond the desired limits.

source/core/Array.cpp
segaljared: It actually already works like that. In Desktop if you wired in -1 as the index and a 2 as the length, then it will delete the first element of the array (i.e. it sets the startIndex to 0 and the endIndex to offset {-1} + length {2 = 1).  This works both in ArrayDeleteND and ArrayDelete.

segaljared: It looks like I can do this with just a few modifications. Should I remove ArrayDelete or just mark it as deprecated?

sanmut: oh ok, thanks. If we have test cases that cover these two scenarios, then we are good.

sanmut: I would leave it for now. Just mark it with [unused] in the comments just above the function for now. We should create a strategy for deprecation. There may be some APIs that we may want to leave in VIA but just identify them as not being used by LabVIEW and there may be some that we may want to deprecate/delete. We don't need to finalize how we want to do all that now.

source/core/Array.cpp
segaljared: The code could probably be changed to make this more clear, but it is looking at whether offset + deletedPortionLength is greater than array length - startIndex. It doesn't add offset to whatever is returned from the conditional. endIndex gets assigned to what is returned. I could add parentheses to make it more clear, though.

Also, we want endIndex to equal the length of the array in the true case, which is why we don't subtract startIndex.

cglzaguilar: Ok, we should make the code to be clearer. We can do this later.

PR #372, opened 2018-02-01
Initial submission to enable Vireo to call user-defined JavaScript functions

cglzaguilar:
- Add base files
- Add new entry point
- Add js files to invoke user-defined js function
- Add basic set of tests to exercise the feature
- Fixed Vireo module load order, by adding dependencies.

### DIFF
--- a/source/io/JavaScriptInvoke.cpp
+++ b/source/io/JavaScriptInvoke.cpp
@@ -1,0 +1,123 @@
+/**
+
+Copyright (c) 2018 National Instruments Corp.
+
+This software is subject to the terms described in the LICENSE.TXT file
+
+SDG
+*/
+
+/*! \file
+\brief Functions to call user defined JavaScript functions
+*/
+
+#include "TypeDefiner.h"
+#include "ExecutionContext.h"
+#include "StringUtilities.h"
+#include "TDCodecVia.h"
+#include <stdio.h>
+
+#if defined (VIREO_TYPE_JavaScriptInvoke)
+namespace Vireo {
+
+#if kVireoOS_emscripten
+extern "C" {
+    // JavaScript function prototypes
+    // Parameters: functionName, parameters*, parametersCount, errorCheckingEnabled, errorStatus*, errorCode*, errorSource*
+    extern void jsJavaScriptInvoke(StringRef, void *, Int32, Boolean, Boolean *, Int32 *, StringRef);
+}
+#endif
+
+// Keep in synchrony with JAVASCRIPT_PARAMETER_TYPE in module_javaScriptInvoke.js
+enum JavaScriptParameterTypeEnum {
+    kJavaScriptParameterType_None = 0,
+    kJavaScriptParameterType_Int32,
+    kJavaScriptParameterType_UInt32,
+    kJavaScriptParameterType_String,
+    kJavaScriptParameterType_Boolean
+};
+
+//------------------------------------------------------------
+//! Return parameter type for the given element(index) in the parameters array
+VIREO_EXPORT Int32 Data_GetParameterType(StaticTypeAndData *parameters, Int32 index)
+{
+    TypeRef parameterType = parameters[index]._paramType;
+    Int32 type = kJavaScriptParameterType_None;
+    if (parameterType->BitEncoding() == kEncoding_UInt) {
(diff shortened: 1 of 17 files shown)


source/io/JavaScriptInvoke.cpp
```diff
@@ -0,0 +1,123 @@
...
+{
+    _ParamDef(StringRef, functionName);
+    _ParamDef(Boolean, errorCheckingEnabled);
+    _ParamDef(ErrorCluster, errorCluster);
+    _ParamImmediateDef(StaticTypeAndData, Parameters[1]);
+    NEXT_INSTRUCTION_METHODV()
+};
+
+extern void AddCallChainToSourceIfErrorPresent(ErrorCluster *errorCluster, ConstCStr methodName);
+extern void GenerateNotSupportedOnPlatformError(ErrorCluster *errorCluster, ConstCStr methodName);
```
spathiwa: Probably should move these to a header file

cglzaguilar: Good idea. I'll do that in a later submission.
Regarding the VIREO_REQUIRE blocks. I did that because adding a new Vireo module, changed the order of how Vireo modules are loaded and I have to add require dependencies so they load in a determined other.

cglzaguilar:
Thanks a lot for your feedback! Since this is going to a branch. I'll address your comments in the next round of changes.

source/io/module_javaScriptInvoke.js
```diff
@@ -0,0 +1,157 @@
...
+            MESSAGE: 'Unsupported parameter type in JavaScript invoke.'
+        },
+
+        kNIUnableToFindFunctionForJavaScriptInvoke: {
+            CODE: 363658, // TO_DO (CGA): We need to generate a new error code
+            MESSAGE: 'Unable to find JavaScript function'
+        }
+    };
+
+    // Keep in synchrony with JavaScriptParameterTypeEnum in JavaScriptInvoke.cpp
```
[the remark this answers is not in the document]
rajsite: +1 I did that in EggShell as well. You can see the eslint exception for that case: https://github.com/ni/VireoSDK/blob/master/source/io/module_eggShell.js#L30

While I really don't want lots of eslint exceptions around JS code, being able to keep consistent names between JS and CPP for shared things is helpful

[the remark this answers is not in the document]
rajsite: If you follow that pattern make sure to do the same thing and put a reason why you are adding the exception (if there isnt a good reason it ain't worth doing)

source/io/JavaScriptInvoke.cpp
```diff
@@ -0,0 +1,123 @@
...
+    kJavaScriptParameterType_None = 0,
+    kJavaScriptParameterType_Int32,
+    kJavaScriptParameterType_UInt32,
+    kJavaScriptParameterType_String,
+    kJavaScriptParameterType_Boolean
+};
+
+//------------------------------------------------------------
+//! Return parameter type for the given element(index) in the parameters array
+VIREO_EXPORT Int32 Data_GetParameterType(StaticTypeAndData *parameters, Int32 index)
```
rajsite: The Data_* exports are meant to be fairly generic for the C API of talking to Vireo and are in EggShell.
These functions look very specific to JSLI. I think they should not be called Data_*.

Maybe something to emphasize they are JSLI specific like JavaScriptInvoke_GetParameterType or JavaScriptInvoke_GetParameterPointer.

source/io/module_javaScriptInvoke.js
```diff
@@ -0,0 +1,157 @@
...
+    var JAVASCRIPT_PARAMETER_TYPE = Object.freeze({
```
rajsite: We talked about it briefly but I'm interested in trying to leverage Vireo's existing type representation for specifying these types.

Vireo already knows how to turn its types into strings, maybe we make some JS code to interpret Vireo's representation of types instead of creating a new way to represent them?

If we develop that code it would be great to make it so that the NIType library can read and serialize Vireo's type representation: https://github.com/ni-kismet/data-types

Tagging @gleonoliva in-case this topic is interesting.

rajsite: Also @prestwick maybe we need to look at open sourcing the NIType library as well

PR #373, opened 2018-02-13
Added support for more data types and return values

cglzaguilar:
- Added support for more data types
- Added support for return values
- Added tests to exercise error conditions and normal working scenarios


cglzaguilar:
This commit is not going to master branch. .Adding commit to branch to enable further testing.

PR #374, opened 2018-02-19
Enable Vireo to call user defined JavaScript functions

cglzaguilar:
- Added support for scalar types
- Added support for return types
- Added tests for normal and error paths


test-it/karma/fixtures/javascriptinvoke/DataTypes.via
```diff
@@ -0,0 +1,45 @@
...
+            e(.Boolean status)
+            e(.Int32 code)
+            e(.String source)
+        ) error)
+        e(c(
+            e(.Boolean status)
+            e(.Int32 code)
+            e(.String source)
+        ) error2)
+        e(dv(.Boolean t) booleanTrueValue)
```
sanmut: I think we should go with "true" and "false" for boolean values.

make-it/EmMakefile
```diff
@@ -113,11 +114,22 @@ EM_EXPORTS = -s EXPORTED_FUNCTIONS="[\
     '_Occurrence_Set',\
     '_Data_GetStringBegin',\
     '_Data_GetStringLength',\
-    '_Data_WriteString',\
     '_Data_ReadBoolean',\
+    '_Data_ReadUInt8',\
+    '_Data_ReadUInt16',\
+    '_Data_ReadUInt32',\
```
sanmut: If we are writing the "Data_ReadUInt*" versions, we should write "Data_ReadInt*" versions too (especially as we have the Data_Write* versions for both signed and Unsigned).

PR #375, opened 2018-03-06
Refactor RefNumVal to separate maxSize from queue refnum's, also moved out of TypeArray so it can be dynamic without de-uniqueing types.

spathiwa:


cglzaguilar:
Sorry, this is taking longer. But we are looking into a memory leak fix that will require a new Vireo release.

PR #376, opened 2018-03-09
Fix memory leak due to passing wildcard arguments in SubVI calls

sanmut:


sanmut:
@spathiwa , I agree. We should print the test name during JS exceptions or kill dots (Carlos and I were exactly talking about this a minute ago). However, I am not able to reproduce this failure even with Debug VireoJS locally. So, this is going to be tricky to debug.

spathiwa:
FYI, I can crash using only:

./test.js  ViaTests/WildcardArguments.via ViaTests/WildCardArgumentsSubVICalls.via

Without running all the other tests.

Craig

spathiwa:
I have a potential fix, but it requires disallowing something that is in the Vireo test suite (but which I think shouldn't be supported and I don't think we would ever generate from DFIR):  explicitly copying to inputs in a subvi.  This assert is what is causing the appveyor failure.  Changing this case to use CopyObject instead of CopyTop (or disallowing) avoids the crash but makes it so the subvi is actually able to modify its input so it's visible to the caller (clearly bad).

diff --git a/source/core/GenericFunctions.cpp b/source/core/GenericFunctions.cpp
index 6298337f..dc8629e7 100644
--- a/source/core/GenericFunctions.cpp
+++ b/source/core/GenericFunctions.cpp
@@ -90,7 +90,10 @@ InstructionCore* EmitGenericCopyInstruction(ClumpParseState* pInstructionBuilder
                 }
             }
         } else if (sourceType->IsArray()) {
-            VIREO_ASSERT(!destType->IsInputParam());
+            //VIREO_ASSERT(!destType->IsInputParam());
+            if (destType->IsInputParam() && !destType->IsOutputParam()) {
+                return null;  // disallow copy to an input-only arg
+            }
             if (destType->IsAlias() && !destType->IsOutputParam()) {
                 // If its a local alias then just copy the pointer.
                 // Outputs are alias to the callers params, so that still needs a deep copy

And then commenting out/removing line 48 of ViaTests/WildCardArgumentsSubVICalls.via

sanmut:
Craig has a PR on top of this fix that addresses an existing data corruption crash that this PR made it happen intermittently on CI.

PR #377, opened 2018-03-12
Fix memory leak due to passing wildcard arguments in SubVI calls

spathiwa:
Disallow writing to inputs in subVIs, which (because inputs are either
aliases to caller's data or allocated in parm block for wildcard args)
can overwrite the only reference to a wild/default'd allocation and
cause a leak.

When passing multiple args to the native esh command line program, run
them in the same TypeManager shell to stress test and help find leaks.

Turn on VIREO_ASSERT for native builds, even in release target, to better find
issues on the CI builds.


source/CommandLine/main.cpp
```diff
@@ -45,44 +45,44 @@ int VIREO_MAIN(int argc, const char * argv[])
     }
 
     if (argc >= 2) {
+        gShells._pRootShell = TypeManager::New(null);
```
sanmut: I see your point about the change in this file (main.cpp). However, I tend to think this need not be done as part of this PR. We should create a Vireo release without this change and with only the other fixes. This may be done separately as a commit in a different PR and with adding some tests with the workflow of passing more than one VIA file as input.

source/core/VirtualInstrument.cpp
```diff
@@ -564,7 +564,10 @@ void ClumpParseState::ResolveActualArgument(SubString* argument, void** ppData,
     // See if it is in the VI's locals or paramblock
     _actualArgumentType = _vi->GetVIElementAddressFromPath(argument, _vi, (void**)ppData, false);
     if (_actualArgumentType) {
-        _argumentState = kArgumentResolvedToVIElement;
+        if (_actualArgumentType->ElementUsageType() == kUsageTypeInput && _formalParameterType->ElementUsageType() != kUsageTypeInput)
```
sanmut: This is a good idea. I am not sure about kUsageTypeAlias situations but not likely used much in ViaTests and not likely used from DFIR at all.

source/CommandLine/main.cpp
spathiwa: This to me is like submitting test via files together with the change that fixes them.
With this change, it's possible to reproduce buggy behavior associated with this fix using the native driver by running WildCardArgumentSubVICalls.vi twice, and without it, it is not.
I don't think it's important whether or not it is rolled into this Vireo release because main.cpp doesn't affect the JS build at all.   So I'm pushing back on this, but if you feel strongly, I can pull it out.

source/core/VirtualInstrument.cpp
spathiwa: I don't think kUsageTypeAlias can show up in this context because _actualArgumentType has to be an input for the failure case, and inputs only show up in the parameter block.  Also, the SharedArray test does some testing of aliases and still works

source/include/BuildConfig.h
```diff
@@ -244,6 +232,22 @@ SDG
...
 #endif
 
+// TODO(PaulAustin): allow for thread locals on linux/unix
+#define VIVM_THREAD_LOCAL
+
+#ifdef VIREO_DEBUG
+#define VIREO_TRACK_MEMORY_QUANTITY
+#define VIREO_USING_ASSERTS
+#else
+#if (kVireoOS_macosxU || kVireoOS_linuxU || kVireoOS_windows)
```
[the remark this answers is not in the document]
spathiwa: I did this because I want local builds by default to also have asserts, and it's also nice if the CI build has asserts but is still optimized -- it has no use for debugging info, and runs slower with a debug build, and we really do want to test the optimizer in at least one of the CIs since that frequently finds bugs that don't show up in debug builds, yet the asserts still catch real issues.  

If you'd like, I could add a separate flag besides DEBUG that lets you turn on asserts independently and use that instead (as well as making it the default for local builds).  (Btw, the LLVM build a Release+Asserts build option, so there's precedent. :-) )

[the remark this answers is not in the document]
spathiwa: Ok, I've made the enabling of asserts in a non-debug build explicit by pre-defining VIREO_USING_ASSERTS in the Makefile (instead of in code when DEBUG is off)  (commit 1deb65).

source/CommandLine/main.cpp
sanmut: I did a quick sanity test on this locally now. With this change, it looks like if two VIA files are passed as input to Vireo native, output from the first VIA file goes to the console twice and output from second VIA file does not go to the console. If we can fix this (and optionally/preferably add a small test on this console behavior), let's check this in.

We are using Vireo Native in the NXG side of CI tests, so we need to make sure that won't be broke with Vireo updates.

source/core/VirtualInstrument.cpp
sanmut: Got it. thanks.

source/CommandLine/main.cpp
spathiwa: Ok, please try again, I think I fixed it.   The user scope still needs to be instantiated per via file or the global namespace might have conflicts.

spathiwa:
@cglzaguilar Are you okay with going ahead and merging this?

cglzaguilar:
I'm ok, with the changes approving this one.

PR #379, opened 2018-03-13
Spit STDERR also to console from test runner.

sanmut:
Spit STDERR to console from test runner to force test failures due to
exceptions and asserts.
Turn on ASSEERTS in Visual Studio Debug builds


sanmut:
**Note to the reviewers:** 
You will see a lot of test failures from CIs. You can ignore them for now. I expect all of the failures to pass after Craig's outstanding PR gets merged. Just review the changes. I will merge this if no failures after Craig's change goes in.

PR #381, opened 2018-03-19
Created module and entry points for property node support

cglzaguilar:


cglzaguilar:
Initial work for property node entry points that is going into a branch and not in master to allow for more people to work on it. Not ready for review, once tests pass, I'll rebase and merge into the PropertyNode branch.

PR #382, opened 2018-03-21
Removed array rank assertion

cglzaguilar:
After we turned on assertions, this one was causing failures on LVjs CI.

The assertion is no longer valid since we now support n-dimensional arrays.

Added one of the test failures here, since we weren't exercising that code path.


cglzaguilar:
Already reviewed by sanmut and spathiwa. Once the tests pass I'll rebase and merge

PR #383, opened 2018-03-21
Fix bug in LossyEnqueue caused by queue refnum maxsize refactor.

spathiwa:
Problem occurs when you use lossyEnqueue before enqueue on a fixed size
queue.


sanmut:
Awesome, Thanks @cglzaguilar

PR #387, opened 2018-03-27
Created module and entry points for property node support

cglzaguilar:


cglzaguilar:
Just created a new branch with user defined event work and adding current property node work on top

sanmut:
It looks like this branch is not up to date with master. For branching and merging to work effectively, we should keep the branch up-to-date with master as often as possible. We should rebase local changes in the branch on top of the latest commits from master. That way, merging from branch onto master would be painless.

We should avoid two feature branches dependent on each other and diverging from master. That's a situation that can get more complicated. We don't seem to run into that yet.

cglzaguilar:
I already talked with Sankara about his concerns. We will only have one branch and delete the other one. The branch that we will keep is up to date with master.

PR #388, opened 2018-03-27
Property node JS read/write functions

gleonoliva:
In this pull request we add to vireo js the capability of setting propertyRead and propertyWrite functions. This follows the same pattern as setFPsyncFunction and JavaScriptInvoke.

Right now callback functions receive as parameters: viName, controlId, propertyName, propertyType, propertyPath. We are leveraging the work of reading/writing from/to vireo's memory to these callback functions. This means that vireo is only responsible of passing the right information so that they can do their job.

Implementation will change as soon as we get control references working.


source/io/module_propertyNode.js
```diff
@@ -30,48 +30,48 @@
...
+        kNIObjectReferenceIsInvalid: {
+            CODE: 1055,
+            MESSAGE: 'Object reference is invalid.'
         }
     };
 
-    /* var formatMessageWithException = function (messageText, exception) {
+    var formatMessageWithException = function (messageText, exception) {
         if (typeof exception.message === 'string' && exception.message.length !== 0) {
             return messageText + ', Additional information: ' + exception.message;
```
rajsite: I think this is supposed to be `<APPEND>\n` instead of `Additional Information` see the following: https://github.com/ni/VireoSDK/blob/master/source/io/module_httpClient.js#L112

source/io/PropertyNode.cpp
```diff
@@ -15,18 +15,22 @@ SDG
 #include "ExecutionContext.h"
 #include "StringUtilities.h"
 #include "TDCodecVia.h"
+#include "VirtualInstrument.h"
 #include <stdio.h>
 
+
+#define VIREO_TYPE_PropertyNode 1
```
spathiwa: Suggest renaming this; this name just doesn't 'sound' like an token/enabling macro.

source/io/PropertyNode.cpp
```diff
@@ -94,12 +111,25 @@ VIREO_FUNCTION_SIGNATUREV(PropertyNodeRead, PropertyNodeReadParamBlock)
     StringRef propertyName = _Param(propertyName);
     StaticTypeAndData *value = _ParamImmediate(value);
 
+    ExecutionContextRef exec = THREAD_EXEC();
+    VIClump* runningQueueElt = exec->_runningQueueElt;
+    VirtualInstrument* vi = runningQueueElt->OwningVI();
+    TypeManagerRef typeManager = vi->TheTypeManager();
```
spathiwa: Instead of all this, you can get the TypeManager out of the StaticTypeAndData parm you already have.
value->_paramType->TheTypeManager()

source/io/PropertyNode.cpp
```diff
@@ -15,18 +15,22 @@ SDG
...
 namespace Vireo {
 
 #if kVireoOS_emscripten
 extern "C" {
     // JavaScript function prototypes
-    // Parameters: viName, controlId, propertyName, value, errorStatus*, errorCode*, errorSource*
-    extern void jsPropertyNodeWrite(StringRef, StringRef, StringRef, void *, Boolean *, Int32 *, StringRef);
-    // Parameters: viName, controlId, propertyName, value, errorStatus*, errorCode*, errorSource*
-    extern void jsPropertyNodeRead(StringRef, StringRef, StringRef, void *, Boolean *, Int32 *, StringRef);
+    // Parameters: viName, controlId, propertyName, propertyType, propertyPath, errorStatus*, errorCode*, errorSource*
```
sanmut: I thought controlId would be resolved at JS layer from the path (viName and dataItem name). Is that changing?

rajsite: shouldn't be changing, dataItemId might be a better name

source/io/module_propertyNode.js
gleonoliva: I found this in module_JavaScriptInvoke.js should I modify that one too?

source/io/PropertyNode.cpp
gleonoliva: I wasn't sure if that was a recommended way to get. Great to know it is. I'll change it.

source/io/PropertyNode.cpp
gleonoliva: I guess I should change the name to be dataItemName instead.

source/io/PropertyNode.cpp
gleonoliva: I saw this convention being used in a couple of other files. What is the recommended style?

spathiwa: That form, with VIREO_TYPE_ prefix, is supposed to only be used for guarding inclusion of support of Vireo primitive data types, not other features or primitives.
However, I notice that JSLI, HttpClient, and several others have already used this form as a guard for features instead of types, so I retract my objection.  (Maybe this should be a tech debt item to clean this up.)

source/io/PropertyNode.cpp
gleonoliva: As it is now, without control references, we pass the viName and controlId (this could be the dataItemName). Once we have control references we would pass a control reference where this information would be encoded. i.e. we are expeting this to change.

gleonoliva: dataItemId will be

source/io/PropertyNode.cpp
cglzaguilar: Perhaps we can change this one now and do the others later. Use VIREO_MODULE_PropertyNode, instead. The idea is that you can compile modules out as required by a specific target. The idea was to implement HTTP and JSLI as Vireo Modules.

test-it/karma/fixtures/propertynode/PropertyNodeRead.via
```diff
@@ -0,0 +1,27 @@
...
+            PropertyNodeRead("MyVI" "123" "value" booleanLocal error)
+            PropertyNodeRead("MyVI" "123" "value" stringLocal error)
+            PropertyNodeRead("MyVI" "123" "value" doubleLocal error)
+            PropertyNodeRead("MyVI" "123" "value" int32Local error)
+            PropertyNodeRead("MyVI" "123" "value" uint32Local error)
+            PropertyNodeRead("MyVI" "123" "value" complexDoubleLocal error)
+            PropertyNodeRead("MyVI" "123" "value" timestampLocal error)
+        )
+)))
+enqueue(MyVI)
```
cglzaguilar: Add a new line at the end of the file to remove the red thing.

cglzaguilar: Update 123 to something that looks like a "DataItem" is this and other tests.

test-it/karma/fixtures/propertynode/PropertyNodeWrite.via
```diff
@@ -0,0 +1,27 @@
...
+            PropertyNodeWrite("MyVI" "123" "value" myBoolValue error)
+            PropertyNodeWrite("MyVI" "123" "value" stringLocal error)
+            PropertyNodeWrite("MyVI" "123" "value" doubleLocal error)
+            PropertyNodeWrite("MyVI" "123" "value" int32Local error)
+            PropertyNodeWrite("MyVI" "123" "value" uint32Local error)
+            PropertyNodeWrite("MyVI" "123" "value" complexDoubleLocal error)
+            PropertyNodeWrite("MyVI" "123" "value" timestampLocal error)
+        )
+)))
+enqueue(MyVI)
```
cglzaguilar: Add a new line at the end of the file to remove the red thing.

PR #389, opened 2018-03-27
Chrome headless support

rajsite:
Adds Headless Chrome as a dependency
Switches to Headless Firefox

Runs emsdk_env commands as separate shell instances to avoid polluting build environments (was replacing node in PATH with an older and incompatible version)

Moves HTTP tests to their correct location in the http folder so tests are organized by feature and not by how the test behaves (fails phantom or runs slow). Instead relying on tags to filter tests based on different properties.

Updated the scripts in package.json to reflect the naming convention used in TFS (ie npm run karma -> npm run test). 

**We can squash merge this commit, has lots of test build noise**

Edit: Disabled IE11 HTTP tests to avoid intermittency in AppVeyor


.travis.yml
```diff
@@ -27,28 +26,27 @@ install:
...
 - while ! nc -q 1 127.0.0.1 64526 </dev/null; do sleep 2; done
 - curl http://127.0.0.1:64526/get?show_env=1
 - make lint
 - npm run lint
-- make
-- make vjs BUILD=debug
-- make vjs BUILD=profile
+- make native
 - make unittest
+- bash -c 'source emscripten-sdk/emsdk_env.sh && make vjs'
```
spathiwa: Why is this necessary?  Was emsdk being directly in the environment interfering with other commands?

rajsite:
@spathiwa I was thinking of squash merging this as one commit as mentioned in the comment at the top. The only substantive changes are to the appveyor and travis yml. The rest are moving tests around.

I personally think looking at the results as one commit is not that confusing: https://github.com/ni/VireoSDK/pull/389/files

If you think 1 commit is not enough then I'd propose two commits maybe where the tests are a separate commit. Let me know if that is worth it or if you want a different partition.

.travis.yml
rajsite: yep, nailed it
> Runs emsdk_env commands as separate shell instances to avoid polluting build environments (was replacing node in PATH with an older and incompatible version)

spathiwa:
The problem with using the Files view is that you lose the descriptions explaining the purpose of the change. 
I don't have a problem with squash-merging this commit.

I just think squashing stuff into a handful of commits **before** you request the PR would be helpful to reviewers when there are lots of files or there is any question about what changes go with which subpart of the PR's overall description.

I was saying this just as a general thing, not that this PR specifically is that hard to review, but it reminded me of suggesting this, which I've thought of before in previous reviews with lots of noisy commits.

rajsite:
Not trying to be defensive, just trying to make sure I'm aware of the quality of code I'm submitting.

For my last 15 pull requests (excluding the patch) only 1 on them had more than 5 commits in them. Two of them had a commit for the sake of getting CI validation. Three of those commits changed more than 5 files changing and each of those cases the large number of changes were doing to refactoring of tests.

I think this pull request is an exception which is why I advocated for the squash merge.

I think in general the number of noisy (read: useless for building CI) commits are low, but I think it would be worthwhile to prepend a [Refactor] tag to commit messages that refactor small changes in a large number of files. It might make it easier to focus on the important commits in the few larger PRs.

Because I definitely like to refactor and keep the code clean and admit that it makes PRs harder to navigate, so tagging those commits in a PR as [Refactor] I think will help. Ideas?

spathiwa:
Sure, that sounds like a good idea ([Refactor] tag).

I don't actually think the PRs I'm remembering are recent at all; I think one of them was back when you were getting AppVeyor working the first time, and it feels like there was at least one other time, also dealing with build system changes.   Sorry if I made this sound like more prevalent an issue than it is.  A lot of your changes are JS-only and I don't get auto-added as a code owner for them, so they don't contribute towards my impression.  It makes sense that build changes are likely to touch a lot more files and more likely to pull me in.
Anyway, just something to keep in mind, when appropriate, and that may well be very rare.  Thanks for your consideration.

package.json
```diff
@@ -13,27 +13,25 @@
...
     "xhr2": "^0.1.4"
   },
   "scripts": {
-    "test": "npm run lint && npm run karma -- --browsers Chrome",
+    "test": "karma start --browsers ChromeHeadlessNoSandbox,FirefoxHeadless",
+    "test-dev": "karma start --browsers ChromeHeadlessNoSandbox --skip-tags Slow",
+    "test-manual": "karma start --no-single-run",
+    "test-win": "karma start --browsers Chrome,Firefox,Edge && karma start --browsers IE --skip-tags FailsIE",
+    "test-coverage": "karma start karma.coverage.conf.js",
     "lint": "eslint .",
```
sanmut: I wonder if we should run "cpplint" also as part of the lint script.

rajsite: The npm scripts are just an abstraction around the js tooling (in-case we change linters, etc). If anything `make lint` should maybe call `npm run lint` and CPPLint. That would be a separate change.

PR #391, opened 2018-03-29
Implement Control References

spathiwa:
Implement control references.  …
Note: Control references are defined in the Locals: block of a VI (or inside an EventSpec),
rather than in a "ControlRefs:" block as we had discussed. (The reason is
that each ControlReference definition defines a local name which must be
visible as a local from within the clumps of the VI.)

Example:

define(MyVI dv(.VirtualInstrument (
    Locals: c(
        e(ControlReference("dataItem_Foo") ctlref1)  // static control ref linked to dataItem_Foo
        // -or-
        e(dv(ControlRefNum ControlReference("dataItem_Bar")) ctlref2)  // more explicit syntax
    ) ...

The actual Type of ctlref1 is ControlRefNum (as you can see by the more explicit syntax
using dv for ctlref2)  and stores a refnum (cookie) which opaquely holds onto the VI and 
control tag (data item).
The VI isn't passed directly, it's implicitly taken from the context,
whatever VI the ControlReference is defined in.  (Note, this is a refinement of
the original model we worked out which passed the VI's name as a string.)
ControlRefNums defined with ControlReference() exist during the entire
lifetime of the VI and are only deallocated when the last clump of the last
top-level VI finishes.

Non-statically-bound ControlRefNums variables can also be declared in the Params:
or Locals: sections, e.g.  'i(ControlRefNum ref)' or  'e(ControlRefNum ref)'.
These can be used in subVIs to refer to a reference passed in by a caller.

See test-it/ViaTests/ControlRefBasic.via for examples.

### DIFF
--- a/Vireo_VS/VireoCommandLine.vcxproj
+++ b/Vireo_VS/VireoCommandLine.vcxproj
@@ -15,6 +15,7 @@
     <ClCompile Include="..\source\core\Array.cpp" />
     <ClCompile Include="..\source\core\Assert.cpp" />
     <ClCompile Include="..\source\core\CEntryPoints.cpp" />
+    <ClCompile Include="..\source\core\ControlRef.cpp" />
     <ClCompile Include="..\source\core\DataQueue.cpp" />
     <ClCompile Include="..\source\core\Date.cpp" />
     <ClCompile Include="..\source\core\EventLog.cpp" />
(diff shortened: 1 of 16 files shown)


spathiwa:
FYI, I originally branched this before Memo's last PR, to propertyNode2 so I just pulled it in and rebased.
This PR has been updated (force pushed).

PR #394, opened 2018-04-02
Switch to using the official emsdk binaries on TravisCI Trusty

rajsite:
The approach is based on the discussion in the following thread: https://github.com/kripken/emscripten/pull/5087#issuecomment-320124568

The benefit of using this approach is being able to continue to use TravisCI container builds which startup quicker and being able to use the official binaries making it easier to track versions between AppVeyor, TravisCI, and local development.

The approach is as follows:
- Extract the bundled libstdc++6_5.1.0 deb (we include it in the repo because we cannot use apt to install it and launchpad serves over http and would need to verify a hash)
- Add the bundled libstd++6_5.1.0.deb to the path only when invoking the emsdk tools built with that version (it could causes problems to change it for the whole environment)
- extract the emsdk tools to a known working commit version (same as AppVeyor)
- install and activate emsdk and built (same as AppVeyor)


.travis.yml
```diff
@@ -27,20 +26,26 @@ install:
 - nohup npm run httpbin &
 - if [ "$CXX" = "g++" ]; then export CXX="g++-4.9" CC="gcc-4.9"; fi
 - gcc -v
-- if ! git clone --depth 1 https://github.com/ni-kismet/emscripten-sdk.git; then cd emscripten-sdk && git remote update -p && git merge --ff-only origin/master && cd ..; fi
-- emscripten-sdk/emsdk activate --build=Release sdk-master-64bit
+- mkdir -p emsdk-tools
+- cp make-it/travis-support/libstdc++6_5.1.0-0ubuntu11~10.04.2_amd64.deb emsdk-tools/
+- cd emsdk-tools && ar xv libstdc++6_5.1.0-0ubuntu11~10.04.2_amd64.deb && cd ..
```
spathiwa: Use a subshell instead?
(cd emsdk-tools && ar xv libstdc++6_5.1.0-0ubuntu11~10.04.2_amd64.deb )

.travis.yml
```diff
@@ -27,20 +26,26 @@ install:
...
+- tar -xzf emsdk-tools/data.tar.gz -C emsdk-tools
+- git clone https://github.com/juj/emsdk.git
+- cd emsdk && git checkout b2ebcffbbbc2ba64d5697dc5c37fed530bafecda && cd ..
```
spathiwa: What's this commit #?  Does it have a readable named tag you could use instead?

.travis.yml
rajsite: done! looks good

.travis.yml
rajsite: Unfortunately the emsdk repo doe not use tags :/

The emsdk tool recommends always using the latest from master and running the `emsdk update` command does that. Unfortunately that has broken our builds before so using a specific commit version.

This is the same commit version used in appveyor.yml so at least now we are using the same version of the emsdk tooling and emscripten libraries across both CIs

PR #397, opened 2018-04-03
Moved error formatting functions to coreHelpers.

gleonoliva:
Also cleaned up some eslint comments from module_propertyNode.


rajsite:
Looks good and I was the only reviewer so submitting

PR #398, opened 2018-04-04
Fix typos, clean up language and specify linux

mehagar:
The "shell" markdown format was showing the text "test", "in", and "if" as colored when it didn't make sense for it to be in this context, so I switched it to be "console". I put comments explaining what only works on Linux, and changed the directory where you're supposed to put new .via tests and test results to be located.


README.md
```diff
@@ -134,15 +136,16 @@ This will be a simple example that will add the test suite `rpi` with the `RpiTe
...
-```shell
+```console
 $ node ./test.js -n -t rpi
 ```
 
 ## Running the HTTP test server
 
 ### Overview
-Vireo utilizes [httpbin](https://httpbin.org/) for testing the HTTP Client functionality. The tests rely on a locally running instance of the httpbin server. The test suite will skip tests (without failing) that rely on the httpbin server running locally if the server is not running. However, if you would like to run the HTTP client tests locally these instructions will help you get configured.
+Vireo uses [httpbin](https://httpbin.org/) for testing the HTTP Client functionality. If the httpbin server is not running locally, the test suite will skip those tests that rely on it. If you would like to run the HTTP client tests locally these instructions show you how to do so.
```
rajsite: It's more accurate to say Vireo HTTP tests rely on a locally running instance of the [httpbin](https://github.com/kennethreitz/httpbin) server. Just saying we use httpbin might not be clear enough.

rajsite:
Looks good! thanks

PR #400, opened 2018-04-05
Fix multi-event structure support (leak), add test

spathiwa:


source/core/Events.cpp
```diff
@@ -123,6 +127,16 @@ class EventQueueObject {
     void SetStatus(QIDStatus status) { _status = status; }
 
  private:
+    bool DiscardTopEvent() {
+        if (_eventQueue.size() > 0) {
+            EventData &eventData = _eventQueue.front();
+            eventData.eventDataType->ClearData(eventData.pEventData);
+            THREAD_TADM()->Free(eventData.pEventData);
```
sanmut: We should probably have a single cleanup method in EventData that does both eventDataType->ClearData and Free(eventData.pEventData)

source/core/Events.cpp
```diff
@@ -155,6 +169,10 @@ class EventOracle {
     void DoneProcessingEvent(EventQueueID qID) {
         return _qObject[qID].DoneProcessingEvent();
     }
+    void ClearEventQueue(EventQueueID qID) {
+        if (qID < _qObject.size())
```
sanmut: Why not assert qID < _qObject.size() and then call _qObject[qID].ClearQueue();?

source/core/Events.cpp
```diff
@@ -155,6 +169,10 @@ class EventOracle {
...
+            _qObject[qID].ClearQueue();
```
sanmut: Can qID be 0 and should we ClearQueue in that case? Is that the kNotAQueueID case?

source/core/Events.cpp
spathiwa: This is called by clean up procs, so the queue could have been already freed explicitly, so this is not an 'assert' type condition, it could really be out of range for legitimate reasons.

source/core/Events.cpp
spathiwa: qID is not the kNotAQueueID case.  Calling clear on that qID shouldn't happen because the qID would have had to be valid before to set up the cleanup proc on it, but if it did it would be innocuous because it is unused and already empty.

source/core/Events.cpp
spathiwa: Done.

PR #402, opened 2018-04-05
JSLI typed array support

mehagar:
Following the conventions of the existing types support and tests I added functions and tests for each of the 8 different typed arrays we are now going to support.


source/io/JavaScriptInvoke.cpp
```diff
@@ -46,10 +46,19 @@ VIREO_EXPORT const char* JavaScriptInvoke_GetParameterType(StaticTypeAndData *pa
     if (returnBuffer) {
         SubString typeName = parameterType->Name();
         returnBuffer->Append(typeName.Length(), (Utf8Char*)typeName.Begin());
+
+        if (parameterType->IsArray() && !parameterType->IsString()) {
```
mehagar: I need to check that it is not a string explicitly here for arrays because string types in vireo are also array types, but I only want to append the subtype to javascript arrays and not javascript strings.

test-it/karma/javascriptinvoke/ReturnDataTypes.Test.js
```diff
@@ -212,6 +292,14 @@ describe('A JavaScript function invoke', function () {
             expect(viPathParser('returnSingle')).toBe(3.0);
             expect(viPathParser('returnDouble')).toBe(6.0);
             expect(viPathParser('returnString')).toBe('National Instruments');
+            expect(viPathParser('returnInt8Array')).toEqual([1, 2, 3]);
```
mehagar: I am using toEqual here because toBe uses === comparison which is not sufficient for array comparison. Also note that I am comparing the returned JSON to a regular array and not a TypedArray - I think this is because the JSON we get at this point is separate from the actual typed array type. I'm not completely sure on this, so let me know if that is the case or not.

source/io/module_javaScriptInvoke.js
```diff
@@ -153,198 +177,339 @@
...
+            (returnTypeName === 'undefined') ||
+            (isTypedArray(returnValue));
+        };
+
+        var updateReturnValueGivenType = function (
+            returnTypeName,
+            returnValuePointer,
+            returnValue) {
+            var javaScriptReturnTypeName = typeof returnValue;
             switch (returnTypeName) {
```
gleonoliva: I'm concerned about these switch statements that keep growing. Perhaps we can do something about them before is too late to refactor.
Consider having an object that has all the information of which functions to call depending on the type.
If we could have an structure similar to:
```javascript
var types_io = { // Please choose a better name.
	'Int8': {
		reader: Module.eggShell.dataReadInt8,
		writer: Module.eggShell.dataWriteInt8,
		isValidReturnType : isJavaScriptNumber
	}, ..., 
	'ArrayDouble' : {
		reader: Module.eggShell.dataReadDoubleArray,
		writer: Module.eggShell.dataWriteDoubleArray,
		isValidReturnType : isFloat64Array
	}
}
```

Then we could get rid of both switches that we have. As an example, here is how the `updateReturnValueGivenType` would look like.
```javascript
var updateReturnValueGivenType = function (
            returnTypeName,
            returnValuePointer,
            returnValue) {
	var javaScriptReturnTypeName = typeof returnValue;
	var typeConfig = types_io[returnTypeName]; // If this is undefined then the type is not registered.
	
	if (typeConfig.isValidReturnType(returnValue) === false) {
		// Not sure if we should throw here or just let the callee know that it failed.
	}
	
	var writer = typeConfig.writer(returnValuePointer, returnValue);
}
```

Something similar would happen with `createJavaScriptParametersArray()`. Also we could get rid of all the `update*ReturnValue()` family of functions since they only validate the value type and then call `dataWrite*()`. Here are the implementations for `isValidReturnType`:

```javascript
var isJavaScriptNumber = function (value) {
	return typeof value === 'number';
};

var isFloat64Array = function (value) {
	return value instanceof Float64Array;
};
```

gleonoliva: Last line of `updateReturnValueGivenType` is wrong.
```javascript
typeConfig.writer(returnValuePointer, returnValue);
```

spathiwa: Great suggestion.

mehagar: I like it. I needed to refactor this method to begin with to avoid a complexity linting rule, and this will make maintaining that requirement even easier.

source/io/module_eggShell.js
```diff
@@ -423,6 +429,54 @@
             return numericValue;
         };
 
+        Module.eggShell.dataReadInt8Array = function (arrayPointer) {
```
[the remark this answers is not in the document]
mehagar: The existing EggShell_GetArrayMetadata does not return the length of the array, but you say we should not add Data_GetTypedArrayLength. How else could I get the array length without adding a function like Data_GetTypedArrayLength?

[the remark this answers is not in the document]
mehagar: As for refactoring the dataReadAsXXArray methods, what would you think if I added additional fields to what memo suggested so for example:

```javascript
ArrayInt16: {
                reader: Module.eggShell.dataReadInt16Array,
                writer: Module.eggShell.dataWriteInt16Array,
                isValidReturnType: isInt16Array,
                heap: supportedArrayTypeConfig.Int16,
                constructor: Int16Array
           }
```

?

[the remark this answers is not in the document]
rajsite: Sorry about that, I overlooked that GetNumericArray was using the EggShell_GetArrayDimLength function. But following the same pattern as the others, instead of GetTypedArrayLength which hard codes reading dimension 0, make a Data_GetArrayDimLength and an EggShell_GetArrayDimLength.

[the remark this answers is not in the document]
rajsite: Right now these metadata objects are in different files so I don't think it is worth merging them yet unless we have a good way to describe what is being represented.

Then maybe we would also merge it with the object used by getNumericArray and move it to core or something.

mehagar:
Whoops, I forgot to remove a debugging aid #include.

source/io/module_javaScriptInvoke.js
```diff
@@ -153,198 +264,85 @@
...
 
-            if (returnTypeMistmatch) {
-                newErrorStatus = true;
-                newErrorCode = ERRORS.kNITypeMistmatchForReturnTypeInJavaScriptInvoke.CODE;
-                newErrorSource = ERRORS.kNITypeMistmatchForReturnTypeInJavaScriptInvoke.MESSAGE + '\'' + functionNameString +
-                    '\'. JavaScript return type is \'' + javaScriptReturnTypeName +
-                    '\'. LabVIEW return type is \'' + returnTypeName + '\'';
-                Module.coreHelpers.mergeErrors(newErrorStatus, newErrorCode, newErrorSource, errorStatusPointer, errorCodePointer, errorSourcePointer);
+            try {
+                updateReturnValueGivenType(returnTypeName, returnValuePointer, returnValue);
```
gleonoliva: .... here. This is the second time we check if it is a a valid JS return type. From what I see in the function `updateReturnValueGivenType()` the code is added as the name of the exception but I don't see it being used here.

gleonoliva: I think we can get rid of the `isValidJavaScriptReturnType` function as well as `isTypedArray` since we now have the `isValidReturnType` family of functions declared in `functionTypes`. Basically would be the work of moving the validation in `updateReturnValueGivenType()` outside of it and validating before calling it.

source/core/CEntryPoints.cpp
```diff
@@ -235,18 +249,14 @@ VIREO_EXPORT EggShellResult EggShell_GetArrayMetadata(TypeManagerRef tm,
     }
 
     SubString arrayTypeNameSubString = pathType->GetSubElement(0)->Name();
-    arrayTypeNameBuffer->Append(arrayTypeNameSubString.Length(), (Utf8Char*)arrayTypeNameSubString.Begin());
+    CopyArrayTypeNameStringToBuffer(arrayTypeNameBuffer, arrayTypeNameSubString);
     arrayTypeNameBuffer->Append((Utf8Char)'\0');
```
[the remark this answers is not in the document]
mehagar: Yes, you're right. I removed the unnecessary line.

source/core/CEntryPoints.cpp
```diff
@@ -316,6 +326,60 @@ VIREO_EXPORT Int32 Data_GetTypedArrayLength(TypedArrayCoreRef arrayObject)
...
+        return kEggShellResult_UnableToCreateReturnBuffer;
+    }
+
+    TypeRef arrayElementType = arrayObject->ElementType();
+    CopyArrayTypeNameStringToBuffer(arrayTypeNameBuffer, arrayElementType->Name());
+    *arrayTypeName = (char*) arrayTypeNameBuffer->Begin();
+
+    *arrayRank = arrayObject->Rank();
+
+    if (arrayObject->GetLength(0) <= 0) {
```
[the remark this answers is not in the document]
mehagar: Done

source/io/module_eggShell.js
```diff
@@ -40,6 +40,8 @@
             'EggShell_ResizeArray',
             'Data_GetStringBegin',
             'Data_GetStringLength',
+            'Data_GetTypedArrayBegin',
+            'Data_GetTypedArrayLength',
```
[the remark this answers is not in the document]
mehagar: Yep. Removed.

make-it/EmMakefile
```diff
@@ -134,6 +134,11 @@ EM_EXPORTS = -s EXPORTED_FUNCTIONS="[\
     '_Data_WriteUInt32',\
     '_Data_WriteSingle',\
     '_Data_WriteDouble',\
+    '_Data_GetTypedArrayBegin',\
```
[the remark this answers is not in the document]
mehagar: Done

source/io/module_eggShell.js
```diff
@@ -189,14 +197,38 @@
...
-            Int16: Module.HEAP16,
-            Int32: Module.HEAP32,
-            UInt8: Module.HEAPU8,
-            UInt16: Module.HEAPU16,
-            UInt32: Module.HEAPU32,
-            Single: Module.HEAPF32,
-            Double: Module.HEAPF64
+            Int8: {
+                heap: Module.HEAP8,
+                constructor: Int8Array
```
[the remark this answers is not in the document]
mehagar: Didn't know that.

source/io/module_eggShell.js
```diff
@@ -423,6 +455,25 @@
             return numericValue;
         };
 
+        Module.eggShell.dataReadTypedArray = function (arrayPointer) {
```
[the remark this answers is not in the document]
mehagar: I did the rename, but I wasn't sure how to have getNumericArray call dataReadNumericArrayAsTypedArray. The latter takes a Vireo Array pointer, and I couldn't figure out a way to get an array pointer in getNumericArray to pass into it.

source/core/CEntryPoints.cpp
```diff
@@ -303,6 +314,72 @@ VIREO_EXPORT Int32 Data_GetStringLength(StringRef stringObject)
...
+    } else {
+        *arrayBegin = arrayObject->BeginAt(0);
+    }
+
+    return kEggShellResult_Success;
+}
+//------------------------------------------------------------
+//! Get the Length of a dimension in an Array Symbol. Returns -1 if the Symbol is not found or not
+//! an Array or dimension requested is out of the bounds of the rank.
+VIREO_EXPORT Int32 Data_GetArrayDimLength(TypeManagerRef tm, TypedArrayCoreRef arrayObject, Int32 dim)
```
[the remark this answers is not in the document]
mehagar: Done.

test-it/karma/fixtures/javascriptinvoke/DataTypes.via
```diff
@@ -27,6 +27,14 @@ define(MyVI dv(VirtualInstrument (
         e(dv(.Single 3.0) singleValue)
         e(dv(.Double 6.0) doubleValue)
         e(dv(.String "National Instruments") stringValue)
+        e(dv(a(.Int8 *) (1 2 3)) arrayInt8)
```
cglzaguilar: For integer types, please add the Min and Max value in the integer range. For floating point values please do negative, zero, positive values, at this moment I don't have a better idea about testing them

source/core/CEntryPoints.cpp
```diff
@@ -199,81 +199,62 @@ VIREO_EXPORT const char* EggShell_ReadValueString(TypeManagerRef tm,
...
-
     SubString objectName(viName);
     SubString path(eltName);
+    void *pData = null;
+
     TypeRef pathType = tm->GetObjectElementAddressFromPath(&objectName, &path, &pData, true);
     if (pathType == null)
         return kEggShellResult_ObjectNotFoundAtPath;
 
-    if (!pathType->IsArray() || pathType->Rank() <= 0)
```
cglzaguilar: Please elaborate why this check is no longer needed.

source/io/module_eggShell.js
```diff
@@ -460,6 +509,29 @@
             Data_WriteDouble(destination, value);
         };
 
+        Module.eggShell.dataWriteTypedArray = function (destination, value) {
+            var int32Byte = 4;
+            var rank = 1;
+            var newLengths = Module._malloc(rank * int32Byte);
```
cglzaguilar: We need to deallocate this memory. Otherwise, it will become a memory leak.

test-it/karma/javascriptinvoke/DataTypes.Test.js
```diff
@@ -108,6 +116,14 @@ describe('A JavaScript function invoke', function () {
             expect(window.NI_SingleFunction).toHaveBeenCalledWith(3.0);
             expect(window.NI_DoubleFunction).toHaveBeenCalledWith(6.0);
             expect(window.NI_StringFunction).toHaveBeenCalledWith('National Instruments');
+            expect(window.NI_ArrayFunction).toHaveBeenCalledWith(Int8Array.from([1, 2, 3]));
```
cglzaguilar: For integer types, please add the Min and Max value in the integer range. For floating point values please do negative, zero, positive values, at this moment I don't have a better idea about testing them

test-it/karma/javascriptinvoke/ReturnDataTypes.Test.js
```diff
@@ -98,6 +97,38 @@ describe('A JavaScript function invoke', function () {
             return 'National Instruments';
         };
 
+        window.NI_Int8ArrayFunction = function () {
+            return Int8Array.from([1, 2, 3]);
+        };
+
+        window.NI_Int16ArrayFunction = function () {
+            return Int16Array.from([1, 2, 3]);
```
cglzaguilar: For integer types, please add the Min and Max value in the integer range. For floating point values please do negative, zero, positive values, at this moment I don't have a better idea about testing them

source/io/JavaScriptInvoke.cpp
rajsite: That's a good idea 👍

source/io/module_javaScriptInvoke.js
```diff
@@ -153,199 +264,39 @@
...
+            var returnValueIndex = 0;
+            var returnTypeName = getParameterTypeString(returnPointer, returnValueIndex);
 
-            var javaScriptReturnTypeName = typeof returnValue;
-            var validJavaScriptReturnType =
-                (javaScriptReturnTypeName === 'number') ||
-                (javaScriptReturnTypeName === 'boolean') ||
-                (javaScriptReturnTypeName === 'string') ||
-                (javaScriptReturnTypeName === 'undefined');
-            if (!validJavaScriptReturnType) {
```
cglzaguilar: Please elaborate why this check is no longer needed. The intent was to return an error when the user was returning a JavaScript type we currently don't support.

source/core/CEntryPoints.cpp
rajsite: Shouldn't the Rank() <= 0 check be moved to Data_GetArrayMetadata?

mehagar: Because it's checking the path and in Data_GetArrayMetadata we've already working with the array pointer instead of the path at this point, that check should indeed still exist but exist in the EggShell_GetArrayMetadata (which I am changing to be called EggShell_ArrayPointer)

source/io/module_javaScriptInvoke.js
```diff
@@ -90,53 +90,164 @@
             return typeName;
         };
 
+        var isJavaScriptNumber = function (value) {
```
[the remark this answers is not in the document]
mehagar: C# mentality man.

source/io/module_eggShell.js
cglzaguilar: Consider the function may throw and the deallocation needs to happen before that.

source/io/module_javaScriptInvoke.js
mehagar: I added this check back.

source/io/module_eggShell.js
mehagar: Fixed.

source/include/CEntryPoints.h
```diff
@@ -41,13 +41,17 @@ VIREO_EXPORT void EggShell_WriteValueString(TypeManagerRef tm, const char* viNam
...
-VIREO_EXPORT EggShellResult EggShell_GetArrayMetadata(TypeManagerRef tm, const char* viName, const char* eltName,
-    char** arrayTypeName, Int32* arrayRank, unsigned char** arrayBegin);
+VIREO_EXPORT EggShellResult EggShell_GetArrayPointer(TypeManagerRef tm,
+        const char* viName, const char* eltName, void** arrayPointer);
 VIREO_EXPORT Int32 EggShell_GetArrayDimLength(TypeManagerRef tm, const char* viName, const char* eltName, Int32 dim);
 VIREO_EXPORT Int32 EggShell_ResizeArray(TypeManagerRef tm, const char* viName, const char* eltName,
                                         Int32 rank, Int32* newLengths);
 VIREO_EXPORT void* Data_GetStringBegin(StringRef stringObject);
 VIREO_EXPORT Int32 Data_GetStringLength(StringRef stringObject);
+VIREO_EXPORT EggShellResult Data_GetArrayMetadata(TypeManagerRef tm,
```
cglzaguilar: I wonder if this function is used in the JS control layer of the Web VI. Please run an LVjs CI with the changes and report the results.

source/io/JavaScriptInvoke.cpp
```diff
@@ -43,13 +40,40 @@ VIREO_EXPORT const char* JavaScriptInvoke_GetParameterType(StaticTypeAndData *pa
...
     if (returnBuffer) {
         SubString typeName = parameterType->Name();
         returnBuffer->Append(typeName.Length(), (Utf8Char*)typeName.Begin());
+
         // Add an explicit null terminator so it looks like a C string.
         returnBuffer->Append((Utf8Char)'\0');
         return (const char*)returnBuffer->Begin();
     }
+
+    return "";
```
rajsite: It might make more sense for this to return null. We don't really know the lifetime of that empty string once the function returns.

source/io/JavaScriptInvoke.cpp
```diff
@@ -43,13 +40,40 @@ VIREO_EXPORT const char* JavaScriptInvoke_GetParameterType(StaticTypeAndData *pa
...
+    if (returnBuffer) {
+        SubString elementTypeName = arrayObject->ElementType()->Name();
+        returnBuffer->Append(elementTypeName.Length(), (Utf8Char*)elementTypeName.Begin());
+
+        // Add an explicit null terminator so it looks like a C string.
+        returnBuffer->Append((Utf8Char)'\0');
+        return (const char*)returnBuffer->Begin();
+    }
+
     return "";
```
rajsite: Same as above, should return null instead

source/io/JavaScriptInvoke.cpp
```diff
@@ -63,7 +87,10 @@ VIREO_EXPORT void* JavaScriptInvoke_GetParameterPointer(StaticTypeAndData *param
         // We a have pointer to a StringRef, we just need the StringRef.
         // So we can use functions that already work with StringRef on the JavaScript side.
         pData = *(StringRef*)pData;
+    } else if (parameterType->IsArray()) {
+        pData = *(TypedArrayCoreRef*)pData;
     }
+
```
rajsite: Similar to the other EggShell_GetPointer we should validate supported cases and return null otherwise

source/io/module_eggShell.js
```diff
@@ -260,15 +292,15 @@
...
+                    currArr[i] = arr[i];
                 }
                 return currArr;
             }
 
             // Perform nd array creation for rank > 1
             // TODO mraj this is O((m-1)n) for rank m. So rank 2 is O(n) and can be improved for rank > 2
-            currArr = arr;
-            var currStartIndex = startIndex;
+            currArr = Array.from(arr);
```
rajsite: The Array.from is going to make a full copy of the arr Array. Is making a copy of the array needed? Seems like it can be expensive

test-it/karma/javascriptinvoke/DataTypes.Test.js
```diff
@@ -66,6 +66,46 @@ describe('A JavaScript function invoke', function () {
         window.NI_StringFunction = function (value) {
             return value;
         };
+
+        window.NI_Int8ArrayFunction = function (value) {
+            expect(value instanceof Int8Array).toBeTruthy();
```
rajsite: prefer toBeTrue

Truthiness is definitely a flaw of JS, should avoid relying on implicit type conversions

make-it/EmMakefile
```diff
@@ -110,9 +110,10 @@ EM_EXPORTS = -s EXPORTED_FUNCTIONS="[\
     '_EggShell_ReadValueString',\
     '_EggShell_WriteValueString',\
     '_EggShell_ResizeArray',\
-    '_EggShell_GetArrayMetadata',\
+    '_EggShell_GetPointer',\
     '_EggShell_GetArrayDimLength',\
     '_Occurrence_Set',\
+	'_Data_ValidateArrayType',\
```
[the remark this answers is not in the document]
mehagar: Fixed.

source/io/module_eggShell.js
mehagar: You're right - I didn't need the Array.from here. I removed it.

source/io/JavaScriptInvoke.cpp
```diff
@@ -89,6 +89,8 @@ VIREO_EXPORT void* JavaScriptInvoke_GetParameterPointer(StaticTypeAndData *param
         pData = *(StringRef*)pData;
     } else if (parameterType->IsArray()) {
         pData = *(TypedArrayCoreRef*)pData;
+    } else if (!(parameterType->IsNumeric() || parameterType->IsString() || parameterType->IsFloat() || parameterType->IsBoolean())) {
```
rajsite: IsNumeric allows Int64 and IntU64 which are not supported

rajsite:
@mehagar Can you give me the comment for the squash merge? The pull request comment is light on details.

mehagar:
@rajsite What about "Add JSLI support for 1D typed arrays and refactor EggShell_GetArrayMetadata function"?

PR #410, opened 2018-04-17
Prevent builds on gh-pages branch and keep history

rajsite:
Based on discussion in https://github.com/travis-ci/travis-ci/issues/7737


rajsite:
Buddied with @cglzaguilar in person

PR #414, opened 2018-04-18
Add support for clusters of event reg. refs to WaitForEvents and fix race condition in Observables 

spathiwa:
(The race condition appears to be the source of the long-standing intermittent CI failure in Queue tests.)

### DIFF
--- a/source/CommandLine/main.cpp
+++ b/source/CommandLine/main.cpp
@@ -115,6 +115,8 @@ int VIREO_MAIN(int argc, const char * argv[])
 void Vireo::RunExec() {
     TypeManagerRef tm = gShells._pUserShell;
     TypeManagerScope scope(tm);
+    // These numbers may need further tuning (numSlices and millisecondsToRun).
+    // They should match the values for VJS in io/module_eggShell.js
     Int32 state = tm->TheExecutionContext()->ExecuteSlices(10000, 4);
     Int32 delay = state > 0 ? state : 0;
     gShells._keepRunning = (state != kExecSlices_ClumpsFinished);
--- a/source/CommandLine/main.cpp
+++ b/source/CommandLine/main.cpp
@@ -115,7 +115,7 @@ int VIREO_MAIN(int argc, const char * argv[])
 void Vireo::RunExec() {
     TypeManagerRef tm = gShells._pUserShell;
     TypeManagerScope scope(tm);
-    Int32 state = tm->TheExecutionContext()->ExecuteSlices(400, 10000000);
+    Int32 state = tm->TheExecutionContext()->ExecuteSlices(10000, 4);
     Int32 delay = state > 0 ? state : 0;
     gShells._keepRunning = (state != kExecSlices_ClumpsFinished);
     if (delay)
--- a/source/core/Events.cpp
+++ b/source/core/Events.cpp
@@ -431,7 +431,10 @@ Int32 EventOracle::GetPendingEventInfo(EventQueueID *pActiveQID, Int32 nQueues, 
         DynamicEventRegInfo *regInfo = NULL;
         if (dynRegRefs
             && EventRegistrationRefNumManager::RefNumStorage().GetRefNumData(dynRegRefs[regIndex].GetRefNum(), &regInfo) == kNIError_Success) {
-            if (_qObject[regInfo->_qID].GetPendingEventSequenceNumber(&eventSeq) && (!eventQID || EventTimeSeqCompare(eventSeq, earliestSeq) < 0)) {
+            // For each valid dynamic reference, find the event with the earliest event sequence number in its queue.
+            if (_qObject[regInfo->_qID].GetPendingEventSequenceNumber(&eventSeq)
+                && (!eventQID  // first time through loop, take starting seq number
+                    || EventTimeSeqCompare(eventSeq, earliestSeq) < 0)) {
                 eventQID = regInfo->_qID;
                 earliestSeq = eventSeq;
                 earliestIndex = regIndex+1;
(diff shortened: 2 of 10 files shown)


spathiwa:
[Closed and recreated this PR to force the CIs to run again, since AppVeyor got a spurious failure in npm install.)

source/core/Events.cpp
```diff
@@ -396,6 +409,42 @@ class EventRegistrationRefNumManager : public RefNumManager {
 
 EventRegistrationRefNumManager EventRegistrationRefNumManager::_s_singleton;
 
+// GetPendingEventInfo -- given a static queue and a set of dynamic event queuse (passed via refnum), return the queue with the earliest event
```
cglzaguilar: Fix typo

spathiwa: Fixed.

Makefile
```diff
@@ -24,7 +24,7 @@ testhttpbin:
 	cd test-it && node test.js -j -t httpbin --dots
 
 testjs:
-	cd test-it && node test.js -j --dots
+	cd test-it && node test.js -j
```
cglzaguilar: Just for consistency, please make the same change to testnative case. If I remember correctly in failure conditions having the dots is not really helpful.

source/CommandLine/main.cpp
```diff
@@ -115,7 +115,7 @@ int VIREO_MAIN(int argc, const char * argv[])
 void Vireo::RunExec() {
     TypeManagerRef tm = gShells._pUserShell;
     TypeManagerScope scope(tm);
-    Int32 state = tm->TheExecutionContext()->ExecuteSlices(400, 10000000);
+    Int32 state = tm->TheExecutionContext()->ExecuteSlices(10000, 4);
```
cglzaguilar: Please add a comment as to why we decided to change the value. I know is not clear what would be the best value, but having a history of changes with explanations might help in the future.

source/core/Events.cpp
```diff
@@ -45,6 +45,9 @@ UserEventRefNumManager UserEventRefNumManager::_s_singleton;
 typedef Int32 EventQueueID;
 enum { kNotAQueueID };
 
+// Monotonically incremented event number, so we can always tell which event is generated first even if timestamps are the same
+UInt32 EventData::_s_eventSequenceNumber = 0;
```
cglzaguilar: Are we concerned about this number wrapping around when the value gets to Max number in the UInt32 range?

source/core/Events.cpp
```diff
@@ -396,6 +409,42 @@ class EventRegistrationRefNumManager : public RefNumManager {
...
+            earliestSeq = eventSeq;
+        } else {
+            eventQID = kNotAQueueID;
+        }
+    }
+    while (regIndex < nQueues) {
+        DynamicEventRegInfo *regInfo = NULL;
+        if (dynRegRefs
+            && EventRegistrationRefNumManager::RefNumStorage().GetRefNumData(dynRegRefs[regIndex].GetRefNum(), &regInfo) == kNIError_Success) {
+            if (_qObject[regInfo->_qID].GetPendingEventSequenceNumber(&eventSeq) && (!eventQID || EventTimeSeqCompare(eventSeq, earliestSeq) < 0)) {
```
cglzaguilar: For long if statements consider using a local boolean named with the intent of the check, basically the name should answer the question, what is the code checking here? and then use the boolean in the if statement. This is helpful for readability as long as it doesn't have a performance impact.

Apply this rule where you consider is useful in other sections of the code.

source/io/module_eggShell.js
```diff
@@ -517,7 +517,7 @@
         // A good starting point for most vireo uses but can be copied and modified as needed
         // If a callback (stdout, stderr) is provided, it will be run asynchronously to completion
         Module.eggShell.executeSlicesUntilClumpsFinished = publicAPI.eggShell.executeSlicesUntilClumpsFinished = function (callback) {
-            var SLICE_SETS_PER_TIME_CHECK = 100000;
+            var SLICE_SETS_PER_TIME_CHECK = 10000;
```
cglzaguilar: Please add a comment as to why we want this value.

Makefile
spathiwa: This was mistakenly submitted.  It was for debugging a CI failure and I forgot to back it out.  Retracted.

spathiwa: Although I agree the dots are not helpful, Milan really like them though.  If you agree they should be removed I wouldn't be averse to trying to win over Milan.

source/core/Events.cpp
spathiwa: This is actually handled -- the comparison of this value with previous values if done using a helper function EventTimeSeqCompare which handles wrapping.

source/core/Events.cpp
spathiwa: I agree in general, but this is used in dispatching events, so I am actually concerned about performance.  I have added more comments to help readability, though.

source/io/module_eggShell.js
spathiwa: Done.

source/CommandLine/main.cpp
spathiwa: Done; added a reference to the other place in JS with a note they should be in sync.

rajsite:
Since the TFS side does not use executeSlicesUntilClumpsFinished directly, is there a block diagram test that can be written to test the behavior of the loop on that side?

spathiwa:
I'd say we leave them tunable for testing while we discern whether the current values are good or still have issues (or find any reason they might need to be different in different contexts), and then turn them to constants instead of parameters.

PR #417, opened 2018-04-18
Add 'ce' (const element) type, which can be used instead of 'e' for const locals.

spathiwa:
For Gandi so she can move global constants into individual VI's Locals: section.

E.g.
define(HelloWorld dv(.VirtualInstrument (
  Locals: c(
      ce(dv(.Int32 456) c1)   // ce means value is constant
      e(dv(.Int32 123) local1)
  )
  clump(1
    Copy(c1 local1)  // ok
    Copy(local1 c1)  // error - c1 is immutable
  )
)))


rajsite:
Is this just a way to make the via syntax more expressive of the intent for the value? Conceptually we could just emit a local assuming dfir does not emit anything to overwrite its value?

spathiwa:
@rajsite Basically, yes.  This is for Gandi's work to move the definition of constants from global defines to VI locals.  She could just use normal mutable locals, but it'd be nice to get errors from Vireo if we accidentally emit anything that tries to overwrite them, so this allows that.

sanmut:
Please add test cases for the following:
define constants using "ce" for the data types int, bool, string, array, and cluster and try to copy to these constants and check for the "not mutable" error. I think for clusters, we should try to write to fields of clusters also directly and make sure they are immutable too.

spathiwa:
Added test cases except for cluster fields; filed an issue for this, because this is difficult to check for efficiently.

sanmut:
Talked to Craig about it. Field type in a cluster marked constant (with ce) makes it const. This is acceptable for now to get us going. When we emit cluster constants from G, we will mark all the fields also constant.

PR #418, opened 2018-04-18
Fix for propertyNode path format

gleonoliva:
The main purpose of this change is to change the format of the path that propertyRead/Write JS callbacks receive.
Previously we were using `TypeManager::PointerToSymbolPath` however this method adds the decoded VI name and the `Locals` section to the resulting path.

I started just tweaking and modifying TypeAndDataReflection, but since it already has some tests and I didn't want to break it, I added TypeAndDataPathFinder visitor which is based on the former.

I also added a few more cases to `PropertyNode.Test.js` to validate that the path passed to the js callbacks can be used to read/write from Vireo.

A few more comments before starting.
1. I ended up not using the EncodedSubString I implemented because while building the path we are already using encoded names.
2. Ended up not saving the encoded VI name since JS uses the decoded VI name to search for the control model. The model (on JS side) has the encoded VI name which can be used to read/write to/from vireo.

### DIFF
--- a/source/core/TypeAndDataPathFinder.cpp
+++ b/source/core/TypeAndDataPathFinder.cpp
@@ -11,79 +11,48 @@ SDG
  */
 
 #include "TypeAndDataManager.h"
+#include "DataReflectionVisitor.h"
 
 namespace Vireo
 {
 
 //------------------------------------------------------------
-/* Create a visitor with a needle being looked for and a TypeRef
-* that describes what the needle is a pointer to. The visitor will visit
-* all types in the type manger and thus ultimate all values owned by that type manager
-* and if necessary parent type managers as well.
-* initially the hay stack is null, but when a type is visited that owns a value ( constant or var)
-* it will establish a haystack then as the type is visited the stack is narrowed.
+/* Inherits searching behavior from DataReflectionVisitor to build
+* a path for the specified needle and haystack. Path format does
+* not include the VI name nor "Locals" section.
 */
 
-class PathFinderVisitor : public TypeVisitor
+class PathFinderVisitor : public DataReflectionVisitor
 {
  public:
     PathFinderVisitor(TypeRef tNeedle, DataPointer pHaystack, StringRef path);
-    void Accept(TypeRef tHaystack, DataPointer pHaystack);
     void Accept(TypeManagerRef tm);
-    Boolean Found() {return _found;}
 
  private:
-    // What is being searched through
-    void*           _pHayStack;
-
-    // What is being looked for
-    TypeRef         _tNeedle;
-    void*           _pNeedle;
-
     // Used as frames are unwound
     Boolean         _pathEmpty;
-    Boolean         _found;
-    StringRef       _path;
 
  private:
-    virtual void VisitBad(TypeRef type);
-    virtual void VisitBitBlock(BitBlockType* type);
-    virtual void VisitBitCluster(BitClusterType* type);
     virtual void VisitCluster(ClusterType* type);
-    virtual void VisitParamBlock(ParamBlockType* type);
-    virtual void VisitEquivalence(EquivalenceType* type);
-    virtual void VisitArray(ArrayType* type);
-    virtual void VisitElement(ElementType* type);
-    virtual void VisitNamed(NamedType* type);
-    virtual void VisitPointer(PointerType* type);
-    virtual void VisitEnum(EnumType* type);
-    virtual void VisitRefNumVal(RefNumValType* type);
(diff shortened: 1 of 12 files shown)


source/core/TypeAndDataPathFinder.cpp
```diff
@@ -0,0 +1,244 @@
+/**
+
+Copyright (c) 2015 National Instruments Corp.
```
cglzaguilar: Change 2015 to 2018

source/core/TypeAndDataPathFinder.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+#include "TypeAndDataManager.h"
+
+namespace Vireo
+{
+
+//------------------------------------------------------------
+/* Create a visitor with a needle being looked for and a TypeRef
+* that describes what the needle is a pointer to. The visitor will visit
+* all types in the type manger and thus ultimate all values owned by that type manager
```
cglzaguilar: Fix typo "manger" to manager is this file and the where this file was copied from.

source/io/PropertyNode.cpp
```diff
@@ -82,12 +82,12 @@ VIREO_FUNCTION_SIGNATUREV(PropertyNodeWrite, PropertyNodeWriteParamBlock)
...
     StringRef dataItemId = dataItemIdVar.Value;
     if (!LookupControlRefForPropertyNode(refNumPtr, errorClusterPtr, viName, dataItemId, propNodeWriteName))
-        return _NextInstruction();  // control refnum lookup failed and set errorCluter
+        return _NextInstruction();  // control refnum lookup failed and set errorCluster
 
     TypeManagerRef typeManager = value->_paramType->TheTypeManager();
 
     STACK_VAR(String, pathRef);
-    typeManager->PointerToSymbolPath(value->_paramType, value->_pData, pathRef.Value);
+    typeManager->GetPathFromPointer(value->_paramType, value->_pData, pathRef.Value);
```
cglzaguilar: So we are not returning a symbol path. Consider renaming the function to be more specific about the nature of the path we are returning.

source/core/TypeAndDataPathFinder.cpp
```diff
@@ -0,0 +1,241 @@
...
+        // Dig deeper into the data.
+        void* saveData = _pHayStack;
+        _pHayStack = pHayStack;
+        tHayStack->Accept(this);
+        _pHayStack = saveData;
+    }
+}
+//------------------------------------------------------------
+void PathFinderVisitor::VisitBad(TypeRef type)
+{
```
spathiwa: Can you move the empty Visit members and the trivial ones that just call Accept() to be inline in the class definition?

source/core/TypeAndDataPathFinder.cpp
```diff
@@ -0,0 +1,241 @@
...
+        return;
+
+    IntIndex count = type->SubElementCount();
+    for (IntIndex j = 0; j < count; j++) {
+        TypeRef elementType = type->GetSubElement(j);
+        IntIndex offset = elementType->ElementOffset();
+        AQBlock1* pElementData = (AQBlock1*)_pHayStack + offset;
+        Accept(elementType, pElementData);
+        if (_found) {
+            if (!elementType->ElementName().CompareCStr("Locals")) {
```
spathiwa: This seems like it would get false positives -- anyone could name a cluster or cluster elements "Locals".
You only want to identify and avoid the top-level Locals cluster under VI, right?
Could you keep track of search depth or something to make this test more targeted?

source/core/TypeAndDataPathFinder.cpp
```diff
@@ -0,0 +1,117 @@
...
+namespace Vireo
+{
+
+//------------------------------------------------------------
+/* Inherits searching behavior from DataReflectionVisitor to build
+* a path for the specified needle and haystack. Path format does
+* not include the VI name nor "Locals" section.
+*/
+
+class PathFinderVisitor : public DataReflectionVisitor
```
sanmut: Correct me if I am wrong. DataReflectionVisitor does not seem to be used or exported to JS layer except for one test (.via file). In that case, why don't we fix that directly to what we want and update the test result file with the new output? If we don't want to do that, I am curious what use cases due to which we need to maintain DataReflectionVisitor and PathFinderVisitor?

gleonoliva:
After discussing this approach we decided to go with a different one. I'll create another pull request that encodes the VI name and uses the current TypeAndDataReflection but instead builds on top of it.

PR #420, opened 2018-04-19
add appveyor-retry command

mehagar:
See appveyor/ci#418 for how script works. The idea to use it came from https://help.appveyor.com/discussions/problems/3147-error-eperm-operation-not-permitted-rename-on-npm-install (See the last comment).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -11,8 +11,8 @@ init:
 - ps: $PSVersionTable
 install:
   - ps: Install-Product node 8
-  - chocolatey install make
-  - chocolatey install gnuwin32-coreutils.portable
+  - appveyor-retry chocolatey install make
+  - appveyor-retry chocolatey install gnuwin32-coreutils.portable
   - appveyor-retry npm install
   - pip install tox
   - ps: Start-Process npm "run httpbin" -PassThru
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -13,7 +13,7 @@ install:
   - ps: Install-Product node 8
   - chocolatey install make
   - chocolatey install gnuwin32-coreutils.portable
-  - npm install
+  - appveyor-retry npm install
```
sanmut: This is great. Good find. 
Can (should) we set the number of retries at all?

.appveyor.yml
```diff
@@ -13,7 +13,7 @@ install:
   - ps: Install-Product node 8
   - chocolatey install make
   - chocolatey install gnuwin32-coreutils.portable
```
rajsite: Can you add the retry to the two above `chocolatey install` lines as well? They also fail appveyor builds occasionally.

.appveyor.yml
mehagar: I don't think so.

.appveyor.yml
mehagar: Done.

PR #422, opened 2018-04-19
Add support for async calls through the JavaScriptInvoke function

siddhukrs:
Updated the JavaScriptInvoke method call to include an occurrence pointer, similar to the existing HTTP methods. With this change, synchronous calls to JavaScriptInvoke will also create an occurrence and wait for the occurrence to be set before completing the call. I am come up with details about the performance implications of this change on synchronous calls before this PR gets merged.

Added tests to exercise this behavior.

This change is to add support for async function calls through the JSLI document.


test-it/karma/fixtures/javascriptinvoke/AsyncFunctions.via
```diff
@@ -0,0 +1,58 @@
...
+        e(dv(.UInt8 6) inputCallCallback)
+        e(dv(.UInt8 0) returnCallCallback)
+        e(.Occurrence occurrence)
+    )
+    clump (
+        JavaScriptInvoke(occurrence false error 'NI_CallCompletionCallbackMoreThanOnce' returnCallCallback inputCallCallback)
+        JavaScriptInvoke(occurrence false error 'NI_CallCompletionCallbackMoreThanOnceAfterSecondCallbackRetrieval' returnCallCallback inputCallCallback)
+        JavaScriptInvoke(occurrence false error 'NI_CompletionCallbackReturnsUndefined' * inputCallCallback)
+    )
+) ) )
```
cglzaguilar: Add an empty line at the end of the file to remove the red icon on the diff

source/io/module_javaScriptInvoke.js
```diff
@@ -285,49 +285,62 @@
...
 
-            typeConfig.writer(returnValuePointer, returnValue);
+            typeConfig.writer(returnValuePointer, returnUserValue);
+        };
+
+        var generateCompletionCallback = function (occurrencePointer, functionName, returnTypeName, returnValuePointer, errorStatusPointer, errorCodePointer, errorSourcePointer) {
+            var completionCallbackInvoked = false;
+            return function (returnValue) {
+                if (completionCallbackInvoked === true) {
+                    throw new Error('The completion callback was invoked more than once.');
```
cglzaguilar: Add the function name as discussed in the error string.

source/io/module_javaScriptInvoke.js
```diff
@@ -285,49 +285,62 @@
...
                 return;
             }
 
-            typeConfig.writer(returnValuePointer, returnValue);
+            typeConfig.writer(returnValuePointer, returnUserValue);
+        };
+
+        var generateCompletionCallback = function (occurrencePointer, functionName, returnTypeName, returnValuePointer, errorStatusPointer, errorCodePointer, errorSourcePointer) {
+            var completionCallbackInvoked = false;
+            return function (returnValue) {
```
rajsite: It would be nice to name this function so it has a good name in stack traces

source/io/module_javaScriptInvoke.js
```diff
@@ -358,30 +372,52 @@
...
             var returnValue = undefined;
+            var context = generateContext();
             try {
                 returnValue = functionToCall.apply(context, parameters);
             } catch (ex) {
                 newErrorStatus = true;
                 newErrorCode = ERRORS.kNIUnableToInvokeAJavaScriptFunction.CODE;
                 newErrorSource = Module.coreHelpers.formatMessageWithException(ERRORS.kNIUnableToInvokeAJavaScriptFunction.MESSAGE + '\'' + functionName + '\'.', ex);
                 Module.coreHelpers.mergeErrors(newErrorStatus, newErrorCode, newErrorSource, errorStatusPointer, errorCodePointer, errorSourcePointer);
+                Module.eggShell.setOccurrenceAsync(occurrencePointer);
```
rajsite: Hmm so suppose hypothetically during functionToCall.apply(context, parameters) that the user has a function like:
```javascript
function () {
   var cb = this.generateCompletionCallback();
   document.getElementById('mybutton').addEventListener('click', cb.bind(undefined, 'hello world'));
   throw new Error('woops');
}
```

So here we capture the Exception and finish the javaScriptInvoke call and wire out error terminals. But the user can click a button on the page, which will trigger the completion callback, and start writing things into Vireo memory and corrupt it.

We might need a state object that can be passed through multiple levels:

```javascript
state = {
   completionCallbackRetrieved: false, //aka the current isAsync flag
   completionCallbackInvoked: false
};
```
And that reference gets passed to the two creation functions. Ideas?

source/io/module_javaScriptInvoke.js
```diff
@@ -358,30 +372,52 @@
...
+            if (returnTypeName !== 'StaticTypeAndData') {
+                // User doesn't want return value. We're passing '*' for the return in VIA code, we get StaticTypeAndData
+                returnValuePointer = JavaScriptInvoke_GetParameterPointer(returnPointer, 0);
+            }
+            if (returnTypeName === 'Array') {
+                returnTypeName += getArrayElementTypeString(returnValuePointer);
+            }
+
+            var asyncFlag = false;
+            var generateContext = function () {
```
rajsite: I think it would be cleaner to move the generateContext function out next to the generateCompletionCallback function. With the state object discussed below that would be possible

source/io/module_javaScriptInvoke.js
```diff
@@ -358,30 +372,52 @@
...
-            try {
-                updateReturnValue(functionName, returnPointer, returnValue, errorStatusPointer, errorCodePointer, errorSourcePointer);
-            } catch (ex) {
-                newErrorStatus = true;
-                newErrorCode = ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke.CODE;
-                newErrorSource = Module.coreHelpers.formatMessageWithException(ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke.MESSAGE + '\'' + functionName + '\'.', ex);
-                Module.coreHelpers.mergeErrors(newErrorStatus, newErrorCode, newErrorSource, errorStatusPointer, errorCodePointer, errorSourcePointer);
+            if (!asyncFlag) {
+                updateReturnValue(functionName, returnTypeName, returnValuePointer, returnValue, errorStatusPointer, errorCodePointer, errorSourcePointer);
+                Module.eggShell.setOccurrenceAsync(occurrencePointer);
```
rajsite: Based on the changes we talked about in person, I think most of these calls will be able to use the synchronous setOccurence function that we were discussing

source/io/module_javaScriptInvoke.js
rajsite: If eslint gets mad at you just do:
var completionCallback = function () {
};
return completionCallback;

In ES6 the binding name to an anonymous function expression becomes the functions names

rajsite:
It doesn't look like there is any test showing occurrences resolving asynchronously out of order (which requires calls to have unique occurrences)

source/io/module_javaScriptInvoke.js
siddhukrs: Good point. Added a test for this case.

source/io/module_javaScriptInvoke.js
siddhukrs: Implemented the state object and moved the generateContext function closer to the generateCompletionCallback function.

source/io/module_javaScriptInvoke.js
siddhukrs: I don't think they can all be changed to synchronous. This one can, though.

test-it/karma/javascriptinvoke/Async.Test.js
```diff
@@ -0,0 +1,145 @@
...
+    beforeAll(function (done) {
+        fixtures.preloadAbsoluteUrls([
+            jsAsyncFunctionsUrl
+        ], done);
+    });
+
+    beforeEach(function () {
+        vireo = new Vireo();
+
+        window.NI_AsyncSquareFunction = function (inputInteger) {
```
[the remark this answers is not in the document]
siddhukrs: Implemented.

source/io/module_javaScriptInvoke.js
```diff
@@ -285,49 +285,77 @@
...
+            };
+            return completionCallback;
+        };
+
+        var generateContext = function (occurrencePointer, functionName, returnTypeName, returnValuePointer, errorStatusPointer, errorCodePointer, errorSourcePointer, currentInvocationState) {
+            var context = {};
+            context.getCompletionCallback = function () {
+                if (currentInvocationState.completionCallbackRetrieved === true) {
+                    throw new Error('The completion callback was retrieved more than once.');
```
rajsite: IT would be useful to mention the function name that resulted in the error

source/io/module_javaScriptInvoke.js
```diff
@@ -349,6 +377,7 @@
                 newErrorCode = ERRORS.kNIUnsupportedParameterTypeInJavaScriptInvoke.CODE;
                 newErrorSource = Module.coreHelpers.formatMessageWithException(ERRORS.kNIUnsupportedParameterTypeInJavaScriptInvoke.MESSAGE + '\'' + functionName + '\'.', ex);
                 Module.coreHelpers.mergeErrors(newErrorStatus, newErrorCode, newErrorSource, errorStatusPointer, errorCodePointer, errorSourcePointer);
+                Module.eggShell.setOccurrenceAsync(occurrencePointer);
```
rajsite: These error functions should setOccurence synchronously (ie this function is resolving synchronously as an error, no point waiting to do it later)

source/io/module_javaScriptInvoke.js
```diff
@@ -358,30 +387,43 @@
...
 
-            try {
-                updateReturnValue(functionName, returnPointer, returnValue, errorStatusPointer, errorCodePointer, errorSourcePointer);
-            } catch (ex) {
-                newErrorStatus = true;
-                newErrorCode = ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke.CODE;
-                newErrorSource = Module.coreHelpers.formatMessageWithException(ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke.MESSAGE + '\'' + functionName + '\'.', ex);
-                Module.coreHelpers.mergeErrors(newErrorStatus, newErrorCode, newErrorSource, errorStatusPointer, errorCodePointer, errorSourcePointer);
+            if (!currentInvocationState.completionCallbackRetrieved) {
+                updateReturnValue(functionName, returnTypeName, returnValuePointer, returnValue, errorStatusPointer, errorCodePointer, errorSourcePointer);
```
rajsite: Hmm the user could hold onto a reference to `this` and then invoke getCompletionCallback asynchronously, ie:

```javascript
(function () {
var savedContext;

var doEvil = function () {
   var completionCallback = savedContext.getCompletionCallback();
};

window.evilSaveContext = function () {
savedContext = this;
setTimeout(doEvil,0);
};
}());

```

I think the guard for this is to defensively set
currentInvoactionState.completionCallbackRetrieved = true

and maybe the completionCallbackState as well

and a test :)

test-it/karma/javascriptinvoke/Async.Test.js
```diff
@@ -0,0 +1,188 @@
...
+
+    var vireo;
+
+    var jsAsyncFunctionsUrl = fixtures.convertToAbsoluteFromFixturesDir('javascriptinvoke/AsyncFunctions.via');
+    var javaScriptInvokeFixtures = {
+        NI_AsyncSquareFunction: function (inputInteger) {
+            var completionCallback = this.getCompletionCallback();
+            setTimeout(function () {
+                completionCallback(inputInteger * inputInteger);
+            }, 4);
```
rajsite: Like above it is better to declare your intent. You want this to run an asynchronous task as quickly as possible, ie setTimeout(fn, 0);

Today the browsers internally use a minimum of 4ms, but maybe tomorrow it will drop and our tests will run quicker.

rajsite: ommitting it is fine as well like below, I would say just be consistent https://github.com/ni/VireoSDK/search?l=JavaScript&q=setTimeout&type=

source/io/module_javaScriptInvoke.js
```diff
@@ -358,30 +387,43 @@
...
+                // User doesn't want return value. We're passing '*' for the return in VIA code, we get StaticTypeAndData
+                returnValuePointer = JavaScriptInvoke_GetParameterPointer(returnPointer, 0);
+            }
+            if (returnTypeName === 'Array') {
+                returnTypeName += getArrayElementTypeString(returnValuePointer);
+            }
+
+            var currentInvocationState = {
+                completionCallbackRetrieved: false,
+                completionCallbackResolved: 'unresolved'
```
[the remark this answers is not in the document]
rajsite: Also if we are going with the Promise terminology then we can use Pending, Fulfilled, Rejected: https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Promise.jsm/Promise

source/io/module_javaScriptInvoke.js
```diff
@@ -358,30 +401,46 @@
...
-            try {
-                updateReturnValue(functionName, returnPointer, returnValue, errorStatusPointer, errorCodePointer, errorSourcePointer);
-            } catch (ex) {
-                newErrorStatus = true;
-                newErrorCode = ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke.CODE;
-                newErrorSource = Module.coreHelpers.formatMessageWithException(ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke.MESSAGE + '\'' + functionName + '\'.', ex);
-                Module.coreHelpers.mergeErrors(newErrorStatus, newErrorCode, newErrorSource, errorStatusPointer, errorCodePointer, errorSourcePointer);
+            // assume synchronous invocation since the completion callback was never retrieved
+            if (completionCallbackStatus.retrievalState === completionCallbackRetrievalEnum.AVAILABLE) {
+                updateReturnValue(functionName, returnTypeName, returnValuePointer, returnValue, errorStatusPointer, errorCodePointer, errorSourcePointer);
```
rajsite: Why was the try - catch around updateReturnValue removed?
@cglzaguilar are there known cases that can cause updateReturnValue to throw? Are they covered in tests?

If y'all think we should keep the exception handling then maybe we can do:

```
tryUpdateReturnValue = function () {

try {
updateReturnValue
} catch() {
update lv error
}
```

So we do not have to duplicate the exception logic in two places

rajsite:
The exception handling looks good. I had a couple of concerns in the JS tests that I wanted to hammer out.

I made a commit that shows them: https://github.com/rajsite/VireoSDK/commit/512d7ef6d2ae5b91d51bc4a2bd68da19263982b6

And some details:
The goal of the javaScriptInvokeFixtures object was to help reliably add to the global scope and clean-up. Keeping the global scope clean is particularly important as it is shared between all running tests.

The problem was that now some other none function things (the `running` property and some references to the context) were being added to javaScriptInvokeFixtures and could end up in the global scope by accident.

To help emphasize the intent I show calling Object.freeze on javaScriptInvokeFixtures and move those variables out to an outer, non-global, scope where they can be accessed in the test.

After those minor issues are addressed I'm good to go 👍

rajsite:
@siddhukrs Everything looks great, let me know if it passed LVjs CI and then I'll merge and release

siddhukrs:
Passing LVjs CI:  http://lvngtestexperts/Builds/Report?definition=R4-Bajor-Main%5BLVjs%5D&number=20180503.6

PR #423, opened 2018-04-23
Fix InRangeAndCoerce NaN bug

mehagar:
We were not handling the case where the inputs were NaN. The correct output should be NaN in that case.


source/core/Math.cpp
```diff
@@ -350,7 +350,9 @@ DECLARE_SCALE2X_INTN_HELPER(Single)
          /* Args:  x loLimit hiLimit includeLo includeHi coercedOut  inRangeOut */\
         VIVM_TRACE_FUNCTION(InRangeAndCoerce##TYPE)    \
         if (_ParamPointer(5)) { \
-            if (_Param(1) > _Param(2)) { \
+            if (::isnan((double)_Param(0)) || ::isnan((double)_Param(1)) || ::isnan((double)_Param(2))) { \
```
sanmut: Is the "In Range" output argument affected if x or hiLimit or LoLimit is nan? Should we check for isnan there too? 

Would be worth adding checks for InRange out in the tests if not already done. 

Also, do we have tests for this node for both scalar and vector and Compare Elements and Compare Aggregates modes?

sanmut:
BTW, I think there is a merge commit that can be removed.

rajsite:
@spathiwa Have you had a chance to review this?

source/core/Math.cpp
spathiwa: _Param(6) should be set to False if any of the inputs are NaN, but it looks like the code already there that assigned _Param(6) would do this, because all the comparisons (>, ==) would return false.  And his tests verifies that.

spathiwa:
@rajsite Sorry.  Just did.

source/core/Math.cpp
mehagar: Also, should I even add these tests to their own test file? Or should I add them into InRangeAndCoerce.via as well?

sanmut: I meant coverage for scalars/vectors and Elements/Aggregates in general. If those tests are in the GVI test suite, that is sufficient. No need to add to .via file.

test-it/ViaTests/InRangeAndCoerceNaN.via
```diff
@@ -47,25 +43,13 @@ define (%3A%3AWeb%20Server%3A%3AInteractive%3A%3AWebApp%3A%3AInRangeAndCoerceNaN
         clump(1
         InRangeAndCoerce(c2 c1 c0 local34 local33 local10 local9)
         Println(local10)
-        IsNotANumPathRefnum(local10 local11 )
```
sanmut: Checking for a NaN with IsNotANumPathRefum primitive is still valid and need not be removed, I think. dataItem_Pass at the end could be printed and checked for.

mehagar: My reasoning was that printing out the values of the variables (that they equal "NaN") was equivalent to what the IsNotANumPathRefnum was doing, and checking the dataItem_Pass at the end provides no additional value over the test as a whole, since if dataItem_Pass was false it would imply that one of the variables was not what I was expecting, meaning that the results printed were different than in the .vtr file. Does that make sense?

PR #425, opened 2018-04-26
Update for PropertyNode module API

gleonoliva:
After doing some testing with globals I noticed that passing only the path returned by the reflection API did not contain enough information to know whether a path pointed to a global or to a local in a VI.

So first I added another method to the reflection API to let the caller know if the variable that is looking for was found in a VI, also I used this opportunity to encode the VI name of the path returned. This way is possible to split the full path and get the VI name as well as the rest of the path, which is what the API for eggshell on the JS side needs.

In order to keep our API consistent I also encode the VI name where the Control reference lives, this way the local update service can immediately call back to vireo without any further modification of the parameters passed from `module_propertyNode.js`.

Also added more test cases with encoded VI names in .via files.

Here is how calls for propertyRead look after the change:
```javascript
propertyRead('Main%2Egviweb', // [Now encoded] VI name of ControlRef
             'dataItem_Boolean', // ControlRef data item id 
             'Value', // Property name
             'Boolean', // Property type
             'SubVI%2Egviweb', // [NEW!] Encoded VI name for property value, will be '' if global. 
             'subVIbooleanLocal'); // Rest of the path for storing the value read.
```


source/io/PropertyNode.cpp
```diff
@@ -60,10 +60,45 @@ static bool LookupControlRefForPropertyNode(RefNumVal *refNumPtr, ErrorCluster *
...
+{
+    TypeManagerRef typeManager = value->_paramType->TheTypeManager();
+
+    STACK_VAR(String, pathRef);
+    StringRef symbolPath = pathRef.Value;
+    Boolean foundInVI = false;
+    typeManager->PointerToSymbolPath(value->_paramType, value->_pData, symbolPath, &foundInVI);
+
+    SubString symbolPathSubStr = symbolPath->MakeSubStringAlias();
+    if (symbolPathSubStr.CompareCStr("*pointer-not-found*")) {
```
gleonoliva: Wondering if instead we should modify `PointerToSymbolPath` to return a boolean instead of always returning `null`

Vireo_VS/VireoCommandLine.vcxproj
```diff
@@ -165,4 +166,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
rajsite: Add newlines to the end of the files to prevent the red arrows

source/include/StringUtilities.h
```diff
@@ -424,6 +427,42 @@ class DecodedSubString {
...
+    explicit EncodedSubString(const SubString &s, bool encode = true, bool alwaysAlloc = false);
+
+    void Init(const SubString &s, bool encode, bool alwaysAlloc = false);
+    SubString GetSubString() {
+        SubString s;
+        s.AliasAssignLen(_encodedStr, IntIndex(strlen(ConstCStr(_encodedStr))));
+        return s;
+    }
+    Boolean NeedsEncoding(Utf8Char c) {
+        return !SubString::IsNumberChar(c) && !SubString::IsLetterChar(c) && (c != '+' || c != '*' || c != '_' || c != '-' || c != '$');
```
rajsite: Is this based on the JS side? Did we find a spot in Vireo C++ where it's clear which characters need to be encoded for identifiers?

Vireo_VS/VireoCommandLine.vcxproj
sanmut: may want to fix end of file line-ending

source/core/StringUtilities.cpp
```diff
@@ -1132,6 +1132,64 @@ DecodedSubString::DecodedSubString(const SubString &s, bool decode, bool alwaysA
...
+    } else {
+        memcpy(_encodedStr, s.Begin(), len);
+    }
+    _encodedStr[len] = 0;
+}
+
+EncodedSubString::EncodedSubString(const SubString &s, bool encode, bool alwaysAlloc) : _encodedStr(null) {
+    Init(s, encode, alwaysAlloc);
+}
+
```
sanmut: I think we should have encoding / decoding unit tests that should be done in all the different layers/languages. Please track this if you are not planning to address it now.

source/include/StringUtilities.h
```diff
@@ -424,6 +427,42 @@ class DecodedSubString {
     Utf8Char *_decodedStr;
 };
 
+//------------------------------------------------------------
+//! A class for making temporary %-code encoded SubStrings
```
cglzaguilar: Is this %-code encoding something unique to us or is it something that is more common in the industry? If the answer is the industry, then we should use the name used in the industry to describe this algorithm and have more test cases to ensure we are properly implementing it.

source/include/DataReflectionVisitor.h
```diff
@@ -0,0 +1,69 @@
+/**
+Copyright (c) 2015 National Instruments Corp.
+This software is subject to the terms described in the LICENSE.TXT file
+SDG
+*/
+
+/*! \file
```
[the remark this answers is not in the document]
cglzaguilar: Thank you for putting the definition behind a .h file.

source/core/StringUtilities.cpp
```diff
@@ -1132,6 +1132,64 @@ DecodedSubString::DecodedSubString(const SubString &s, bool decode, bool alwaysA
...
+        Utf8Char c;
+        Int32 value;
+        Utf8Char *pDest = _encodedStr;
+
+        if (ss.ReadRawChar(&c)) {
+            if (SubString::IsLetterChar(c)) {
+                *pDest++ = c;
+            } else {
+                *pDest++ = '%';
+                *pDest++ = (Utf8Char)(HEX_VALUES[(0xF0 & c) >> 4]);
```
cglzaguilar: If we are not concerned about performance, consider adding a utility function to make the code clearer. Something along GetHexadecimalRepresentationForUtf8Char, that will take pDest and c as parameters and basically do these 3 lines of code.

source/core/TypeAndDataReflection.cpp
```diff
@@ -11,59 +11,13 @@ SDG
...
-    virtual void VisitPointer(PointerType* type);
-    virtual void VisitEnum(EnumType* type);
-    virtual void VisitRefNumVal(RefNumValType* type);
-    virtual void VisitDefaultValue(DefaultValueType* type);
-    virtual void VisitDefaultPointer(DefaultPointerType* type);
-    virtual void VisitCustomDataProc(CustomDataProcType* type);
-};
 //------------------------------------------------------------
-TypeRef TypeManager::PointerToSymbolPath(TypeRef tNeedle, DataPointer pNeedle, StringRef path)
+TypeRef TypeManager::PointerToSymbolPath(TypeRef tNeedle, DataPointer pNeedle, StringRef path, Boolean *foundInVI)
```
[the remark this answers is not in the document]
gleonoliva: I'll do

source/include/DataReflectionVisitor.h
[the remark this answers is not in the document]
gleonoliva: No problem, it was asked in the previous PR :)
I'll add docs.

source/core/StringUtilities.cpp
```diff
@@ -1091,45 +1091,118 @@ IntIndex SubString::FindFirstMatch(SubString* searchString, IntIndex offset, Boo
...
+void PercentEncodedSubString::Encode(const SubString &s) {
+    SubString ss = s;
+    Utf8Char c;
+    Int32 value;
+    Utf8Char *pDest = _convertedStr;
+
+    // Custom encoder percent-encodes if first char
+    // is not a letter [A-Za-z]
+    if (ss.ReadRawChar(&c)) {
+        if (SubString::IsLetterChar(c)) {
```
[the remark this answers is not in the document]
gleonoliva: That's the thing. The algorithm on the C# that generates identifiers only encodes the first character if is not a letter. IMO is a little weird but I didn't want to change it and risk breaking stuff without sufficient test coverage.

PR #426, opened 2018-04-26
ReadJSON should put 64 bit numbers in quotes

mehagar:
This also involved making sure WriteJSON handled quoted 64 bit numbers.

### DIFF
--- a/source/core/TDCodecVia.cpp
+++ b/source/core/TDCodecVia.cpp
@@ -59,7 +59,7 @@ TDViaParser::TDViaParser(TypeManagerRef typeManager, SubString *typeString, Even
     } else if (format->ComparePrefixCStr(TDViaFormatter::formatJSON._name)) {
         _options._bEscapeStrings = true;
         _options._fmt = jsonLVExt ? (quoteInfNaN ? TDViaFormatter::formatJSONEggShell : TDViaFormatter::formatJSONLVExt) : TDViaFormatter::formatJSON;
-		_options._bQuote64BitNumbers = jsonLVExt;
+        _options._bQuote64BitNumbers = jsonLVExt;
         if (strictJSON)
            _options._fmt._fieldNameFormat = ViaFormat(_options._fmt._fieldNameFormat | kViaFormat_JSONStrictValidation);
     } else if (format->ComparePrefixCStr(TDViaFormatter::formatC._name)) {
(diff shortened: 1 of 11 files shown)


source/core/TDCodecVia.cpp
```diff
@@ -1058,12 +1059,17 @@ Int32 TDViaParser::ParseData(TypeRef type, void* pData)
         case kEncoding_UInt:
         case kEncoding_S2CInt:
             {
+                Boolean is64Bit = type->IsA(&TypeCommon::TypeInt64) || type->IsA(&TypeCommon::TypeUInt64);
                 IntMax value = 0;
                 Boolean overflow = false;
                 Utf8Char sign = 0;
                 _string.EatWhiteSpaces();
+                if (_options._bQuote64BitNumbers && is64Bit) {
+                    _string.EatChar('"');
```
rajsite: When reading JSON the parser should take a JSON number and turn it to the Vireo numeric or a JSON string that represents a number.

Make sure for 64 Int and Unint coming from JSON it can take both the json:

```json
18446744073709551615
```

and 

```json
"18446744073709551615"
```

rajsite: You may need to save the result of EatChar to use below

source/core/TDCodecVia.cpp
```diff
@@ -2258,7 +2278,8 @@ void TDViaFormatter::FormatData(TypeRef type, void *pData)
         case kEncoding_DimInt:
             {
                 IntMax intValue = ReadIntFromMemory(type, pData);
-                FormatInt(type->BitEncoding(), intValue);
+                Boolean is64Bit = type->IsA(&TypeCommon::TypeInt64) || type->IsA(&TypeCommon::TypeUInt64);
```
rajsite: Maybe we should add isInteger64 as a helper in https://github.com/ni/VireoSDK/blob/master/source/core/TypeAndDataManager.cpp#L869

We now have several places we check for a 64bit integer. The two places in this pull request and here: https://github.com/ni/VireoSDK/blob/master/source/io/JavaScriptInvoke.cpp#L94

test-it/karma/publicapi/WriteJson.Test.js
```diff
@@ -82,10 +83,9 @@ describe('The Vireo EggShell writeJSON api can write', function () {
         });
 
         it('Int64', function () {
```
rajsite: We should add some tests for writing 64 bit numbers as both JSON numbers and JSON strings.

source/core/TDCodecVia.cpp
```diff
@@ -1059,13 +1059,14 @@ Int32 TDViaParser::ParseData(TypeRef type, void* pData)
         case kEncoding_UInt:
         case kEncoding_S2CInt:
             {
-                Boolean is64Bit = type->IsA(&TypeCommon::TypeInt64) || type->IsA(&TypeCommon::TypeUInt64);
+                Boolean is64Bit = type->IsInteger64();
                 IntMax value = 0;
                 Boolean overflow = false;
                 Utf8Char sign = 0;
                 _string.EatWhiteSpaces();
+                Boolean ateBeginningQuote = false;
```
rajsite: Maybe we can choose a name that better matches the intent like `isNumericString`

mehagar:
The LVjs failed with new failures - please do not merge this until I confirm what is happening.

mehagar:
OK, I confirmed that the failures will be fixed when I updated niVireoPeeker.js in TFS to handle the now quoted int64s. Please merge.

PR #428, opened 2018-05-02
Fix bug in Event re-registration and add test case for it.

spathiwa:


sanmut:
We should look at the Queue failure. Keeps showing up. Do we have a github issue or CAR to track this failure?

PR #429, opened 2018-05-04
Delete test-it/x folder

mehagar:
These files were not being used anywhere and there are no corresponding expected results.


sanmut:
Some files seem to have useful comments. We should just rename the folder name "x" to "UnusedInTests" or something to that effect so that it is obvious these files are not used by any automated tests. Or, move the tests/comments to other tests (or files themselves directly) under ViaTests folder and delete only duplicate/trivial ones.

mehagar:
@sanmut , I understand your concern about losing valuable information that may be useful later on. Since we are using source control we will always be able to go back and look at these comments should we need to. Could you read http://www.informit.com/articles/article.aspx?p=1334908 and tell me if you agree with the author's sentiment regarding commented out (unused) code?

rajsite:
Sounds like the general consensus is to not delete the x folder wholesale but try to filter through and enable valuable tests.

Closing the PR for now, feel free to re-open if re-visiting.

PR #430, opened 2018-05-14
 Add polymorphic input support to Register for Events.

spathiwa:


spathiwa:
(The -std=c++11 addition to EmMakefile is to allow an anonymous union in a class which is a C++ 11 feature; it seemed useful to not waste storage between the scalar and non-scalar cases.)

make-it/EmMakefile
```diff
@@ -87,7 +87,7 @@ ifeq ($(TARGET), wasm32-unknown-emscripten)
 endif
 
 # NO_DYNAMIC_EXECUTION=1 Allows Vireo to be used in CSP environments where eval() is forbidden: https://github.com/kripken/emscripten/blob/master/src/settings.js#L563
-EM_OPT = $(EM_OPTFLAGS) $(WASM_FLAG) -s NO_EXIT_RUNTIME=1 -fno-exceptions --memory-init-file 0 -s NO_DYNAMIC_EXECUTION=1
+EM_OPT = $(EM_OPTFLAGS) $(WASM_FLAG) -s NO_EXIT_RUNTIME=1 -std=c++11 -fno-exceptions --memory-init-file 0 -s NO_DYNAMIC_EXECUTION=1
```
sanmut: Does the emscripten that we use support C++14 or even C++17?

spathiwa: Release vireo.js is only 3271 bytes bigger with this commit; that's probably all due to the size of the new event code.  It doesn't look like -std=c++11 had any material effect on code size.
I'm not sure the best way to test for performance behavior change but I didn't notice obvious difference in the time to run unit tests.

spathiwa: @sanmut Yes, it should support C++14 and C++17.    The FAQ says it derives this support completely from the version of clang and libc++ it uses, and so it should support C++17.
Rather, emcc 1.37.9 (clang 4.0.0) should have partial C++17 support and 1.37.36 (clang 5.0) full support.  
https://github.com/kripken/emscripten/issues/5535
https://clang.llvm.org/cxx_status.html

sanmut: I think we are using 1.37.36. Why not switch to using C++17 directly? We should avoid partial support of a standard (as some drafts get pulled out in the final standard). However, if we are switching to use modern C++, then I favor switching to the latest C++ standard supported unless we have specific reasons not to. Thoughts?

rajsite: As long as we choose a C++ standard supported by all the environments on the CIs:

g++-4.9: https://github.com/ni/VireoSDK/blob/master/.travis.yml#L12
emscripten 1.37.36: https://github.com/ni/VireoSDK/blob/master/.travis.yml#L38
Visual Studio 2017: https://github.com/ni/VireoSDK/blob/master/.appveyor.yml#L4

spathiwa: I think I'd prefer a compromise of C++14.  C++17 seems too new to me, is only recently supported by emcc, and might be more prone to vary in support between compilers. I've been burned in the past by trying to adopt new language features before they're fully supported by compilers.  Perhaps the vendors have gotten better about it, but I'd prefer to err on the side of caution.

sanmut: Agreed. We can start with C++14 now.

spathiwa: Changed.

rajsite: So I looked at the build output and it seems like gcc 4.8 is what is actually running in Travis: https://travis-ci.org/ni/VireoSDK/jobs/379930935#L713

And GCC 5 is where more complete support for C++ 14 is available: https://gcc.gnu.org/projects/cxx-status.html

spathiwa: True...  I'm okay with going back to just C++11 as originally submitted if this concerns you.
Unless we just wanna trust the CI to tell us if we try to use a C++14 feature that isn't supported by gcc-4.8.    Not sure that's a great idea, but I don't have a strong opinion on it.

rajsite: It looks like there is a path to newer GCC versions that does not require sudo builds (I've put some effort in keeping builds fast as container builds by avoiding sudo): https://stackoverflow.com/a/49412522/338923

It also looks like VS 2017 is pretty well covered on C++ 14 features: https://docs.microsoft.com/en-us/cpp/visual-cpp-language-conformance

So I don't have a strong opinion either. Just the first person who uses a C++14 feature GCC on Travis doesn't support will have to take the effort to upgrade the GCC version.

source/core/Events.cpp
```diff
@@ -376,21 +390,71 @@ class DynamicEventRegInfo {
...
+    // If the registered data is a cluster or nested cluster updated *dynIndexBase with the nested element
+    // index as it traverses, and returns a non-zero index if a match occcurs.  (The dynIndex value is
+    // the same one storied in the EventSpec in the event structure configuration data.)
+    Int32 DynamicEventMatchCore(TypeRef regRefType, void *pData, RefNum refnum, Int32 *dynIndexBase) {
         Int32 dynIndex = 0;
-        for (size_t i = 0; i < _entry.size(); i++) {
+        if (regRefType->IsRefnum() && pData && ((RefNumVal*)pData)->GetRefNum() == refnum) {
+            dynIndex = *dynIndexBase;
+            return dynIndex;
+        } else if (regRefType->IsArray() && regRefType->Rank() == 1 && regRefType->GetSubElement(0)->IsRefnum()) {
```
cglzaguilar: Do you know if the VI is broken in NG when the user attempts to connect an N-Dimensional array?

source/core/Events.cpp
```diff
@@ -586,6 +650,60 @@ static inline InstructionCore *ReturnRegForEventsFatalError(const char *errorMes
     return THREAD_EXEC()->Stop();
 }
 
+// Recursive helper function to handle event registration (or re-registration) of polymorphic inputs (clusters/arrays of refnums)
+LVError RegisterForEventsCore(EventQueueID qID, DynamicEventRegInfo *regInfo, Int32 refInput, TypeRef refType, EventSource eSource,
```
cglzaguilar: This function appears to be a duplicate of DynamicEventMatchCore. Is it possible to share code?

rajsite:
Realized that using the "comment" review status does not fulfill the code owner review status, have to either approve or reject

source/core/Events.cpp
spathiwa: I thought about this when I wrote them. Their recursive structure is similar, but they have very different arguments and do different things in the base case; I couldn't find a clean way to share code.

PR #432, opened 2018-05-18
Support ByteArrayToString and StringToByteArray.

sanmut:
Support only UTF-8 encoding.
Use UTF8-cpp from https://github.com/ledger/utfcpp to
validate byte arrays and strings


Vireo_VS/VireoCommandLine.vcxproj
```diff
@@ -170,4 +174,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
cglzaguilar: Please add another line at the end to remove the "red character"

Vireo_VS/VireoCommandLine.vcxproj.filters
```diff
@@ -247,5 +250,14 @@
...
+    </ClInclude>
+    <ClInclude Include="..\source\include\utf8\unchecked.h">
+      <Filter>VireoSource\Include\utf8</Filter>
+    </ClInclude>
+    <ClInclude Include="..\source\include\utf8\utf8.h">
+      <Filter>VireoSource\Include\utf8</Filter>
+    </ClInclude>
   </ItemGroup>
-</Project>
+</Project>
```
cglzaguilar: Please add another line at the end to remove the "red character"

source/core/NumericString.cpp
```diff
@@ -2715,6 +2807,8 @@ DEFINE_VIREO_BEGIN(NumericString)
     DEFINE_VIREO_FUNCTION(StringFormat, "p(i(VarArgCount) o(String)   i(String) io(ErrorCluster err) i(StaticTypeAndData))")
     DEFINE_VIREO_FUNCTION(StringScanValue, "p(i(String) o(String) i(String) o(StaticTypeAndData))")
     DEFINE_VIREO_FUNCTION(StringScan, "p(i(VarArgCount) i(String) o(String) i(String) i(UInt32) o(UInt32) io(ErrorCluster err) o(StaticTypeAndData))")
+    DEFINE_VIREO_FUNCTION(ByteArrayToString, "p(o(String) i(Array) i(Int32) io(ErrorCluster err))")
```
cglzaguilar: For both functions how about inputs first and outputs last?

spathiwa: I agree, that would be better.  I wish we were consistent everywhere, but it's generally the rule that the output is first when there are a variable number of inputs, and not otherwise.

rajsite:
1. Can we switch to using a release from the original author (it looks like the github repo chosen is just a random person uploading the library):
https://sourceforge.net/projects/utfcpp/files/utf8cpp_2x/

2. Can we place it in a location where it is clear that it is 3rd party source:
   `/source/include/thirdparty/utf8/` and if it had cpp files `/source/thirdparty/utf8/`

3. Can you add a README.md file in `/source/include/thirdparty/utf8/` that describes where the 3rd party dependency came from (helpful in-case we need to upstream a bug fix in the future)

4. The dependency does not seem very large, but can you confirm that it does not make vireo.js grow in size unexpectedly

source/core/NumericString.cpp
```diff
@@ -1953,6 +1954,97 @@ VIREO_FUNCTION_SIGNATUREV(StringScan, StringScanParamBlock)
...
+struct ByteArrayToStringParamBlock : InstructionCore
+{
+    _ParamDef(StringRef, StringOut);
+    _ParamDef(TypedArrayCoreRef, ByteArrayIn);
+    _ParamDef(Int32, StringEncodingIn);  // 0 = UTF-8 (only supported)
+    _ParamDef(ErrorCluster, ErrorInOut);
+    NEXT_INSTRUCTION_METHOD()
+};
+
+static bool IsSupportedEncodingAndUpdateErrorAndMakeOutputArrayEmpty(ErrorCluster* errorCluster, Int32 stringEncoding, TypedArrayCoreRef outputArray) {
```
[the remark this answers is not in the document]
mehagar: Or break this up into two functions - one that does the check and one that updates the error and output array.

source/core/NumericString.cpp
```diff
@@ -1953,6 +1954,97 @@ VIREO_FUNCTION_SIGNATUREV(StringScan, StringScanParamBlock)
...
+    if ((stringOut == nullptr && errorCluster == nullptr) || (errorCluster != nullptr && errorCluster->hasError()))
+        return _NextInstruction();
+    TypedArrayCoreRef byteArray = _Param(ByteArrayIn);
+    TypeRef eltType = byteArray->ElementType();
+    VIREO_ASSERT(eltType->IsA(&TypeCommon::TypeInt8) || eltType->IsA(&TypeCommon::TypeUInt8));
+    IntIndex rank = byteArray->Rank();
+    VIREO_ASSERT(rank == 1);
+    Int32 stringEncoding = _Param(StringEncodingIn);
+    Int32 arrayLength = byteArray->DimensionLengths()[0];
+    InstructionCore* value;
```
mehagar: I don't see where this variable is used?

source/core/NumericString.cpp
```diff
@@ -1953,6 +1954,97 @@ VIREO_FUNCTION_SIGNATUREV(StringScan, StringScanParamBlock)
...
+    _ParamDef(StringRef, StringOut);
+    _ParamDef(TypedArrayCoreRef, ByteArrayIn);
+    _ParamDef(Int32, StringEncodingIn);  // 0 = UTF-8 (only supported)
+    _ParamDef(ErrorCluster, ErrorInOut);
+    NEXT_INSTRUCTION_METHOD()
+};
+
+static bool IsSupportedEncodingAndUpdateErrorAndMakeOutputArrayEmpty(ErrorCluster* errorCluster, Int32 stringEncoding, TypedArrayCoreRef outputArray) {
+    if (stringEncoding != 0) {
+        if (errorCluster != nullptr) {
```
mehagar: I remember seeing somewhere in the code that we should use null instead of of nullptr.

source/core/NumericString.cpp
```diff
@@ -1953,6 +1954,97 @@ VIREO_FUNCTION_SIGNATUREV(StringScan, StringScanParamBlock)
...
+    _ParamDef(ErrorCluster, ErrorInOut);
+    NEXT_INSTRUCTION_METHOD()
+};
+
+VIREO_FUNCTION_SIGNATURET(StringToByteArray, StringToByteArrayParamBlock)
+{
+    StringRef stringIn = _Param(StringIn);
+    TypedArrayCoreRef byteArrayOut = _ParamPointer(ByteArrayOut) ? _Param(ByteArrayOut) : null;
+    ErrorCluster *errorCluster = _ParamPointer(ErrorInOut);
+    if (byteArrayOut == nullptr || (errorCluster != nullptr && errorCluster->hasError()))
```
mehagar: In the other function you are checking if the output array is null AND the error cluster is null. Please do the same here.

source/core/NumericString.cpp
```diff
@@ -1953,6 +1954,97 @@ VIREO_FUNCTION_SIGNATUREV(StringScan, StringScanParamBlock)
...
+    IntIndex rank = byteArrayOut->Rank();
+    VIREO_ASSERT(rank == 1);
+    Int32 stringEncoding = _Param(StringEncodingIn);
+    if (!IsSupportedEncodingAndUpdateErrorAndMakeOutputArrayEmpty(errorCluster, stringEncoding, byteArrayOut)) {
+        return _NextInstruction();
+    }
+    if (!IsValidUTF8AndUpdateErrorAndMakeOutputArrayEmpty(stringIn->RawBegin(), stringIn->RawBegin() + stringIn->Length(), errorCluster, byteArrayOut)) {
+        return _NextInstruction();
+    }
+    Int32 arrayLength = stringIn->Length();
```
mehagar: You computed stringIn->Length on line 2040 - you could define this variable before that and use it there.

source/core/NumericString.cpp
spathiwa: No, this is correct here. These are the conditions where we should do nothing and return early.

mehagar: Just to be clear, I was asserting that the check should be (byteArrayOut == nullptr && errorCluster == nullptr) || (errorCluster != nullptr && errorCluster->hasError()).

test-it/ViaTests/ByteArrayString.via
```diff
@@ -0,0 +1,59 @@
...
+
+        Copy(noError err)
+        ByteArrayToString(stringOut babad_in3 0 err)
+        Printf("3: array = <%z>, string = <%s>, err = <%z>\n" babad_in3 stringOut err)
+
+        Copy(noError err)
+        ByteArrayToString(stringOut babad_in4 0 err)
+        Printf("4: array = <%z>, string = <%s>, err = <%z>\n" babad_in4 stringOut err)
+
+        Copy("" stringOut)
```
mehagar: Is this line necessary? It's not used in the previous calls.

test-it/ViaTests/ByteArrayString.via
```diff
@@ -0,0 +1,59 @@
...
+        e(dv(a(.UInt8 *) (245)) babad_in3)
+        e(dv(a(.UInt8 *) (246)) babad_in4)
+        e(dv(a(.UInt8 *) (255)) babad_in5)
+        e(dv(.String) s1)
+        e(dv(.String) stringOut)
+        e(dv(.Boolean) isSame)
+    )
+    clump (
+        Copy(noError err)
+        ByteArrayToString(stringOut babad_in1 0 err)
```
mehagar: You're manually specifying 0 as the encoding, and at the bottom of this test you're explicitly using the "encoding" variable. Can you be consistent here?

test-it/ViaTests/ByteArrayString.via
```diff
@@ -0,0 +1,59 @@
+//define (true -43 'shut up about moon men')
```
mehagar: Delete this commented out line.

test-it/ViaTests/UTF-8HelloWorld.via
```diff
@@ -6,24 +6,38 @@
...
+    // to see if the String Length looks correct, but not always.
     // some fonts will use two spaces per character or fractional amounts
-    
+
     Printf("------------------------------------123456789.123456789.123456789.123456789-.\n")
     Printf("ArrayLength (%2d) StringLength (%2z) -%z-\n" aLen sLen string)
+    StringToByteArray(byteArray string 0 err)
+    ByteArrayToString(stringOut byteArray 0 err)
+    IsEQ(string stringOut isSame)
+    BranchIfTrue(2 isSame)
```
mehagar: Could you put a comment here saying why we aren't doing the normal thing and always printing the value of these variables. I understand why, because otherwise we'd have to add a lot more lines to the .vtr file.

test-it/ViaTests/ByteArrayString.via
spathiwa: Leave it, it’s innocuous and it made me laugh.

Vireo_VS/VireoCommandLine.vcxproj
sanmut: Since these files are generated by IDE, I think it is likely ok to ignore the end of file character related diff.

source/core/NumericString.cpp
sanmut: Will update. Looks like emission side is that way.

source/core/NumericString.cpp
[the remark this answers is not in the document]
sanmut: Will do.

source/core/NumericString.cpp
```diff
@@ -1953,6 +1954,97 @@ VIREO_FUNCTION_SIGNATUREV(StringScan, StringScanParamBlock)
...
+            errorCluster->SetError(true, -1, "Encoding type not supported", true);  // TODO(sanmut): Update error code.
+        }
+        if (outputArray != nullptr)
+            outputArray->Resize1D(0);
+        return false;
+    }
+    return true;
+}
+
+static bool IsValidUTF8AndUpdateErrorAndMakeOutputArrayEmpty(UInt8* begin, UInt8* end, ErrorCluster* errorCluster, TypedArrayCoreRef outputArray)
```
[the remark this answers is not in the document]
sanmut: Will do.

source/core/NumericString.cpp
```diff
@@ -1953,6 +1954,97 @@ VIREO_FUNCTION_SIGNATUREV(StringScan, StringScanParamBlock)
...
+
+VIREO_FUNCTION_SIGNATURET(ByteArrayToString, ByteArrayToStringParamBlock)
+{
+    StringRef stringOut = _ParamPointer(StringOut) ? _Param(StringOut) : nullptr;
+    ErrorCluster *errorCluster = _ParamPointer(ErrorInOut);
+    if ((stringOut == nullptr && errorCluster == nullptr) || (errorCluster != nullptr && errorCluster->hasError()))
+        return _NextInstruction();
+    TypedArrayCoreRef byteArray = _Param(ByteArrayIn);
+    TypeRef eltType = byteArray->ElementType();
+    VIREO_ASSERT(eltType->IsA(&TypeCommon::TypeInt8) || eltType->IsA(&TypeCommon::TypeUInt8));
```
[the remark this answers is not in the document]
sanmut: Converted these to compiler errors. However, prototypes using i(a(UInt8 *)) do not catch parse errors with VIA using other array types.

source/core/NumericString.cpp
```diff
@@ -1953,6 +1954,97 @@ VIREO_FUNCTION_SIGNATUREV(StringScan, StringScanParamBlock)
...
+        return _NextInstruction();
+    TypeRef eltType = byteArrayOut->ElementType();
+    VIREO_ASSERT(eltType->IsA(&TypeCommon::TypeUInt8));
+    IntIndex rank = byteArrayOut->Rank();
+    VIREO_ASSERT(rank == 1);
+    Int32 stringEncoding = _Param(StringEncodingIn);
+    if (!IsSupportedEncodingAndUpdateErrorAndMakeOutputArrayEmpty(errorCluster, stringEncoding, byteArrayOut)) {
+        return _NextInstruction();
+    }
+    if (!IsValidUTF8AndUpdateErrorAndMakeOutputArrayEmpty(stringIn->RawBegin(), stringIn->RawBegin() + stringIn->Length(), errorCluster, byteArrayOut)) {
```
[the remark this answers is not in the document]
sanmut: Discussed with @spathiwa.  We are choosing to do validation for StringToByteArray because a LV string might have been gotten by other means and may still contain invalid UTF-8 characters. However, we will populate the ByteArray Output regardless of finding invalid UTF-8 characters. This will allow users to be able to debug also.

sanmut:
Response to @rajsite 
1) Updating to use released version of 2.3.5 from github. Github is newer than sourceforge as you found it. Thanks.
2) Carlos and I discussed about this. Creating a folder called ThirdParty and moved utf8cpp under that.  This will help keep all the files from a third party component together and keep the relative paths of files correct and thus enable us to upgrade releases with least or no changes.
3) Adding a Readme.utf8.md under ThirdParty folder that describes about the source location. Good idea, thanks.
4) Looks like there is about ~1.6 KB increase in the generated vireo.js just due to UTF-8 (on the same machine with and without pulling utf8cpp)
5) It looks like this has some utf8 <--> utf16, though I have not played with it much. I could not find any other third party library or std C++ library that does what we want. This one looks fairly extensive with good test coverage. I tried the tests included in the library and they pass. However, it will be extra work to automate their existing tests.

source/core/NumericString.cpp
sanmut: Addressed this.

source/core/NumericString.cpp
sanmut: Good catch. Done. Thanks.

test-it/ViaTests/ByteArrayString.via
sanmut: Updated all the test cases to write default values before calling the node under test.

test-it/ViaTests/ByteArrayString.via
sanmut: Testing using constant and a variable.

test-it/ViaTests/UTF-8HelloWorld.via
sanmut: Done :+1:

test-it/ViaTests/ByteArrayString.via
spathiwa: Boo! Put it back!!

rajsite:
It looks like the commit comment is pointing to the wrong GitHub so it should either be rewritten or it should be fixed / removed on squash merge

sanmut:
oops. Good catch @rajsite . Thanks. Updated the commit comment.

source/core/NumericString.cpp
```diff
@@ -19,6 +19,7 @@ SDG
 #include "ExecutionContext.h"
 #include "StringUtilities.h"
 #include "TDCodecVia.h"
+#include "../../ThirdParty/utfcpp/source/utf8.h"
```
spathiwa: I think it would be better to add the top level directory to the compile search path, so you can just do 
#include "ThirdParty/utfcpp/source/utf8.h".

But fine with submitting as-is for now.

test-it/ViaTests/ByteArrayString.via
```diff
@@ -0,0 +1,87 @@
...
+    Locals:
+    c(
+        ce(dv(a(.UInt8 *) ()) ba0)
+        ce(dv(a(.UInt8 *) (65 66 67 97 98 99)) ba1)
+        e(dv(.UInt16 0) encoding)
+        e(dv(.String) s1)
+        ce(dv(.String "ABCabc") s2)
+        e(dv(a(.UInt8 *)) ba2)
+        e(.ErrorCluster err)
+        e(dv(c(
```
spathiwa: I think it might be good to add a few more tests cases with more representative UTF-8 strings (accented characters, non-English languages), such as those in the sample file in the utf8 source directory.

sanmut: The existing test UTF-8HelloWorld.via has UTF-8 strings and that is modified to test these two prims. We will do more testing in G CI side too.

source/core/NumericString.cpp
sanmut: Good idea. Will do that later.

PR #433, opened 2018-05-22
PropertyNode Sequential Execution fix

gleonoliva:
I created a build with these changes and ran a pre for our native tests and I didn't find anything broken.

I'm just adding a few JS tests that exposed the issue and that now all of them pass.


test-it/karma/utilities/TestHelpers.SpyHelpers.js
```diff
@@ -0,0 +1,17 @@
+(function () {
+    'use strict';
+
+    window.testHelpers = window.testHelpers || {};
```
mehagar: What does this line do?

test-it/karma/propertynode/PropertyNodeSequential.Test.js
```diff
@@ -0,0 +1,83 @@
+describe('The Vireo PropertyNode', function () {
+    'use strict';
+
+    var Vireo = window.NationalInstruments.Vireo.Vireo;
```
mehagar: How is uppercase Vireo different from lowercase vireo in this file? Can we just use one?

test-it/karma/propertynode/PropertyNodeSequential.Test.js
```diff
@@ -0,0 +1,83 @@
...
+        ], done);
+    });
+
+    beforeEach(function () {
+        vireo = new Vireo();
+        spy = jasmine.createSpy();
+    });
+
+    it('sequential read read operations invoke JavaScript callbacks with correct parameters', function (done) {
+        var viName = '%3A%3AWeb%20Server%3A%3AInteractive%3A%3AWebApp%3A%3AReadRead%2Egviweb';
```
mehagar: Could you change the vi name so its just ReadRead or similar?

test-it/karma/utilities/TestHelpers.SpyHelpers.js
gleonoliva: Basically the || operator works on truthy or falsy values. Depending on the order in which TestHelpers are loaded (or any other script) it is possible that this is the first one loaded, in that case window.testHelpers is `undefined` which is considered a falsy value. Therefore this line is saying, window.testHelpers is either an object (truthy) already defined by another script which loaded first, or is an empty object.

test-it/karma/propertynode/PropertyNodeSequential.Test.js
gleonoliva: This is just a pattern to make it easier to transition to ES6 modules which would use a `require()`. You can think of Vireo as the function constructor of a vireo object. We could, for example, only do a `vireo = new window.NationalInstruments.Vireo.Vireo()` in the  `beforeEach()` and remove this line.

test-it/karma/propertynode/PropertyNodeSequential.Test.js
gleonoliva: I didn't want to do that since the size of allocated strings mattered in order to reproduce the issue. VI name might no be related but it was difficult to reproduce once you changed the name of a control or in some cases just adding some dummy code at the beginning of the .via file.

test-it/karma/propertynode/PropertyNodeSequential.Test.js
rajsite: Also the uppercase is a reference to the class itself while the lowercase is a reference to an instance of the class inside this file. One of our linting rules enforces that classes start with uppercase names.

test-it/karma/utilities/TestHelpers.SpyHelpers.js
```diff
@@ -0,0 +1,17 @@
...
+
+    var verifySpyArgumentsForCalls = function (spy, expectedCallArgs) {
+        var i = 0;
+        var callsCount = spy.calls.count();
+        for (;i < callsCount; i += 1) {
```
rajsite: is there a reason to not place the initializer in the for-loop? like
```
var i;
for (i = 0; i < callsCount; i +=1) { ... } 
```

source/io/PropertyNode.cpp
```diff
@@ -114,8 +112,8 @@ VIREO_FUNCTION_SIGNATUREV(PropertyNodeWrite, PropertyNodeWriteParamBlock)
     STACK_VAR(String, controlRefVINameVar);
     STACK_VAR(String, controlRefDataItemIdVar);
```
cglzaguilar: Do we need to remove this variable? It appears it is not longer needed.

source/core/ControlRef.cpp
```diff
@@ -54,34 +54,42 @@ NIError ControlReferenceDestroy(ControlRefNum refnum) {
 // Clean-up Proc, run when top level VI finishes, disposing control reference.
 static void CleanUpControlReference(intptr_t arg) {
     ControlRefNum refnum = ControlRefNum(arg);
+    ControlRefInfo controlRefInfo;
```
cglzaguilar: Please add this code to ControlReferenceDestroy and update the function comment.

source/io/PropertyNode.cpp
```diff
@@ -177,8 +175,8 @@ VIREO_FUNCTION_SIGNATUREV(PropertyNodeRead, PropertyNodeReadParamBlock)
...
```
cglzaguilar: Same thing with this variable.

source/io/PropertyNode.cpp
gleonoliva: You are right, good catch, I'll do it.

rajsite:
Merging without @sanmut as he is OOO and asked to be declined from reviews.

PR #436, opened 2018-05-22
Fix JSLI function context for function invocations

rajsite:
The current implementation has the following problems:

1. Functions in global scope do not preserve the context. This means `console.log` and `performance.now` cannot be invoked without a wrapper function.
2. We are using the context to pass the Vireo API reference which has the getCompletionCallback call on it. 

   This limits us in future use cases where we would want to enable the user to do method invocation on an opaque JavaScript reference. This limits us because we need access to be able to reconfigure the context, we should not have taken the context for just the Vireo API.

This pull request makes two changes:

1. We use the context of the function as found in the global scope to allow for invocation with the correct context. As such a user does not need to write wrapper code for `console.log` and `performance.now`

2. We pass the Vireo API reference to the JavaScript function only if requested by a user. The user requests the Vireo API reference by adding an additional parameter to the JavaScript function list. For example:

   In the JSLI a user configures a function with Symbol name `myFunction` and one parameter `test` of type string.

   In JavaScript is the user writes the function signature as:
   ```
   window.myFunction = function (test) {
      console.log('test');
   }
   ```
   The function `myFunction` will be invoked with `window` as the context and with a value called test from the block diagram. They will not have access to the vireo api functionality because it was not requested.

    If instead for the same JSLI configuration (Symbol: `myFunction`, 1 string parameter: `test`) the user created the following JavaScript function:
   ```
   window.myFunction = function (test, jsAPI) {
      var cb = jsAPI.getCompletionCallback();
      setTimeout(function () {
         console.log('test delayed 1s');
         cb();
      }, 0);
   }
   ```
   Now in this case the user has explicitly requested the jsAPI object. This was done by requesting an additional parameter than what was configured by the JSLI document. Internally we know the JSLI document was configured for only one parameter. By providing a function that has two parameters the user is requesting a reference to the jsAPI object that we will provide.

   The benefits of this approach are:
1. The context is free to act with the default behavior (using the context from the path in global scope) or to be customized in the editor (using an opaque JS reference as the context for method invocation)

   In other words if we use this behavior we can prevent having an api breaking change when we implement opaque js reference method invocation in the future.

2. We can avoid the overhead of creating the JS API object reference if it is not explicitly requested by the user.

Also, this openly recognizes that relying on the number of parameters is unusual behavior. There is precedence as the jasmine api does something very similar when a user adds the done parameter to a function. They are able to detect that the done parameter is added and cause the test system to wait for the execution of the done function before continuing: https://jasmine.github.io/2.9/introduction#section-Asynchronous_Support

Despite this flaw I believe this is a net benefit:
1. We are allowing the simple use cases to function correctly without wrapper code `console.log`, etc
2. We can avoid breaking changes when opaque refnums are added
3. We improve performance by only creating the jsAPI object when requested
4. We CAN improve the editor experience by adding a checkbox to have the user specify explicitly that the last parameter is the jsAPI reference.

We can get a net improvement today by adopting the change. And we can choose to improve the user experience for the advanced use case (asynchronous resolution with the JS api) by improving the editor experience or we could choose to delay that work while still enabling advanced use cases.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -8,3 +8,4 @@ emscripten-sdk/*
 emsdk/*
 coverage/*
 objs/*
+gh-pages/*
--- a/source/io/module_javaScriptInvoke.js
+++ b/source/io/module_javaScriptInvoke.js
@@ -382,19 +382,19 @@
             return completionCallback;
         };
 
-        var generateContext = function (occurrencePointer, functionName, returnTypeName, returnValuePointer, errorStatusPointer, errorCodePointer, errorSourcePointer, completionCallbackStatus) {
-            var context = {};
-            context.getCompletionCallback = function () {
+        var generateAPI = function (occurrencePointer, functionName, returnTypeName, returnValuePointer, errorStatusPointer, errorCodePointer, errorSourcePointer, completionCallbackStatus) {
+            var api = {};
+            api.getCompletionCallback = function () {
                 if (completionCallbackStatus.retrievalState === completionCallbackRetrievalEnum.RETRIEVED) {
                     throw new Error('The completion callback was retrieved more than once for ' + functionName + '.');
                 }
                 if (completionCallbackStatus.retrievalState === completionCallbackRetrievalEnum.UNRETRIEVABLE) {
-                    throw new Error('The context being accessed for ' + functionName + ' is not valid anymore.');
+                    throw new Error('The API being accessed for ' + functionName + ' is not valid anymore.');
                 }
                 completionCallbackStatus.retrievalState = completionCallbackRetrievalEnum.RETRIEVED;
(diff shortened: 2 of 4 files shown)


rajsite:
Also link to PR showing local build (to be updated to release when merged): https://ni.visualstudio.com/AppCentral/_git/ASW/pullrequest/4668

I discussed it @cglzaguilar already, so please review soon if you can @gleonoliva @sanmut so that the async api changes will be made before for the beta. It would be undesirable for users to have to change their JS again after the beta.

.eslintignore
```diff
@@ -8,3 +8,4 @@ emscripten-sdk/*
 emsdk/*
 coverage/*
 objs/*
+gh-pages/*
```
rajsite: This is output that should be ignored when running `npm run doxygen` locally to generate docs

gleonoliva:
I'm glad that you were able to get this done, great work!

rajsite:
Removed @sanmut because he is OOO today and asked to be declined from reviews.

PR #438, opened 2018-05-24
Array max min infinities and null fix

mehagar:
We weren't correctly setting the right max/min values when NaN's were passed in. I added the logic to handle these cases, and a test for them.


source/core/Array.cpp
```diff
@@ -1121,32 +1121,58 @@ VIREO_FUNCTION_SIGNATURET(ArrayMaxMinInternal, FindArrayMaxMinInstruction)
...
         maxIndex = minIndex = -1;
+
+    Boolean haveSeenPositiveInfinity = false;
+    Boolean haveSeenNegativeInfinity = false;
     for (IntIndex i = 0; i < len; i++) {
-        Boolean less;
-        snippet->_p0 = arrayIn->BeginAt(i);
+        Boolean shouldUpdateMaxOrMin = false;
+        AQBlock1* currentElement = arrayIn->BeginAt(i);
+        Double currentElementValue = *((Double*)currentElement);
```
sanmut: Treating an element in an array always as double seems to be incorrect. Is the same instruction not getting used for array of other datatypes (Int8, Int16, Int32)? For example, when Int32/Int64 is used with data bits representing the same bits as nan, infinity, -infiinity, we could end up with incorrect results. A series of Int8/Int16 will be interpreted as double.

source/core/Array.cpp
```diff
@@ -1121,32 +1121,58 @@ VIREO_FUNCTION_SIGNATURET(ArrayMaxMinInternal, FindArrayMaxMinInstruction)
...
         snippet->_p0 = maxValue;
-        snippet->_p1 = arrayIn->BeginAt(i);
-        snippet->_p2 = &less;
+        snippet->_p1 = currentElement;
+        snippet->_p2 = &shouldUpdateMaxOrMin;
         _PROGMEM_PTR(snippet, _function)(snippet);
-        if (less) {
-           maxValue = arrayIn->BeginAt(i);
-           maxIndex = i;
+        if ((shouldUpdateMaxOrMin && isANumber) || (isPositiveInfinity && !haveSeenPositiveInfinity)) {
```
sanmut: If the logic of handling nan's and infinities in this primitive is isolated to this primitive, what happens when other comparison primitives such as "<", ">" are used with nan's and infinities? Should we not fix those as well or rather should we fix those first?

sanmut: Actually, I think we have a much bigger problem in the existing code. DECLARE_VIREO_COMPARISON_PRIMITIVES_BASE and other macros (IsLTSort, IsEqSearch) that work on all types are casting double unconditionally and suffer the same issues with the proposed implementation in this PR. Those must be fixed as well. isnan, infinity should be called only for float and double and that too handled separately for those float and double themselves as the sizes are different.

I would say we should start cleaning up from the macros (start from the basic ones such as IsLT, IsGT, IsLTSort, ..) and then try to use them as comparators to fix the higher primitives (Search/Sort/ArrayMaxMin) and have as minimal logic as possible (and no logic at all if possible) in the higher primitives.

test-it/ViaTests/ArrayMaxMin.via
```diff
@@ -270,11 +270,11 @@ define (_%54estVI_ArrayMaxMin%2Egvi dv(.VirtualInstrument (
...
-Println(g41)
-Println(local137)
-Println(local138)
-Println(local139)
-Println(local140)
+        Println(g41)
+        Println(local137)
+        Println(local138)
+        Println(local139)
+        Println(local140)
```
sanmut: We should add more tests for all comparison operators and other primitives that use comparison operators. We should come up with contrived data with different sized integer data types that will fail if they have bits representing nan/infinity/-infiniity/etc when they were treated as float/double incorrectly.

Better to write such tests in G and check in VIA here.

source/core/Array.cpp
mehagar: I understand and agree with you. This will occur for the several values of the different integer types, that when cast to double will incorrectly be interpreted as infinity or nan. Because this problem has already existed I would like to submit this as it is and create another CAR to clean up the error you mention.

test-it/ViaTests/ArrayMaxMin.via
mehagar: For the same reason as my previous comment, I would prefer to do this in a different commit.

source/core/Array.cpp
sanmut: Discussed with Michael about what all at the minimum we should do for this PR and what else should be tracked separately.

source/core/Array.cpp
spathiwa: Yeah, this is definitely wrong.  It's not just inf/nan -- the read of memory as double will get the wrong values in ALL cases when the array element type isn't actually doubles.

source/core/Array.cpp
```diff
@@ -1121,32 +1121,58 @@ VIREO_FUNCTION_SIGNATURET(ArrayMaxMinInternal, FindArrayMaxMinInstruction)
...
+    Boolean haveSeenNegativeInfinity = false;
     for (IntIndex i = 0; i < len; i++) {
-        Boolean less;
-        snippet->_p0 = arrayIn->BeginAt(i);
+        Boolean shouldUpdateMaxOrMin = false;
+        AQBlock1* currentElement = arrayIn->BeginAt(i);
+        Double currentElementValue = *((Double*)currentElement);
+        Boolean isANumber = !::isnan(currentElementValue);
+        Boolean isPositiveInfinity = ::isinf(currentElementValue) && currentElementValue > 0;
+        Boolean isNegativeInfinity = ::isinf(currentElementValue) && currentElementValue < 0;
```
spathiwa: I don't think you even need to check for +/- infinity, either.  The normal comparison will already handle these cases.

spathiwa:
I have an alternative simpler proposed fix.  Keep your test changes and the elimination of IsLTMaxMin, but instead of your changes to ArrayMaxMinInternal,  make the following change after the var declarations:

    if (len == 0) **{**
        maxIndex = minIndex = -1;
    **}   else {
        if (arrayIn->ElementType()->IsFloat()) {
            maxIndex = minIndex = -1;
            if (arrayIn->ElementType()->TopAQSize() == sizeof(Double)) {
                for (IntIndex i = 0; i < len; i++) {
                    minValue = arrayIn->BeginAt(i);
                    if (!::isnan(*(Double*)minValue)) {
                        maxIndex = minIndex = i;
                        maxValue = minValue;
                        break;
                    }
                }
            } else {
                for (IntIndex i = 0; i < len; i++) {
                    minValue = arrayIn->BeginAt(i);
                    if (!::isnan(*(Single*)minValue)) {
                        maxIndex = minIndex = i;
                        maxValue = minValue;
                        break;
                    }
                }
            }
        }
    }**
`

spathiwa:
This adds a separate preflight loop to find the first non-NaN, specifically only for float or double arrays.
Once a non-Nan min and max starting value is found, the main loop is already correct to not interfere with it.   -Inf is already less than all values and will be found by the IsLT test, same for Inf (with arguments reversed to act as >).  And NaN is never 'less' than anything, so it will never replace the original found min/max in the preflight.

spathiwa:
Oh, also:

`if (len && maxIndex != -1) {`instead of `if (len)` below in the line guarding the CopyData of  minValue and maxValue.

source/core/Array.cpp
```diff
@@ -1119,24 +1119,43 @@ VIREO_FUNCTION_SIGNATURET(ArrayMaxMinInternal, FindArrayMaxMinInstruction)
...
+                for (IntIndex i = 0; i < len; ++i) {
+                    minValue = arrayIn->BeginAt(i);
+                    if (!::isnan(*(Double*)minValue)) {
+                        maxIndex = minIndex = i;
+                        maxValue = minValue;
+                        break;
+                    }
+                }
+            }
+            else {
```
spathiwa: It actually occurs to me that since this loop will stop the first time it finds a number that's not a NaN, it might be okay to combine this into one loop instead of two, because it will usually only do one iteration.

mehagar: I'm confused. Which two loops are you referring to?

spathiwa: The separate loops for Double and Float.  It's an optimization we'd normally do to avoid the comparison each time through the loop, although in this case, the loop will almost never run so it might not be worth the code bloat.  So, this instead:

```
      if (arrayIn->ElementType()->IsFloat()) {
          maxIndex = minIndex = -1;
          bool isDouble = (arrayIn->ElementType()->TopAQSize() == sizeof(Double));
          for (IntIndex i = 0; i < len; ++i) {
              minValue = arrayIn->BeginAt(i);
              if (isDouble && !::isnan(*(Double*)minValue) || !isDouble &&  !::isnan(*(Single*)minValue) {
                  maxIndex = minIndex = i;
                  maxValue = minValue;
                  break;
              }
          }
      }
```

spathiwa: Unless it's an array full of NaN's (I would think super rare), the for loop will usually break the first iteration, and not be any more expensive than having the if outside the loop.

source/core/Array.cpp
```diff
@@ -1119,34 +1119,63 @@ VIREO_FUNCTION_SIGNATURET(ArrayMaxMinInternal, FindArrayMaxMinInstruction)
...
+
     for (IntIndex i = 0; i < len; i++) {
-        Boolean less;
-        snippet->_p0 = arrayIn->BeginAt(i);
+        Boolean shouldUpdateMaxOrMin = false;
+        AQBlock1* currentElement = arrayIn->BeginAt(i);
+        snippet->_p0 = currentElement;
         snippet->_p1 = minValue;
-        snippet->_p2 = &less;
+        snippet->_p2 = &shouldUpdateMaxOrMin;
```
sanmut: I would keep it "less" and not rename it to "shouldUpdateMaxOrMin" as it is the "less" operator. "less" does the right thing when an input is "NaN" also (The less primitive in G also does the right thing when an input is NaN).

rajsite:
@mehagar is this review still in work?

mehagar:
@rajsite , no, it's finished.

PR #439, opened 2018-05-25
Use timestamp constructor that takes double to correct precision

mehagar:
I wasn't sure how to write a test for this, because I was testing it out using GetCurrentTime, which is not reproducible via a test..


source/core/Timestamp.cpp
```diff
@@ -100,7 +100,7 @@ namespace Vireo
         if (retval == -1) {
             *t = Timestamp(0, 0);
         } else {
-            *t = Timestamp((Double)tv.tv_sec + kStdDT1970re1904, tv.tv_usec / 1E6);
+            *t = Timestamp((((tv.tv_sec * 1E6) + (kStdDT1970re1904 * 1E6)) + tv.tv_usec) / (Double)1E6);
```
sanmut: (Double)1E6 : typecast is redundant. It is already a double.

"Timestamp(tv.tv_sec + kStdDT1970re1904  + (tv.tv_usec / 1E6)) " is less verbose and sufficient.

sanmut:
@mehagar , for a test, you can call GetTimestamp in a loop 10 times and verify whether at least once the call returns a timestamp with a non-zero fractional part.

mehagar:
@sanmut , for the test you mentioned about calling it 10 times in a for loop, was that for a .gvi test?

sanmut:
@mehagar , you could write it as a .via test too.

sanmut:
Could you check if you convert a Timestamp to a double in G, if that preserves precision?

source/core/Timestamp.cpp
```diff
@@ -90,7 +90,7 @@ namespace Vireo
 
         stdTime -= 11644473600000000Ui64;    // DELTA_EPOCH_IN_MICROSECS
         stdTime /= 10;                        // Convert to microseconds
-        *t = Timestamp((Double)(stdTime/1000000UL), (stdTime % 1000000UL) / 1E6);
+        *t = Timestamp(stdTime/(Double)1000000UL);
```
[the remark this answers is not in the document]
spathiwa: Or just 1E6, like you have elsewhere.

source/core/Timestamp.cpp
spathiwa: Agree.

source/core/Timestamp.cpp
```diff
@@ -112,7 +112,7 @@ namespace Vireo
         } else {
             uInt32 tempTime = static_cast<uInt32>(ts.tv_sec);
             TToStd(&tempTime);
-            *t = Timestamp(static_cast<Double>(tempTime), ts.tv_nsec / 1E9);
+            *t = Timestamp(static_cast<Double>(tempTime) + (ts.tv_nsec / (Double)1E9));
         }
```
spathiwa: Same here, (Double) cast not needed for 1E9.

mehagar:
@sanmut , when you ask about verifying the precision being preserved, do you mean with my changes or in the existing desktop implementation?

PR #441, opened 2018-06-05
Null refactor

mehagar:


spathiwa:
Looks good.

cglzaguilar:
Is there a way we can modify our C++ linter to favor nullptr and disallow usages of NULL and null, so that in the future we don't re-introduce?

mehagar:
@cglzaguilar , I looked into it and it is not possible with cpplint to add custom rules. There is a static analysis tool called [cppcheck](http://cppcheck.sourceforge.net/) that supports custom rules, but I don't think it's worth the effort to add a rule for just this case.

rajsite:
@mehagar Please always at least include a high-level description in the pull request notes. For example, "Standardized on usage of C++11 nullptr across Vireo".

PR #440, opened 2018-06-04
Appveyor chrome fix

rajsite:
Many AppVeyor builds have been failing with errors like the following:
```
npm run test
> vireo@9.5.0 test C:\projects\vireosdk
> karma start --browsers ChromeHeadlessNoSandbox,FirefoxHeadless
01 06 2018 10:27:36.215:INFO [karma]: Karma v1.7.1 server started at http://0.0.0.0:9876/
01 06 2018 10:27:36.217:INFO [launcher]: Launching browsers ChromeHeadlessNoSandbox, FirefoxHeadless with concurrency 1
01 06 2018 10:27:36.222:INFO [launcher]: Starting browser ChromeHeadless
01 06 2018 10:28:36.225:WARN [launcher]: ChromeHeadless have not captured in 60000 ms, killing.
01 06 2018 10:28:38.225:WARN [launcher]: ChromeHeadless was not killed in 2000 ms, sending SIGKILL.
01 06 2018 10:28:40.226:WARN [launcher]: ChromeHeadless was not killed by SIGKILL in 2000 ms, continuing.
```

Some searching suggests that proxy settings being misconfigured may result in Chrome failing to run. This change disables the Headless Chrome proxy. In addition it modifies some tests so they can run in browsers currently running on the AppVeyor CI.

PR #443, opened 2018-06-05
Only run Firefox on AppVeyor

rajsite:
Currently Headless Chrome fails to launch intermittently on AppVeyor with the error shown in PR https://github.com/ni/VireoSDK/pull/440

And disabling the proxy as shown in the PR does not seem to improve the behavior reliably. Because the failure to launch results in a hung build instead of failing quickly, opting to disable HeadlessChrome until we can investigate further.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -37,7 +37,7 @@ test_script:
   - make testjs
   - make testnative
   - make testhttpbin
-  - npm run test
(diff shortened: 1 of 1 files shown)


rajsite:
When revisiting this in the future one thing to look at may be the socket.io connection timeout described here: https://github.com/karma-runner/karma/issues/2652#issuecomment-335599121

PR #444, opened 2018-06-06
[Request For Comments] JS opaque object reference with JSInvoke - first pass

ggumen:
Intended for review of direction, not for submission.

Implementation of support of Opaque Javascript Reference in JavaScriptInvoke

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2876,7 +2876,7 @@
     "karma-firefox-launcher": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/karma-firefox-launcher/-/karma-firefox-launcher-1.1.0.tgz",
-      "integrity": "sha512-LbZ5/XlIXLeQ3cqnCbYLn+rOVhuMIK9aZwlP6eOLGzWdo1UVp7t6CN3DP4SafiRLjexKwHeKHDm0c38Mtd3VxA==",
+      "integrity": "sha1-LEcDBFLwRTHrfRPU/HZpYwu5Mzk=",
       "dev": true
     },
     "karma-ie-launcher": {
--- a/source/core/CEntryPoints.cpp
+++ b/source/core/CEntryPoints.cpp
@@ -17,6 +17,7 @@ SDG
 #include "TDCodecLVFlat.h"
 #include "TDCodecVia.h"
 #include "CEntryPoints.h"
+#include "ObjectRef.h"
 
 #if defined (VIREO_C_ENTRY_POINTS)
 namespace Vireo {
@@ -392,6 +393,11 @@ VIREO_EXPORT Double Data_ReadSingle(Single* singlePointer)
 VIREO_EXPORT Double Data_ReadDouble(Double* doublePointer)
 {
     return *doublePointer;
+}
+//------------------------------------------------------------
+VIREO_EXPORT Double Data_ReadObjectRefNum(ObjectRefNum* refNum)
+{
+	return *refNum;
 }
 //------------------------------------------------------------
 VIREO_EXPORT void Data_WriteBoolean(Boolean* destination, Int32 value)
--- a/source/core/ObjectRef.cpp
+++ b/source/core/ObjectRef.cpp
@@ -1,0 +1,142 @@
+
+/**
+
+ Copyright (c) 2018 National Instruments Corp.
+
+ This software is subject to the terms described in the LICENSE.TXT file
+
+ SDG
+ */
+
+/*! \file
+ \brief Opaque Object reference management
+ */
+
+#include "TypeDefiner.h"
+#include "ExecutionContext.h"
+#include "VirtualInstrument.h"
+#include "RefNum.h"
+#include "ObjectRef.h"
+
+namespace Vireo {
+
+//TODO: do something similar
+
+// ObjectRefInfo -- Storage for static opaque object references.
+struct ObjectRefInfo {
+    VirtualInstrument *vi;  // VI containing object
+    StringRef objectTag;   // Unique identifying tag for object // TODO: what is this going to be?
+
+    ObjectRefInfo() : vi(NULL), objectTag(NULL) { }
+    ObjectRefInfo(VirtualInstrument *aVi, StringRef aObjectTag) : vi(aVi), objectTag(aObjectTag) { }
+};
+
+// ObjectRefNumManager -- Allocation, deallocation, and lookup of object references
+class ObjectRefNumManager : public RefNumManager {
+ private:
+    typedef TypedRefNum<ObjectRefInfo, true> ObjectRefNumType;
+    ObjectRefNumType _refStorage;  // manages refnum storage
+
+    static ObjectRefNumManager _s_singleton;
+ public:
+    static ObjectRefNumManager &ObjectRefManager() { return _s_singleton; }
+    static ObjectRefNumType &RefNumStorage() { return _s_singleton.RefNumManager(); }
+
+    ObjectRefNumType &RefNumManager() { return _refStorage; }
+};
+
+ObjectRefNumManager ObjectRefNumManager::_s_singleton;
+
+// ObjectReferenceDestroy -- deallocate object ref
+// ObjectRefInfo owns objectTag string, so it needs to be deallocated before disposing the refnum.
+NIError ObjectReferenceDestroy(ObjectRefNum refnum) {
+    ObjectRefInfo ObjectRefInfo;
+    NIError err = ObjectRefNumManager::RefNumStorage().GetRefNumData(refnum, &ObjectRefInfo);
+    if (err == kNIError_Success) {
+        ObjectRefInfo.objectTag->Delete(ObjectRefInfo.objectTag);
+    }
+    return ObjectRefNumManager::RefNumStorage().DisposeRefNum(refnum, NULL);
+}
+
+// Clean-up Proc, run when top level VI finishes, disposing object reference.
+static void CleanUpObjectReference(intptr_t arg) {
+    ObjectRefNum refnum = ObjectRefNum(arg);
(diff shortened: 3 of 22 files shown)


package-lock.json
```diff
@@ -2876,7 +2876,7 @@
     "karma-firefox-launcher": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/karma-firefox-launcher/-/karma-firefox-launcher-1.1.0.tgz",
-      "integrity": "sha512-LbZ5/XlIXLeQ3cqnCbYLn+rOVhuMIK9aZwlP6eOLGzWdo1UVp7t6CN3DP4SafiRLjexKwHeKHDm0c38Mtd3VxA==",
+      "integrity": "sha1-LEcDBFLwRTHrfRPU/HZpYwu5Mzk=",
```
rajsite: you may need to update your npm version to 5.6.0 to prevent this change
`npm i -g npm@5.6.0`

source/core/CEntryPoints.cpp
```diff
@@ -394,6 +395,11 @@ VIREO_EXPORT Double Data_ReadDouble(Double* doublePointer)
     return *doublePointer;
 }
 //------------------------------------------------------------
+VIREO_EXPORT Double Data_ReadObjectRefNum(ObjectRefNum* refNum)
```
rajsite: Should use Int32 for the refNum return type

source/core/ObjectRef.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+VIREO_FUNCTION_SIGNATURE3(IsEQRefnum, RefNumVal, RefNumVal, Boolean);
+VIREO_FUNCTION_SIGNATURE3(IsNERefnum, RefNumVal, RefNumVal, Boolean);
+
+//------------------------------------------------------------
+DEFINE_VIREO_BEGIN(ObjectRefs)
+    DEFINE_VIREO_REQUIRE(VirtualInstrument)
+
+    DEFINE_VIREO_TYPE(ObjectRefNumInfo, "c(e(VirtualInstrument vi) e(String objectTag))")
```
rajsite: Don't think the ObjectRef should have a reference to the VI...

source/io/module_eggShell.js
```diff
@@ -320,6 +322,8 @@
         var arrayRankPointer = Module._malloc(4);
         var vireoObjectPointer = Module._malloc(4);
         var vireoTypePointer = Module._malloc(4);
+		
+		var objectRefNumToJsObjectMap = {};
```
rajsite: We should use a Map instead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

source/io/module_javaScriptInvoke.js
```diff
@@ -213,7 +213,14 @@
                 isValidReturnType: function (value) {
                     return value instanceof Float64Array;
                 }
-            }
+            },
+			ObjectRefNum: {
+                reader: Module.eggShell.dataReadObjectRefNum,
+                writer: Module.eggShell.dataWriteObjectRefNum,
+                isValidReturnType: function (value) {
+                    return typeof value === 'object';
```
rajsite: Don't think we should validate the value here. The user could pass null, undefined, etc. Maybe just always return true

source/io/module_javaScriptInvoke.js
```diff
@@ -287,6 +294,7 @@
             (returnTypeName === 'boolean') ||
             (returnTypeName === 'string') ||
             (returnTypeName === 'undefined') ||
+			(returnTypeName === 'object') ||
```
rajsite: Since an opaque reference lets anything through this check doesn't make sense. I think the whole isValidJavaScriptReturnType function can be removed because we already validate the return value further down: https://github.com/ni/VireoSDK/blob/7cf54a5755337a71f11497ebb812c59a4cd14f84/source/io/module_javaScriptInvoke.js#L331

rajsite:
I think the private flag on the JavascriptInvoke instruction (that we whiteboarded Friday) is a significant enough change that we should introduce it now to make sure it does not cause problems.

It will be needed both to avoid global function calls and to avoid adding api functions to the vireo object unnecessarily.

source/core/TDCodecVia.cpp
```diff
@@ -2338,8 +2374,10 @@ void TDViaFormatter::FormatData(TypeRef type, void *pData)
                     _string->Append(name.Length(), (Utf8Char*)name.Begin());
                 else
                     _string->AppendCStr(tsRefNumTypeToken);
-                if (name.CompareCStr(tsControlRefNumToken))
-                    ControlReferenceAppendDescription(_string, refnum);
+				if (name.CompareCStr(tsControlRefNumToken))
+					ControlReferenceAppendDescription(_string, refnum);
+				else if (name.CompareCStr(tsObjectRefNumToken))
+					ObjectReferenceAppendDescription(_string, refnum);
```
spathiwa: tabbing issues here?

source/core/TDCodecVia.cpp
```diff
@@ -1354,9 +1386,13 @@ Int32 TDViaParser::ParseData(TypeRef type, void* pData)
...
-            if (tt == TokenTraits_SymbolName && token.CompareCStr(tsControlReferenceToken))
-                ParseControlReference(pData);
-            }
+			if (tt == TokenTraits_SymbolName) {
+				if (token.CompareCStr(tsControlReferenceToken))
+					ParseControlReference(pData);
+				else if (token.CompareCStr(tsObjectReferenceToken))
+					ParseObjectReference(pData);
+			}
+			}
```
spathiwa: check your whitespace usage

source/core/ObjectRef.cpp
spathiwa: Why not? The implementation parallels control ref, which makes sense to me.

source/core/ObjectRef.cpp
```diff
@@ -0,0 +1,142 @@
...
+    DEFINE_VIREO_TYPE(ObjectRefNum, "refnum(ObjectRefNumInfo)")
```
sanmut: I think we should call it JSObjectRefNum. Vireo can have other object refnum's too.

spathiwa: So far, there isn't anything on the Vireo side to tie this to JS, and from what I understand, that association will be created via emitted JSLI invoke calls, so there anything wrong with keeping it generic?  It might be able to be associated with other generic kinds of objects as well in the future.

rajsite: JavaScript has several primitive types (which are not objects) and everything else that is an Object: https://www.ecma-international.org/ecma-262/6.0/#sec-ecmascript-language-types

So if I'm putting on my JS pedant hat then I'd propose `JSValueRefNum`

@spathiwa We plan on exposing the type to users in the LabVIEW editor, does it still make sense for it to be a generic reference then?

spathiwa: @rajsite Ah.  Definitely not, then.  

But actually, this code, the bulk of it which creates Vireo RefNum storage for the ObjectRefNumInfo, does not jive with the Event code Jared has sent, or our discussions about this today.  That code used .JavaScriptRefNum, and expects it to be managed entirely on the JS side, so there should not even be a RefNumManager for it on the Vireo side.  At all.  Right?

source/core/ObjectRef.cpp
```diff
@@ -0,0 +1,142 @@
...
+namespace Vireo {
+
+//TODO: do something similar
+
+// ObjectRefInfo -- Storage for static opaque object references.
+struct ObjectRefInfo {
+    VirtualInstrument *vi;  // VI containing object
+    StringRef objectTag;   // Unique identifying tag for object // TODO: what is this going to be?
+
+    ObjectRefInfo() : vi(NULL), objectTag(NULL) { }
```
mehagar: Just a reminder to use nullptr instead of NULL or null going forward.

source/io/module_javaScriptInvoke.js
```diff
@@ -309,13 +302,6 @@
             errorStatusPointer,
             errorCodePointer,
             errorSourcePointer) {
-            if (!isValidJavaScriptReturnType(returnUserValue)) {
```
mehagar: Why is this being removed?

rajsite: Because with the JSLI Object reference every possible JS value is a valid return value.

The problem is that this check is too broad, it needs to be specific to the return value the node is configured for not all possible return values.

For example if the JavaScriptInvokeNode return type is of type Double then we might only accept JavaScript Numbers as a supported type. We might also in the future support strings and automatically parse it as a double if that is a useful feature.

For the JSObjectReference return type any possible JavaScript value is valid.

Luckily we already have a check that verifies the return type to the JS type: https://github.com/ni/VireoSDK/pull/444/files/80cc0fabd7d9ebd3498ddecd9608fa9def77a580#diff-a7468c287af215525c8ba9f958336bd4R317

So the other check was redundant anyway and can be removed.

source/core/ObjectRef.cpp
rajsite: The ObjectRefNumInfo will not be part of the final submission. JSRefnums will not have data associated with them from via. This is a proof of concept / prebuddy / request for comment. And the feedback for not associating a VI and string with the refnum was already taken but not shown in the code review.

There needs to be a vireo type associated with the refnum but you are right that the manager could be implemented in JS or C++. I don't feel strongly either way other than it would be nice if there was just one system handling allocating refnum values and tracking them. I would lean towards the C++ side as maybe in the future there will be a use case for passing a queue refnum to the javascript node as well as jsrefnums.

ggumen:
This was a prototype code intended for getting feedback. We got a lot of good feedback and lot more discussions after this. Closing this review.

PR #445, opened 2018-06-06
Fix some warnings

mehagar:
These aren't all of the warnings, just some of the easier ones to fix. The rest are mostly warnings about casting larger ints to smaller ints or similar.

### DIFF
--- a/source/core/Events.cpp
+++ b/source/core/Events.cpp
@@ -160,7 +160,7 @@ class EventQueueObject {
 
 typedef std::vector<EventQueueObject> EventQueueObjectVector;
 
-typedef Int32 EventOracleIndex;
+typedef IntIndex EventOracleIndex;
 enum { kNotAnEventOracleIdx = -1, kAppEventOracleIdx = 0 };
 
 // Event Oracle -- manage all event registration and delivery
@@ -326,7 +326,7 @@ bool EventOracle::UnregisterForEvent(EventQueueID qID, EventSource eSource, Even
     if (controlUID) {
         // TODO(spathiwa) -- look up index for given control
     }
-    if (qID >= _qObject.size()) {
+    if (static_cast<size_t>(qID) >= _qObject.size()) {
         THREAD_EXEC()->LogEvent(EventLog::kHardDataError, "UnregisterForEvents with invalid QueueiD");
         return false;
     }
@@ -830,7 +830,7 @@ VIREO_FUNCTION_SIGNATURE2(UnregisterForEvents, RefNumVal, ErrorCluster)
 
 //------------------------------------------------------------
 struct WaitForEventArgs {
-    UInt32 *eventSpecIndex;
+    IntIndex *eventSpecIndex;
     StaticTypeAndData data;
     InstructionCore *branchTarget;
 };
@@ -943,7 +943,7 @@ VIREO_FUNCTION_SIGNATUREV(WaitForEventsAndDispatch, WaitForEventsParamBlock)
                             eventRegRefnumPtr[regRefIndex-1].Type()->GetSubElement(0) : nullptr) : 0;
 
     for (Int32 inputTuple = 0; inputTuple < numTuples; ++inputTuple) {
-        UInt32 eventSpecIndex = *arguments[inputTuple].eventSpecIndex;
+        IntIndex eventSpecIndex = *arguments[inputTuple].eventSpecIndex;
         if (eventSpecIndex >= esCount) {
             THREAD_EXEC()->LogEvent(EventLog::kHardDataError, "Invalid Event Spec Index");
             return THREAD_EXEC()->Stop();
--- a/source/core/Events.cpp
+++ b/source/core/Events.cpp
@@ -184,7 +184,7 @@ class EventOracle {
         return _qObject[qID].DoneProcessingEvent();
     }
     void DeleteEventQueue(EventQueueID qID) {  // marks unallocated
-        if (qID < _qObject.size())
+        if (UInt32(qID) < _qObject.size())
             _qObject[qID].DeleteQueue();
     }
     Int32 GetPendingEventInfo(EventQueueID *pActiveQID, Int32 nQueues, RefNumVal *dynRegRefs, Int32 *dynIndexBase);
@@ -207,7 +207,7 @@ EventOracle EventOracle::_s_singleton;
 
 // OccurEvent -- broadcast an event to all registered event queues
 void EventOracle::OccurEvent(EventOracleIndex eventOracleIdx, const EventData &eData) {
-    if (eventOracleIdx < _eventReg.size()) {
+    if (UInt32(eventOracleIdx) < _eventReg.size()) {
         EventRegList &eRegList = _eventReg[eventOracleIdx]._eRegList;
         EventRegList::iterator eRegIter = eRegList.begin(), eRegIterEnd = eRegList.end();
         while (eRegIter != eRegIterEnd && (eRegIter->_eventSource != eData.common.eventSource
@@ -283,7 +283,7 @@ bool EventOracle::RegisterForEvent(EventQueueID qID, EventSource eSource, EventT
     EventOracleIndex eventOracleIndex = kNotAnEventOracleIdx;
     if (oracleIdxPtr)
         *oracleIdxPtr = eventOracleIndex;
-    if (qID >= _qObject.size()) {
+    if (UInt32(qID) >= _qObject.size()) {
         THREAD_EXEC()->LogEvent(EventLog::kHardDataError, "RegisterForEvents with invalid QueueiD");
         return false;
     }
@@ -303,7 +303,7 @@ bool EventOracle::RegisterForEvent(EventQueueID qID, EventSource eSource, EventT
     }
     if (eventOracleIndex == kNotAnEventOracleIdx) {  // no eventOracleIndex, allocate a new one
         size_t size = _eventReg.size(), idx = kAppEventOracleIdx+1;
-        while (eventOracleIndex < size && !_eventReg[eventOracleIndex]._eRegList.empty()) {
+        while (UInt32(eventOracleIndex) < size && !_eventReg[eventOracleIndex]._eRegList.empty()) {
             ++eventOracleIndex;
         }
         if (idx < size) {  // we found an unused index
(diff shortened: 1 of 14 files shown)


rajsite:
From 98 warnings to 42, nice 👍 

https://ci.appveyor.com/project/vireobot/vireosdk/build/1.0.537/messages

https://ci.appveyor.com/project/vireobot/vireosdk/build/1.0.538/messages

source/core/Events.cpp
```diff
@@ -160,7 +160,7 @@ class EventQueueObject {
 
 typedef std::vector<EventQueueObject> EventQueueObjectVector;
 
-typedef Int32 EventOracleIndex;
+typedef IntIndex EventOracleIndex;
```
cglzaguilar: Use size_t, instead of IntIndex and remove the static_casts, do the same for EventQueueID and remove the casts introduced for qID.

source/core/Events.cpp
```diff
@@ -829,7 +829,7 @@ VIREO_FUNCTION_SIGNATURE2(UnregisterForEvents, RefNumVal, ErrorCluster)
 
 //------------------------------------------------------------
 struct WaitForEventArgs {
-    UInt32 *eventSpecIndex;
+    IntIndex *eventSpecIndex;
```
cglzaguilar: Leave this as it was before (UInt32)

source/core/Events.cpp
```diff
@@ -943,7 +943,7 @@ VIREO_FUNCTION_SIGNATUREV(WaitForEventsAndDispatch, WaitForEventsParamBlock)
                             eventRegRefnumPtr[regRefIndex-1].Type()->GetSubElement(0) : nullptr) : 0;
 
     for (Int32 inputTuple = 0; inputTuple < numTuples; ++inputTuple) {
-        UInt32 eventSpecIndex = *arguments[inputTuple].eventSpecIndex;
+        IntIndex eventSpecIndex = *arguments[inputTuple].eventSpecIndex;
```
cglzaguilar: Leave this as it was before (UInt32)

source/core/Events.cpp
```diff
@@ -943,7 +943,7 @@ VIREO_FUNCTION_SIGNATUREV(WaitForEventsAndDispatch, WaitForEventsParamBlock)
...
         if (eventSpecIndex >= esCount) {
```
cglzaguilar: And either change the type of esCount to be UInt32 or cast here.

source/core/Events.cpp
spathiwa: Disagree, size_t is no good because:
  1) it's unsigned, but the sentinel kNotAnEventOracleIndex is -1.  Zero is a valid EventOracleIndex.
   We could change this (not much code relies on it), but this usage matches NCG.
 2) It's 8 bytes on native 64-bit platforms; I want it to be the same size everywhere.

source/core/NumericString.cpp
```diff
@@ -488,7 +488,7 @@ void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], Strin
                             if (tempDouble < 0) {
                             absDouble = 0.0 - tempDouble;
                             }
-                            exponent = floor(log10(absDouble));
+                            exponent = static_cast<Int32>(floor(log10(absDouble)));
```
spathiwa: I think this is the sort of usage that reads better as a constructor rather than a cast:
`exponent = Int32(floor(log10(absDouble)));`

source/core/NumericString.cpp
```diff
@@ -1315,7 +1315,7 @@ void DoubleScanString(StaticTypeAndData* argument, TypeRef argumentType, TempSta
         case 'B':
         case 'o': {
                 intValue = ScanIntBaseValues(formatChar, beginPointer, endPointer);
-                doubleValue = intValue;
+                doubleValue = static_cast<Double>(intValue);
```
spathiwa: Ditto.

source/core/NumericString.cpp
```diff
@@ -1349,7 +1349,7 @@ void DoubleScanString(StaticTypeAndData* argument, TypeRef argumentType, TempSta
             break;
         default: {
                 intValue = strtoll(beginPointer, endPointer, 10);
-                doubleValue = intValue;
+                doubleValue = static_cast<Double>(intValue);
```
spathiwa: Ditto.

source/core/String.cpp
```diff
@@ -508,15 +508,15 @@ VIREO_FUNCTION_SIGNATURE2(IsPrintable, StringRef, Boolean)
...
+        _Param(1) = isprint(c) != 0;
     }
     return _NextInstruction();
 }
 
 VIREO_FUNCTION_SIGNATURE2(IsPrintableInt, Int32, Boolean)
 {
     Utf8Char c = _Param(0);
-    _Param(1) = isprint(c);
+    _Param(1) = isprint(c) != 0;
```
spathiwa: Parentheses might help readability.

source/core/TDCodecVia.cpp
```diff
@@ -1626,7 +1626,7 @@ void TDViaParser::PreParseClump(VIClump* viClump)
             break;
         } else {
             // Read its arguments.
-            tokenFound = _string.ReadSubexpressionToken(&token);
+            tokenFound = _string.ReadSubexpressionToken(&token) != TokenTraits_Unrecognized;
```
spathiwa: Ditto.

source/core/Events.cpp
cglzaguilar: Ok, I wasn't aware of the negative value.

source/core/Events.cpp
mehagar: See my next comment for why I would prefer to keep this as a signed int.

source/core/Events.cpp
mehagar: esCount is assigned from the value of a call to TypeCommon::SubElementCount, which returns an Int32. For the reason in my above comment I would prefer to keep this a signed int.

spathiwa: This is an argument to WaitForEvents which comes from via code and is checked at run-time.
Changing it to Int32 would require also checking to see it's non-negative to validate it.
So this warning fix would slow run-time; that's not a good trade-off.

Instead, make esCount match.

`UInt32 esCount = UInt32(eventSpecType->SubElementCount());`

source/core/NumericString.cpp
sanmut: With modern C++, we could use uniform initialization (using braces instead of parentheses) by default. Where brace initialization throws compiler errors (due to narrowing), we could use C++ style casts. Is that acceptable to everyone?

source/core/Events.cpp
```diff
@@ -184,7 +184,7 @@ class EventOracle {
         return _qObject[qID].DoneProcessingEvent();
     }
     void DeleteEventQueue(EventQueueID qID) {  // marks unallocated
-        if (qID < _qObject.size())
+        if (UInt32(qID) < _qObject.size())
```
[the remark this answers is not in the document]
mehagar: For now I am going to revert this change.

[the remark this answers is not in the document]
spathiwa: Using size_t(qID) would work and avoid a warning.

source/core/GenericFunctions.cpp
```diff
@@ -722,7 +722,7 @@ static bool GetMinimumArrayDimensions(std::vector<TypedArrayCoreRef> arrays,
     IntIndex *array0Lengths = arrays[0]->DimensionLengths();
     for (int i = 0; i < array0Rank; i++) {
         IntIndex dimension0 = *(array0Lengths + i);
-        for (int j = 1; j < arrayCount; j++) {
+        for (int j = 1; j < Int32(arrayCount); j++) {
```
cglzaguilar: I think in this one we can make the type of j to be size_t, that way we can remove the cast

source/core/Timestamp.cpp
```diff
@@ -141,7 +139,7 @@ namespace Vireo
         time_t t = _mkgmtime(&timeVal);
     #endif
     #endif
-        *this = Timestamp((Double)t + kStdDT1970re1904, fracSecs * 18446744073709551616.0);
+        *this = Timestamp((Double)t + kStdDT1970re1904 + fracSecs);
```
cglzaguilar: Please elaborate why this change is needed. This appears to be a change of behavior rather than a warning fix.

source/core/Events.cpp
cglzaguilar: I'm deferring review of changes to this file (Events.cpp) to @spathiwa

source/include/BuildConfig.h
```diff
@@ -233,6 +233,7 @@ SDG
 #endif
 
 // TODO(PaulAustin): allow for thread locals on linux/unix
+#undef VIVM_THREAD_LOCAL
```
cglzaguilar: Try the following instead:
#ifndef VIVM_THREAD_LOCAL
#define VIVM_THREAD_LOCAL
#endif

source/core/VirtualInstrument.cpp
```diff
@@ -696,7 +696,7 @@ void ClumpParseState::InternalAddArgNeedingPatch(PatchInfo::PatchType patchType,
     // it should point to nullptr when checked if not yet resolved.
     _argPatches.push_back(_argCount);
     ++_argPatchCount;
-    if (_patchInfos.size() <= _patchInfoCount)
+    if (_patchInfos.size() <= UInt32(_patchInfoCount))
```
cglzaguilar: Can we make the type of _patchInfoCount be size_t instead?

source/core/Timestamp.cpp
mehagar: Yes, this was something I found while going through this file. It is the same fix as my previous timestamp precision fix, where we were losing precision by using the Timestamp constructor with two parameters. I am using the one with a Double parameter here instead, which preserves the precision. I don't know if this manifested as a bug or not.

source/core/GenericFunctions.cpp
```diff
@@ -722,7 +722,7 @@ static bool GetMinimumArrayDimensions(std::vector<TypedArrayCoreRef> arrays,
     IntIndex *array0Lengths = arrays[0]->DimensionLengths();
     for (int i = 0; i < array0Rank; i++) {
         IntIndex dimension0 = *(array0Lengths + i);
-        for (int j = 1; j < Int32(arrayCount); j++) {
+        for (int j = 1; j < size_t(arrayCount); j++) {
```
cglzaguilar: Actually I was thinking something like this:
for (size_t j = 1; j<arrayCount; j++)

PR #446, opened 2018-06-06
Fix a bug in event cluster registration.

spathiwa:
The ElementOffset field of a Vireo cluster type is absolute from the
beginning of the cluster, not relative to the previous element.
Misuse of this caused access of the wrong memory when looking up
registered event data.

### DIFF
--- a/source/core/Events.cpp
+++ b/source/core/Events.cpp
@@ -218,7 +218,8 @@ void EventOracle::OccurEvent(EventOracleIndex eventOracleIdx, const EventData &e
             EventRegQueueIDList::iterator rqIter = eRegIter->_qIDList.begin(), rqIterEnd = eRegIter->_qIDList.end();
             RefNum ref = eData.common.eventRef.GetRefNum();
             while (rqIter != rqIterEnd) {
-                if (rqIter->_ref == ref)
+                RefNum rRef = rqIter->_ref;
+                if (rRef == ref)
                     _qObject[rqIter->_qID].OccurEvent(eData);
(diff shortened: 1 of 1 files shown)


source/core/Events.cpp
```diff
@@ -692,9 +693,9 @@ LVError RegisterForEventsCore(EventQueueID qID, DynamicEventRegInfo *regInfo, In
         Int32 numElts = refType->SubElementCount();
         for (Int32 j = 0; j < numElts; ++j) {
             TypeRef eltType = refType->GetSubElement(j);
-            eltPtr += eltType->ElementOffset();
-            if (oldEltPtr)
-                oldEltPtr += eltType->ElementOffset();
+            eltPtr = refClustPtr + eltType->ElementOffset();
+            if (pOldData)
+                oldEltPtr = (AQBlock1*)pOldData + eltType->ElementOffset();
```
mehagar: Could you change this to be a static_cast here? It would make it easy in the future to search for type casts via ctrl+f, and as you were saying earlier the C style cast can do different things depending on the context.

cglzaguilar: We need merge, now. This update can be done later.

spathiwa: The code is actually replete with this style of cast to (AQBlock1*), because that's our generic pointer to unknown data used by StaticTypeAndData, etc.   I'd rather not do just this one; we should fix all of them if we're doing to do one.  I think having a well-named typedef alias for AQBlock1* and using constructor-type casts would be the most readable.

PR #447, opened 2018-06-08
Make EventSpec type built in for more compact event structure configuration

spathiwa:


source/core/Events.cpp
```diff
@@ -1011,6 +1011,9 @@ VIREO_FUNCTION_SIGNATURE3(IsNERefnum, RefNumVal, RefNumVal, Boolean);
 //------------------------------------------------------------
 DEFINE_VIREO_BEGIN(Events)
 
+    // EventSpec used for event structure configuration
+    DEFINE_VIREO_TYPE(EventSpec, "c(e(UInt32 eventSource)e(UInt32 eventType)e(UInt32 controlUID)e(UInt32 dynIndex)");
```
mehagar: Is it OK to have spaces in between the different "e()"s? That would make it a bit easier for me to read.

spathiwa: Sure.

rajsite:
@segaljared review pending discussion on EventSpec possible changes

PR #448, opened 2018-06-12
Remove unmaintained WebSocket prototype

rajsite:
Remove the WebSocket prototype as it is not maintained and if implemented in the future would likely be through the JavaScriptInvoke mechanism instead


sanmut:
Curious. What works and what does not work?

rajsite:
Well nothing works because the js implementation is commented out: https://github.com/ni/VireoSDK/pull/448/files#diff-6080463fda65f9091a7c6e7cf9540134

But conceptually there are a bunch of problems too. Like the jsWebSocketClientRead function registers a new event listener every time it is called causing a memory leak, it does not buffer messages, it relies on polling instead of the asynchronous waiting capabilities that were available to HTTP at the time, does not support the range of UTF-8 characters the rest of our clients do, does not do correct error handling, does not support the channels feature of WebSockets, it does not use the module pattern the rest of the JS code has adopted, it has no tests.

I do believe a rewrite at this point would be easier than adopting any of this code. And I fully expect that we would implement a WebSocket client leveraging the JavaScriptInvoke instead of the approach done here anyway.

sanmut:
Got it, thanks.

PR #449, opened 2018-06-19
Create the valueRef object and createValueRef helper

rajsite:
Also has an initial incomplete implementation of the findValueRef function and the modified readDouble function to use the valueRef object.


test-it/karma/publicapi/FindValueRef.Test.js
```diff
@@ -0,0 +1,27 @@
...
+        fixtures.preloadAbsoluteUrls([
+            publicApiMultipleTypesViaUrl
+        ], done);
+    });
+
+    beforeAll(function () {
+        vireoRunner.rebootAndLoadVia(vireo, publicApiMultipleTypesViaUrl);
+    });
+
+    it('find a double', function () {
```
gleonoliva: Maybe add another test for the error case. Either when the viName was not declared or the path doesn't exist.

Issue #450, opened 2018-06-20
PropertyNode should use opaque JavaScript object reference


ggumen:
To allow PropertyNode implementation to be control agnostic and be easy to add new properties, we want to use JavaScriptInvoke calls with javascript opaque object reference.

This GitHub Issue is created to keep track of all the incremental commits that we will make to support this feature.

rajsite:
Fixed by #647

PR #451, opened 2018-06-20
Introducing JavaScriptReference to vireo

ggumen:
Introducing javascript opaque object reference to vireo (#450)
Adding a small basic test.
Extending PredicateFunctions.via test to also test IsNotANumPathRefnum with integer, to make sure we don't override the behavior for integers.

### DIFF
--- a/source/include/JavaScriptRef.h
+++ b/source/include/JavaScriptRef.h
@@ -1,0 +1,42 @@
+/**
+
+ Copyright (c) 2018 National Instruments Corp.
+
+ This software is subject to the terms described in the LICENSE.TXT file
+
+ SDG
+ */
+
+/*! \file
+ \brief JavaScript refnum
+
+ JavaScript references are defined in the Params or Locals section of a VI,
+ and define a named, non-statically-bound reference to a JavaScript opaque object.
+ 
+ Example:
+ 
+ define(MyVI dv(.VirtualInstrument (
+ Locals: c(
+ e(JavaScriptRefNum  jsref)  // JavaScript ref that is not statically linked
+ ) ...
+ 
+ The actual Type of jsref is JavaScriptRefNum and stores a refnum (cookie) which
+ opaquely holds onto the javascript object.
+ 
+ These can be used in subVIs to refer to a reference passed in by a caller.
+*/
+
+#ifndef JavaScriptRef_h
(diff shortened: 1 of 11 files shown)


source/include/JavaScriptRef.h
```diff
@@ -0,0 +1,42 @@
...
+ 
+ Example:
+ 
+ define(MyVI dv(.VirtualInstrument (
+ Locals: c(
+ e(JavaScriptRefNum  jsref)  // JavaScript ref that is not statically linked
+ ) ...
+ 
+ The actual Type of jsref is JavaScriptRefNum and stores a refnum (cookie) which
+ opaquely holds onto the javascript object.
```
spathiwa: Please mention here that the refnum is managed on the JS side and does not actually use the Vireo RefNumManager.

source/include/JavaScriptRef.h
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+/*! \file
+ \brief JavaScript refnum
+
+ JavaScript references are defined in the Params or Locals section of a VI,
+ and define a named, non-statically-bound reference to a JavaScript opaque object.
+ 
+ Example:
+
```
rajsite: Doxygen supports Markdown so you can specify a code block for this snippet by adding spaces https://www.stack.nl/~dimitri/doxygen/manual/markdown.html#md_codeblock

test-it/ViaTests/JavaScriptRefBasic.via
```diff
@@ -0,0 +1,44 @@
...
+        PrintJavaScriptRef("ref2" ref2)
+        IsNotANumPathRefnum(ref2 bool)
+        Printf("IsNotARefNum: %z\n" bool)
+        IsEQ(ref1 ref2 bool)
+        Printf("IsEQ ref1 ref2: %z\n" bool)
+    )
+
+) ) )
+
+enqueue(JavaScriptRefTestProgram)
```
cglzaguilar: Add a new line at the end of this line to remove the red mark.

cglzaguilar: Please add a subVI case and in the subVI check that IsNotANumPathRefnum returns true, that's all we can do for now. But we get to exercise the subVI case.

PR #452, opened 2018-06-20
Add example vscode intellisense config

rajsite:


gleonoliva:
Other than that, looks great! Thanks Milan!

PR #453, opened 2018-06-20
Type reflection

gleonoliva:
Initial implementation of `reflectOnValueRef` which takes a `valueRef` and is able to appropriately call methods on an object passed as parameter depending on the type visiting.

Right now we can only reflect on the first level of type hierarchy. Once we have `findSubValueRef` we can continue traversing aggregate types.

I found a few things that I'm not entirely sure about, I'll leave as comments asking for suggestions.

### DIFF
--- a/source/core/CEntryPoints.cpp
+++ b/source/core/CEntryPoints.cpp
@@ -503,6 +503,7 @@ VIREO_EXPORT Int32 TypeRef_Alignment(TypeRef typeRef)
 //------------------------------------------------------------
 VIREO_EXPORT const char* TypeRef_Name(TypeRef typeRef)
 {
+    TypeManagerScope scope(typeRef->TheTypeManager());
     SubString name = typeRef->Name();
 
     static StringRef returnBuffer = nullptr;
@@ -521,7 +522,7 @@ VIREO_EXPORT const char* TypeRef_Name(TypeRef typeRef)
         returnBuffer->Append((Utf8Char)'\0');
         return (const char*) returnBuffer->Begin();
     }
-    
+
     return "";
 }
 //------------------------------------------------------------
@@ -590,7 +591,7 @@ VIREO_EXPORT Boolean TypeRef_IsEnum(TypeRef typeRef)
     return typeRef->IsEnum();
 }
 //------------------------------------------------------------
-VIREO_EXPORT Boolean TypeRef_IsFloat(TypeRef typeRef) 
+VIREO_EXPORT Boolean TypeRef_IsFloat(TypeRef typeRef)
 {
     return typeRef->IsFloat();
 }
--- a/source/core/CEntryPoints.cpp
+++ b/source/core/CEntryPoints.cpp
@@ -560,6 +560,66 @@ VIREO_EXPORT TypeRef TypeRef_GetSubElementByIndex(TypeRef typeRef, Int32 index)
     return typeRef->GetSubElement(index);
 }
 //------------------------------------------------------------
+VIREO_EXPORT Boolean TypeRef_IsCluster(TypeRef typeRef)
+{
+    return typeRef->IsCluster();
+}
+//------------------------------------------------------------
+VIREO_EXPORT Boolean TypeRef_IsArray(TypeRef typeRef)
+{
+    return typeRef->IsArray();
+}
+//------------------------------------------------------------
+VIREO_EXPORT Boolean TypeRef_IsBoolean(TypeRef typeRef)
+{
+    return typeRef->IsBoolean();
+}
+//------------------------------------------------------------
+VIREO_EXPORT Boolean TypeRef_IsInteger(TypeRef typeRef)
+{
+    return typeRef->IsInteger();
+}
+//------------------------------------------------------------
+VIREO_EXPORT Boolean TypeRef_IsSigned(TypeRef typeRef)
+{
+    return typeRef->IsSignedInteger();
+}
+//------------------------------------------------------------
+VIREO_EXPORT Boolean TypeRef_IsEnum(TypeRef typeRef)
+{
+    return typeRef->IsEnum();
+}
+//------------------------------------------------------------
+VIREO_EXPORT Boolean TypeRef_IsFloat(TypeRef typeRef) 
+{
+    return typeRef->IsFloat();
+}
+//------------------------------------------------------------
+VIREO_EXPORT Boolean TypeRef_IsString(TypeRef typeRef)
+{
+    return typeRef->IsString();
+}
+//------------------------------------------------------------
+VIREO_EXPORT Boolean TypeRef_IsPath(TypeRef typeRef)
+{
+    return typeRef->IsPath();
+}
+//------------------------------------------------------------
+VIREO_EXPORT Boolean TypeRef_IsTimestamp(TypeRef typeRef)
+{
+    return typeRef->IsTimestamp();
+}
+//------------------------------------------------------------
+VIREO_EXPORT Boolean TypeRef_IsComplex(TypeRef typeRef)
+{
+    return typeRef->IsComplex();
+}
+//------------------------------------------------------------
+VIREO_EXPORT Boolean TypeRef_IsAnalogWaveform(TypeRef typeRef)
+{
+    return typeRef->IsAnalogWaveform();
+}
+//------------------------------------------------------------
 //------------------------------------------------------------
 VIREO_EXPORT Int32 Data_RawBlockSize(TypedBlock* object)
 {
--- a/source/core/CEntryPoints.cpp
+++ b/source/core/CEntryPoints.cpp
@@ -501,10 +501,28 @@ VIREO_EXPORT Int32 TypeRef_Alignment(TypeRef typeRef)
     return typeRef->AQAlignment();
 }
 //------------------------------------------------------------
-VIREO_EXPORT void TypeRef_Name(TypeRef typeRef, Int32* bufferSize, char* buffer)
+VIREO_EXPORT const char* TypeRef_Name(TypeRef typeRef)
 {
     SubString name = typeRef->Name();
-    *bufferSize = name.CopyToBoundedBuffer(*bufferSize, reinterpret_cast<Utf8Char*>(buffer));
+
+    static StringRef returnBuffer = nullptr;
+    if (returnBuffer == nullptr) {
+        // Allocate a string the first time it is used.
+        // After that it will be resized as needed.
+        STACK_VAR(String, tempReturn);
+        returnBuffer = tempReturn.DetachValue();
+    } else {
+        returnBuffer->Resize1D(0);
+    }
+
+    if (returnBuffer) {
+        returnBuffer->AppendSubString(&name);
+        // Add an explicit nullptr terminator so it looks like a C string.
+        returnBuffer->Append((Utf8Char)'\0');
+        return (const char*) returnBuffer->Begin();
+    }
+    
+    return "";
 }
 //------------------------------------------------------------
 VIREO_EXPORT Int32 TypeRef_ElementOffset(TypeRef typeRef)
--- a/source/core/CEntryPoints.cpp
+++ b/source/core/CEntryPoints.cpp
@@ -501,9 +501,9 @@ VIREO_EXPORT Int32 TypeRef_Alignment(TypeRef typeRef)
     return typeRef->AQAlignment();
 }
 //------------------------------------------------------------
-VIREO_EXPORT const char* TypeRef_Name(TypeRef typeRef)
+VIREO_EXPORT const char* TypeRef_Name(TypeManagerRef tm, TypeRef typeRef)
 {
-    TypeManagerScope scope(typeRef->TheTypeManager());
+    TypeManagerScope scope(tm);
     SubString name = typeRef->Name();
 
     static StringRef returnBuffer = nullptr;
--- a/source/core/TypeAndDataManager.cpp
+++ b/source/core/TypeAndDataManager.cpp
@@ -884,7 +884,7 @@ Boolean TypeCommon::IsNumeric()
     return false;
 }
 //------------------------------------------------------------
-Boolean TypeCommon::IsInteger() 
+Boolean TypeCommon::IsInteger()
 {
     TypeRef t = this;
     while (t) {
--- a/source/core/TypeAndDataManager.cpp
+++ b/source/core/TypeAndDataManager.cpp
@@ -612,6 +612,8 @@ const SubString TypeCommon::TypeString = SubString(tsStringType);
 const SubString TypeCommon::TypeTimestamp = SubString("Timestamp");
 const SubString TypeCommon::TypeComplexSingle = SubString("ComplexSingle");
 const SubString TypeCommon::TypeComplexDouble = SubString("ComplexDouble");
+const SubString TypeCommon::TypePath = SubString("NIPath");
+const SubString TypeCommon::TypeAnalogWaveform = SubString("AnalogWaveform");
 const SubString TypeCommon::TypeStaticTypeAndData = SubString("StaticTypeAndData");
 
 TypeCommon::TypeCommon(TypeManagerRef typeManager)
@@ -882,6 +884,34 @@ Boolean TypeCommon::IsNumeric()
     return false;
 }
 //------------------------------------------------------------
+Boolean TypeCommon::IsInteger() 
+{
+    TypeRef t = this;
+    while (t) {
+        if (t->Name().Compare(&TypeInt8) || t->Name().Compare(&TypeInt16) || t->Name().Compare(&TypeInt32)
+            || t->Name().Compare(&TypeInt64) || t->Name().Compare(&TypeUInt8)
+            || t->Name().Compare(&TypeUInt16) || t->Name().Compare(&TypeUInt32)
+            || t->Name().Compare(&TypeUInt64)) {
+            return true;
+        }
+        t = t->BaseType();
+    }
+    return false;
+}
+//------------------------------------------------------------
+Boolean TypeCommon::IsSignedInteger()
+{
+    TypeRef t = this;
+    while (t) {
+        if (t->Name().Compare(&TypeInt8) || t->Name().Compare(&TypeInt16) || t->Name().Compare(&TypeInt32)
+            || t->Name().Compare(&TypeInt64)) {
+            return true;
+        }
+        t = t->BaseType();
+    }
+    return false;
+}
+//------------------------------------------------------------
 Boolean TypeCommon::IsInteger64()
 {
     TypeRef t = this;
@@ -918,7 +948,6 @@ Boolean TypeCommon::IsBoolean()
     }
     return false;
 }
-
 //------------------------------------------------------------
 Boolean TypeCommon::IsString()
 {
@@ -931,7 +960,18 @@ Boolean TypeCommon::IsString()
     }
     return false;
 }
-
+//------------------------------------------------------------
+Boolean TypeCommon::IsPath()
+{
+    TypeRef t = this;
+    while (t) {
+        if (t->Name().Compare(&TypePath)) {
+            return true;
+        }
+        t = t->BaseType();
+    }
+    return false;
+}
 //------------------------------------------------------------
 Boolean TypeCommon::IsTimestamp()
 {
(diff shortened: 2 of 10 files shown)


source/io/module_eggShell.js
```diff
@@ -239,6 +247,257 @@
...
+            validateVisitMethod(typeVisitor.visitArray, 'visitArray');
+            return typeVisitor.visitArray.apply(typeVisitor, args);
+        };
+
+        var dispatchVisitCluster = function (typeVisitor, args) {
+            validateVisitMethod(typeVisitor.visitCluster, 'visitCluster');
+            return typeVisitor.visitCluster.apply(typeVisitor, args);
+        };
+
+        var typeDispatchFunctions = [
```
gleonoliva: Is it a good a idea to put these 2 arrays in a Map?

source/io/module_eggShell.js
```diff
@@ -134,6 +138,9 @@
         var Data_WriteDouble = Module.cwrap('Data_WriteDouble', 'void', ['number', 'number']);
         var EggShell_ExecuteSlices = Module.cwrap('EggShell_ExecuteSlices', 'number', ['number', 'number', 'number']);
         var Occurrence_Set = Module.cwrap('Occurrence_Set', 'void', ['number']);
+        var TypeRef_Name = Module.cwrap('TypeRef_Name', 'string', ['number']);
+        var TypeRef_SubElementCount = Module.cwrap('TypeRef_SubElementCount', 'number', ['number']);
+        var TypeRef_GetSubElementByIndex = Module.cwrap('TypeRef_GetSubElementByIndex', 'number', ['number']);
```
rajsite: If all the types are numbers you don't need to use Module.cwrap

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
             return result;
         };
 
+        Module.eggShell.typeTopAQSize = function (typePointer) {
```
rajsite: To keep the names consistent can you change  typePointer -> typeRef for all these?

source/io/module_eggShell.js
rajsite: Maps are good if things are added and removed dynamically in the application lifecycle. Since this is static maybe an array of objects?

```
var typeHandlers = [{
    typeChecker: Module.eggShell.typeIsBoolean,
    dispatcher: dispatchVisitBoolean
}]
```

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+            Module.eggShell.typeIsComplex,
+            Module.eggShell.typeIsAnalogWaveform,
+            Module.eggShell.typeIsTimestamp,
+            Module.eggShell.typeIsPath,
+            Module.eggShell.typeIsArray,
+            Module.eggShell.typeIsCluster
+        ];
+
+        Module.eggShell.reflectOnValueRef = publicAPI.eggShell.reflectOnValueRef = function (typeVisitor, valueRef) {
+            if (typeof valueRef !== 'object' || Array.isArray(valueRef)) {
```
rajsite: The Array check seems a bit arbitrary. For sanity checking that an object was passed
`typeof valueRef !== 'object' || valueRef === null`
is sufficient.

Same for below, if someone's prerogative is to do:
```
var myWackyVisisitor = [];
myWackyVisitor.visitArray = function () {};
```
then let 'em do it :P

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+                throw new Error('valueRef must be an object. Found: ' + valueRef);
+            }
+
+            if (typeof typeVisitor !== 'object' || Array.isArray(typeVisitor)) {
+                throw new Error('typeVisitor must be an object. Found: ' + typeVisitor);
+            }
+
+            var typeRef = valueRef.typeRef,
+                args = Array.prototype.slice.call(arguments, 1),
```
rajsite: I think it's being a little too clever by letting users pass an arbitrary number of arguments. JS in general is faster when it can know ahead of time the structure of types being used.

So just having a `data` third parameter and then passing a
```
dispatcherParams = {
    type: typeRef,
    data: data
}
```
object to the dispatcher and in the dispatcher doing plain function invocation
```
typeVisitor.visitCluster(dispatcherParams.typeRef, dispatcherParams.data)
```

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+                    break;
+                case 8:
+                    visitFn = typeVisitor.visitUInt64;
+                    break;
+                default:
+                    throw new Error('Unexpected size for Unsigned Integer. Found: ');
+                }
+            }
+
+            var intName = (isSignedInteger ? '' : 'U') + 'Int' + (sizeOfInteger * 8);
```
rajsite: Can we have the string constants assigned in the individual cases instead of string building every time? Seems like extra work for what is used for error messages.

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+                visitFn = typeVisitor.visitSingle;
+                break;
+            case 8:
+                visitFn = typeVisitor.visitDouble;
+                break;
+            default:
+                throw new Error('Unexpected size for a Float value');
+            }
+
+            validateVisitMethod(visitFn, 'visit' + (sizeOfFloat === 4 ? 'Single' : 'Double'));
```
rajsite: same as above, just assign off a string constant in the different cases

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+                visitFn = typeVisitor.visitComplexSingle;
+                break;
+            case 16:
+                visitFn = typeVisitor.visitComplexDouble;
+                break;
+            default:
+                throw new Error('Unexpected size for a Complex value');
+            }
+
+            validateVisitMethod(visitFn, 'visitComplex' + (sizeOfComplex === 8 ? 'Single' : 'Double'));
```
rajsite: same as above, move to constants assigned by the cases

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+
+            for (i = 0; i < registeredTypes.length; i += 1) {
+                if (registeredTypes[i](typeRef) === true) {
+                    foundTypeIndex = i;
+                    break;
+                }
+            }
+
+            if (foundTypeIndex < 0) {
+                throw new Error('Unexpected type. Is typeRef pointing to a valid type?. TypeRef found: ' + typeRef);
```
rajsite: Maybe pass the typeRef and use `eggShell.typeName` to give a better name to the user?

test-it/karma/publicapi/TypeReflection.Test.js
```diff
@@ -0,0 +1,140 @@
...
+
+            it('throws when valueRef is not an object', function () {
+                var throwingFunction = function () {
+                    vireo.eggShell.reflectOnValueRef({}, []);
+                };
+
+                expect(throwingFunction).toThrow();
+            });
+
+            it('throws when visitor does not have visit methods for all types', function () {
```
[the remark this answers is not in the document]
rajsite: Generally if the use case isn't well know I think it's better to throw and then make it less strict if needed. It's a lot trickier to start throwing exceptions than to stop (assuming people are not using exceptions for flow control :P)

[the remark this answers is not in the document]
rajsite: Any use-case you have in mind?

test-it/karma/publicapi/TypeReflection.Test.js
```diff
@@ -0,0 +1,140 @@
...
+    var fixtures = window.testHelpers.fixtures;
+
+    var vireo = new Vireo();
+
+    var publicApiMultipleTypesViaUrl = fixtures.convertToAbsoluteFromFixturesDir('publicapi/MultipleTypes.via');
+    var viName = 'MyVI';
+
+    var typeDescriptor = (function () {
+        var returnTypeName = function (name) {
+            var typeName = name;
```
rajsite: shouldn't need to re-alias this here

source/io/module_eggShell.js
gleonoliva: Agree, I unwrap them when I move them to the module.

source/io/module_eggShell.js
gleonoliva: I'll do

source/io/module_eggShell.js
gleonoliva: Agree, I'll change it.

source/io/module_eggShell.js
gleonoliva: Cool, yeah I forgot about the use case of a Map. I'll switch to this one instead.

source/io/module_eggShell.js
gleonoliva: Yeah, it was just JS catching me off guard, didn't want to allow that, but ok. They'll be responsible for their wacky visitor.

source/io/module_eggShell.js
gleonoliva: Agreed, I'll change it

source/io/module_eggShell.js
gleonoliva: What if typeRef is not even a number? Should we do any kind of validation on that an later call `typeName`?

test-it/karma/publicapi/TypeReflection.Test.js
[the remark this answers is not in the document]
gleonoliva: I like the proposed implementation because if a user decides to write a partial visitor and vireo reflects on a type that the user didn't specify it will throw. The alternative by not throwing would be to give the user the responsibility of keeping track if any of their functions in their visitor was called. Now that I've written down the alternative I'm more convinced to leave it as it is :P

source/io/module_eggShell.js
rajsite: Hmm since we control the creation of typeRefs internally I don't think it will be common for the typeRef to be invalid, but it is likely types will be added that are not supported by the visitor yet. How does that sound?

Maybe a sanity `typeRef === 0 ? 'invalid type' : 'eggShell.typeName(typeRef)'` is sufficient?

source/core/CEntryPoints.cpp
```diff
@@ -501,10 +501,28 @@ VIREO_EXPORT Int32 TypeRef_Alignment(TypeRef typeRef)
...
 }
 //------------------------------------------------------------
-VIREO_EXPORT void TypeRef_Name(TypeRef typeRef, Int32* bufferSize, char* buffer)
+VIREO_EXPORT const char* TypeRef_Name(TypeRef typeRef)
 {
     SubString name = typeRef->Name();
-    *bufferSize = name.CopyToBoundedBuffer(*bufferSize, reinterpret_cast<Utf8Char*>(buffer));
+
+    static StringRef returnBuffer = nullptr;
+    if (returnBuffer == nullptr) {
```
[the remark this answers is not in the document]
gleonoliva: Cool, I'll add it.

source/core/module_typeHelpers.js
```diff
@@ -0,0 +1,306 @@
...
+                case 8:
+                    visitFn = typeVisitor.visitUInt64;
+                    typeName = 'UInt64';
+                    break;
+                default:
+                    throw new Error('Unexpected size for Unsigned Integer. Found: ');
+                }
+            }
+
+            validateVisitMethod(visitFn, 'visit' + typeName);
```
rajsite: Can we move the visit concatenation too? The goal was to prevent the unnecessary allocation :P

source/core/module_typeHelpers.js
```diff
@@ -0,0 +1,306 @@
...
+                break;
+            case 8:
+                visitFn = typeVisitor.visitDouble;
+                typeName = 'Double';
+                break;
+            default:
+                throw new Error('Unexpected size for a Float value');
+            }
+
+            validateVisitMethod(visitFn, 'visit' + typeName);
```
rajsite: same

source/core/CEntryPoints.cpp
```diff
@@ -501,10 +501,29 @@ VIREO_EXPORT Int32 TypeRef_Alignment(TypeRef typeRef)
     return typeRef->AQAlignment();
 }
 //------------------------------------------------------------
-VIREO_EXPORT void TypeRef_Name(TypeRef typeRef, Int32* bufferSize, char* buffer)
+VIREO_EXPORT const char* TypeRef_Name(TypeRef typeRef)
 {
+    TypeManagerScope scope(typeRef->TheTypeManager());
```
rajsite: Talked about it, but should use the same pattern as other functions, sharing the typemanager reference

source/core/module_typeHelpers.js
```diff
@@ -0,0 +1,306 @@
...
+                case 4:
+                    visitFn = typeVisitor.visitUInt32;
+                    typeName = 'UInt32';
+                    break;
+                case 8:
+                    visitFn = typeVisitor.visitUInt64;
+                    typeName = 'UInt64';
+                    break;
+                default:
+                    throw new Error('Unexpected size for Unsigned Integer. Found: ');
```
rajsite: missing the part after `Found:`

source/core/CEntryPoints.cpp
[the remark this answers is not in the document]
cglzaguilar: I don't think we need the TypeManagerScope in this situation. We already have the TypeRef. Most functions use the scope when a TypeManagerRef is passed.

source/core/CEntryPoints.cpp
```diff
@@ -501,10 +501,28 @@ VIREO_EXPORT Int32 TypeRef_Alignment(TypeRef typeRef)
...
+        returnBuffer->AppendSubString(&name);
+        // Add an explicit nullptr terminator so it looks like a C string.
+        returnBuffer->Append((Utf8Char)'\0');
+        return (const char*) returnBuffer->Begin();
+    }
+    
+    return "";
 }
 //------------------------------------------------------------
 VIREO_EXPORT Int32 TypeRef_ElementOffset(TypeRef typeRef)
```
cglzaguilar: Please elaborate why are we returning an Int32, instead of a TypeRef?

source/core/TypeAndDataManager.cpp
```diff
@@ -956,7 +996,20 @@ Boolean TypeCommon::IsComplex()
...
     return false;
 }
-
+//------------------------------------------------------------
+Boolean TypeCommon::IsAnalogWaveform()
+{
+    TypeRef t = this;
+    while (t) {
+        SubString typeName = t->Name();
+        SubString analogTypess(TypeAnalogWaveform);
```
cglzaguilar: Typo: change analogTypess to analogTypes

source/core/CEntryPoints.cpp
gleonoliva: Thanks for explaining the difference :)

source/core/module_typeHelpers.js
gleonoliva: Good catch, I added it to the other dispatchers too.

source/core/TypeAndDataManager.cpp
```diff
@@ -956,7 +996,20 @@ Boolean TypeCommon::IsComplex()
...
+        if (typeName.FindFirstMatch(&analogTypess, 0, false) == 0) {
```
cglzaguilar: Please elaborate why are we doing a FindFirstMatch and not a Compare as in the TypeCommon::IsPath case. If we decide to have a type AnalogWaweformInternal in the future, it might find this one instead.

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+        var dispatchVisitEnum = function (typeVisitor, args) {
+            validateVisitMethod(typeVisitor.visitEnum, 'visitEnum');
+            return typeVisitor.visitEnum.apply(typeVisitor, args);
+        };
+
+        var dispatchVisitInteger = function (typeVisitor, args) {
+            var typeRef = args[0].typeRef;
+            var isSignedInteger = Module.eggShell.typeIsSigned(typeRef);
+            var sizeOfInteger = Module.eggShell.typeTopAQSize(typeRef);
+            var visitFn = undefined;
```
cglzaguilar: Consider renaming visitFn to functionToVisit

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+
+        Module.eggShell.typeIsComplex = function (typePointer) {
+            return Module._TypeRef_IsComplex(typePointer) !== 0;
+        };
+
+        Module.eggShell.typeIsAnalogWaveform = function (typePointer) {
+            return Module._TypeRef_IsAnalogWaveform(typePointer) !== 0;
+        };
+
+        var validateVisitMethod = function (fn, fnName) {
```
cglzaguilar: Consider renaming validateVisitMethod to validateVisitFunction.  Consider renaming fn to functionToVisit or methodToVisit depending on the name of the function you selected.

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+                    visitFn = typeVisitor.visitInt16;
+                    break;
+                case 4:
+                    visitFn = typeVisitor.visitInt32;
+                    break;
+                case 8:
+                    visitFn = typeVisitor.visitInt64;
+                    break;
+                default:
+                    throw new Error('Unexpected size for Integer');
```
cglzaguilar: Add more debugging information to the error as follows: "Unexpected integer size. Found: " + sizeOfInteger. That way if something bad happens you won't wonder about the integer size that caused the exception.

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+                    visitFn = typeVisitor.visitUInt16;
+                    break;
+                case 4:
+                    visitFn = typeVisitor.visitUInt32;
+                    break;
+                case 8:
+                    visitFn = typeVisitor.visitUInt64;
+                    break;
+                default:
+                    throw new Error('Unexpected size for Unsigned Integer. Found: ');
```
cglzaguilar: Interesting. Here you thought about returning the value, but it is missing, please add it.

source/io/module_eggShell.js
cglzaguilar: Consider using integerName instead of intName.

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+
+            var intName = (isSignedInteger ? '' : 'U') + 'Int' + (sizeOfInteger * 8);
+            validateVisitMethod(visitFn, 'visit' + intName);
+            return visitFn.apply(typeVisitor, args);
+        };
+
+        var dispatchVisitFloat = function (typeVisitor, args) {
+            var typeRef = args[0].typeRef;
+            var sizeOfFloat = Module.eggShell.typeTopAQSize(typeRef);
+            var visitFn;
```
cglzaguilar: Consider renaming visitFn to functionToVisit

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+            switch (sizeOfFloat) {
+            case 4:
+                visitFn = typeVisitor.visitSingle;
+                break;
+            case 8:
+                visitFn = typeVisitor.visitDouble;
+                break;
+            default:
+                throw new Error('Unexpected size for a Float value');
```
cglzaguilar: Add the float size as part of the error message that triggered the exception.

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+                visitFn;
+            switch (sizeOfComplex) {
+            case 8:
+                visitFn = typeVisitor.visitComplexSingle;
+                break;
+            case 16:
+                visitFn = typeVisitor.visitComplexDouble;
+                break;
+            default:
+                throw new Error('Unexpected size for a Complex value');
```
cglzaguilar: Add the complex size as part of the error message that triggered the exception.

source/io/module_eggShell.js
```diff
@@ -239,6 +246,257 @@
...
+            if (typeof typeVisitor !== 'object' || Array.isArray(typeVisitor)) {
+                throw new Error('typeVisitor must be an object. Found: ' + typeVisitor);
+            }
+
+            var typeRef = valueRef.typeRef,
+                args = Array.prototype.slice.call(arguments, 1),
+                foundTypeIndex = -1,
+                i = 0;
+
+            for (i = 0; i < registeredTypes.length; i += 1) {
```
cglzaguilar: Consider using i++, instead of i += 1.

test-it/karma/publicapi/TypeReflection.Test.js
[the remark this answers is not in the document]
cglzaguilar: Yes, please throw, and let the user implement a visitor function that does nothing if they don't want to deal with the type.

source/core/CEntryPoints.cpp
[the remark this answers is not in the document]
rajsite: Don't we need it because of the STACK_VAR allocation? It has a check for the TypeManagerScope, so it seemed like we should be setting the scope to the same as other files

source/io/module_eggShell.js
rajsite: Currently not allowed by linter due to bad behavior of JS automatic semicolon insertion: https://eslint.org/docs/rules/no-plusplus

Eslint can allow exceptions for this specific case if it is worth changing the rule: https://eslint.org/docs/rules/no-plusplus#allowforloopafterthoughts

source/core/CEntryPoints.cpp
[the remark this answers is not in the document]
rajsite: Talked to Carlos and agrees that we need the TympeManager ref and scope call to allocate the string with STACK_VAR

PR #455, opened 2018-06-21
Implement static event registration for controls in Vireo.

spathiwa:
-Allow declaring ControlReference("controlID#"), inside an EventSpec in the controlID field, automatically:
  1) allocate a unique refnum for the VI and controlID (which will be the same if another event
    on the same control also declares it in a different EventSpec).
  2) Registers and Unregisters for the event in the Vireo EventOracle, and in JS by calling
      JSRegisterForControlEvent() and JSUnRegisterForControlEvent() before and after running
      the top-level VI, respectively.

-Added OccurEvent JS entry point, which given an eventOracleIndex, controlID, and eventType, fires the event to all registered event structures.  It does not yet support passing event-specific event data (e.g. Old/New Value), but does automatically fill in the control refnum.
-Added Vireo and Karma tests.  The Vireo unit test used an internal prim _OccurEvent, to simulate the value change event firing.


test-it/karma/fixtures/events/ValueChangeEventRegistration.via
```diff
@@ -0,0 +1,48 @@
...
+        Printf("Value Changed Event %z %z\n" eventData.Source eventData.Type)
+        Printf("// ControlRef %z\n" eventData.ControlRef)
+        Branch(0)
+        Perch(2)
+        Printf("Timeout Event\n")
+        Perch(0)
+    )
+)))
+
+enqueue(ValueChangedEventRegisterAndUnregister)
```
cglzaguilar: Add another line at the end tp remove the red mark. (No newline at end of file)

test-it/karma/fixtures/events/ValueChangeStaticControlEvent.via
```diff
@@ -0,0 +1,46 @@
...
+        Perch(1)
+        Copy(true eventOccurred)
+        Branch(0)
+        Perch(2)
+        Printf("Timeout Event\n")
+        Copy(true eventTimedOut)
+        Perch(0)
+    )
+)))
+enqueue(UpdateBooleanOnValueChangeEvent)
```
cglzaguilar: Add another line at the end tp remove the red mark. (No newline at end of file)

test-it/karma/fixtures/events/ValueChangeStaticControlEventWithMultipleRegistrations.via
```diff
@@ -0,0 +1,161 @@
...
+        Printf("Value changed\n")
+        Copy(local38 local53)
+        Copy(local53.OldValue local54)
+        Copy(local53.NewValue local55)
+        Copy(true eventStructure2Notified)
+        Branch(0)
+        Perch(0)
+    )
+)))
+enqueue(MultipleEventStructuresListeningToSameControl)
```
cglzaguilar: Add another line at the end tp remove the red mark. (No newline at end of file)

source/core/module_controlEvents.js
```diff
@@ -0,0 +1,89 @@
...
+            viNamePointer,
+            controlId,
+            eventId,
+            eventOracleIndex
+        ) {
+            var viName = Module.eggShell.dataReadString(viNamePointer);
+            try {
+                registerForControlEvent(viName, controlId, eventId, eventOracleIndex);
+            } catch (ex) {
+                // TODO error handling
```
rajsite: Don't think we should swallow exceptions from user code (which in reality is us making really good event registration code that shouldn't crash but if it does we need to know). Makes it impossible to debug.

source/core/module_controlEvents.js
```diff
@@ -0,0 +1,89 @@
...
+            controlId,
+            eventId,
+            eventOracleIndex
+        ) {
+            var viName = Module.eggShell.dataReadString(viNamePointer);
+
+            try {
+                unRegisterForControlEvent(viName, controlId, eventId, eventOracleIndex);
+            } catch (ex) {
+                // TODO error handling
```
rajsite: Same as above, don't try catch. Alternatively rethrow with even more information

source/core/Events.cpp
```diff
@@ -953,20 +1057,24 @@ VIREO_FUNCTION_SIGNATUREV(WaitForEventsAndDispatch, WaitForEventsParamBlock)
...
-                    // When we support multiple events with a shared event case, this needs to be smarter
-                    // and copy only common data, using a custom emit proc (which could verify types) and
-                    // generate custom snippets for each event data node cluster.
-                    eventData.eventDataType->CopyData(eventData.pEventData, (AQBlock1*)esEventDataNode + commonDataSize);
+                if (eventData.eventDataType) {
+                    Int32 alignment = eventData.eventDataType->AQAlignment();
+                    if (commonDataSize % alignment != 0)
+                        commonDataSize = (commonDataSize + alignment-1) & ~(alignment-1);  // round up to alignment boundary
+                    if (eventData.eventDataType->TopAQSize() == dataNodeSize - commonDataSize) {
+                        // TODO(spathiwa) Should we also verify type of event data node matches event at run-time?
```
segaljared: Is this something we need handle now? Or is this already handled and we can remove the TODO?

spathiwa: It still needs to be handled.  Actually, implementing the common data copying will also necessarily have to do this check.  I think I may go ahead and implement it the straight-forward way instead of using a custom emit proc and then we can consider refactoring that way only if performance turns out to be an issue.

source/core/Synchronization.cpp
```diff
@@ -1168,16 +1165,19 @@ DEFINE_VIREO_BEGIN(Synchronization)
     // Dynamic, refnum-based queues
     DEFINE_VIREO_TYPE(QueueRefNum, "refnum(Queue)")
     DEFINE_VIREO_FUNCTION_CUSTOM(ObtainQueue, QueueRef_Obtain,
-        "p(o(QueueRefNum queue)i(Int32 maxsize)i(String name)i(Boolean create)o(Boolean created)io(ErrorCluster err))")
-    DEFINE_VIREO_FUNCTION_CUSTOM(ReleaseQueue, QueueRef_Release, "p(i(QueueRefNum queue)o(String name)o(Array remainingElems)io(ErrorCluster err))")
+        "p(i(StaticTypeExplicitData)o(QueueRefNum queue)i(Int32 maxsize)i(String name)i(Boolean create)o(Boolean created)io(ErrorCluster err))")
```
sanmut: I will talk to you about "StaticTypeExplicitData"

source/core/module_controlEvents.js
```diff
@@ -0,0 +1,78 @@
...
+    /* global Map */
+
+    // Static Private Variables (all vireo instances)
+
+    var assignControlEvents = function (Module, publicAPI) {
+        Module.controlEvents = {};
+        publicAPI.controlEvents = {};
+
+        var registerForControlEvent = function () {
+            // Dummy no-op function.
```
rajsite: Maybe these functions should throw and say an implementation is needed for event handling

source/io/module_eggShell.js
```diff
@@ -679,6 +681,8 @@
         };
 
         Module.eggShell.setOccurrence = Occurrence_Set;
+
+        Module.eggShell.occurEvent = publicAPI.eggShell.occurEvent = OccurEvent;
```
rajsite: I think we should expose this from Module.controlEvents in the module_controlEvents.js file instead.

source/core/Events.cpp
```diff
@@ -953,20 +1057,24 @@ VIREO_FUNCTION_SIGNATUREV(WaitForEventsAndDispatch, WaitForEventsParamBlock)
...
-
-                if (eventData.eventDataType->TopAQSize() == dataNodeSize - commonDataSize) {
-                    // TODO(spathiwa) Should we also verify type of event data node matches event at run-time?
-                    // When we support multiple events with a shared event case, this needs to be smarter
-                    // and copy only common data, using a custom emit proc (which could verify types) and
-                    // generate custom snippets for each event data node cluster.
-                    eventData.eventDataType->CopyData(eventData.pEventData, (AQBlock1*)esEventDataNode + commonDataSize);
+                if (eventData.eventDataType) {
+                    Int32 alignment = eventData.eventDataType->AQAlignment();
+                    if (commonDataSize % alignment != 0)
```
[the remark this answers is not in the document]
spathiwa: I was concerned about the function call in performance critical code, but now that you mention it, I think I'll make AlignAQOffset an inline function and then use it here.

source/core/ControlRef.cpp
```diff
@@ -39,6 +39,18 @@ class ControlRefNumManager : public RefNumManager {
  public:
     static ControlRefNumManager &ControlRefManager() { return _s_singleton; }
     static ControlRefNumType &RefNumStorage() { return _s_singleton.RefNumManager(); }
+    RefNum FindOrCreateRefNum(ControlRefInfo *info) {
+        ControlRefNumType::RefNumIterator ctlRefIter = _refStorage.Begin(), ctlRefEnd = _refStorage.End();
+        while (ctlRefIter != ctlRefEnd) {
+            if (ctlRefIter->second._refData.vi == info->vi && ctlRefIter->second._refData.controlTag->IsEqual(info->controlTag)) {
+                // Found an existing ref with the same data; return it, and dispose the new StringRef, because we now own it.
+                info->controlTag->Delete(info->controlTag);
```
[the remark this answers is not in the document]
spathiwa: Sorry, I disagree -- controlTag is a StringRef but the static method is in the TypedArray base class; I think accessing the method via the variable type is more robust to changes in the class hierarchy.  What's confusing about it?

source/core/Synchronization.cpp
```diff
@@ -1168,16 +1165,19 @@ DEFINE_VIREO_BEGIN(Synchronization)
...
+    DEFINE_VIREO_FUNCTION_CUSTOM(ReleaseQueue, QueueRef_Release, "p(i(StaticTypeExplicitData)i(QueueRefNum queue)o(String name)"
+                                 "o(Array remainingElems)io(ErrorCluster err))")
     DEFINE_VIREO_FUNCTION_CUSTOM(Enqueue, QueueRef_Enqueue,
-        "p(io(QueueRefNum queue)i(* element)i(Int32 timeOut)o(Boolean timedOut)io(ErrorCluster err))")
+        "p(i(StaticTypeExplicitData)io(QueueRefNum queue)i(* element)i(Int32 timeOut)o(Boolean timedOut)io(ErrorCluster err))")
```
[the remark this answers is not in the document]
spathiwa: I'll do so, but keep in mind these strings are actually embedded in the exe/vireo.js file and parsed at launch, so I'm not sure it's a good trade-off to do that for a very minor increase in readability.  It would be nice if these prim-defining string could be stored in an out-board file and discarded after parsing.

source/core/TypeAndDataManager.cpp
```diff
@@ -1973,12 +1973,12 @@ RefNumValType::RefNumValType(TypeManagerRef typeManager, TypeRef type)
     _isFlat = true;
     _hasCustomDefault = true;
     _topAQSize = sizeof(RefNumVal);
+    _aqAlignment = sizeof(RefNumVal) > 4 ? 8 : 4;
```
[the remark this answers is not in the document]
spathiwa: Added comment to explain, and use alignof(RefNumVal) instead of this since we can use C++11 now.
(Without this, it used the wrappedType's alignment, which is wrong, since we don't store that directly, only via cookie.)

source/include/TypeAndDataManager.h
```diff
@@ -933,19 +933,13 @@ typedef UInt32 RefNum;
 
 class RefNumVal {
```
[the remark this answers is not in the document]
spathiwa: It still differentiates the uses of refnum's as allocated by the TypeManager for a VIA file versus simple refnums used internally, so I'd rather keep it just for that, and it makes it easy to add fields if we need to do in the future, or inherit from it to do so for a specialized refnum (I almost did this for Queue Refnums instead of adding STAD args to the prims, but decided against it.)

source/include/TypeAndDataManager.h
```diff
@@ -1217,6 +1211,9 @@ class String : public TypedArray1D< Utf8Char >
     void InsertSubString(IntIndex position, SubString* str) {
         Insert(position, (IntIndex)str->Length(), (Utf8Char*)str->Begin());
     }
+    Boolean IsEqual(String *rhs) {
```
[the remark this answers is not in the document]
spathiwa: Unfortunately, class String and underlying TypedArray are not const-correct, so it was not possible to add operator== with the right argument types without going through a deep rabbit hole to fix that. (You can't even use a const String & or const String * type for rhs).

The fact that DimensionLengths() returns a pointer to internal dim length data is a big problem.

test-it/ViaTests/ValueChangeStaticControlEvent1.via
```diff
@@ -0,0 +1,79 @@
...
+    )
+        clump(1
+	// Note:  No explicit register is generated here, unlike the JavaScriptRefNum static version.
+	// The register is done automatically; before the VI’s first clump runs (or at parse time), we call:
+	//  jsRegisterForControlEvent(StringRef viName, UInt32 controlID, UInt32 eventType, UInt32 eventOracleIdx);
+	// Since the controlRefnum is a Vireo-side reference, OccurEvent will look up the refnum from the eventOracleIdx
+	// and put it in the EventData; the JS code which calls OccurEvent does not need to supply it.
+
+        _OccurEvent(18 1000 2)  // for testing
+        // TODO(spathiwa) the 1 is a hard-coded eventOracle index -- fix to lookup from controlID
```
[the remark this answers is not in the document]
spathiwa: Oops, yeah, I fixed that but forgot to remove the TODO.

source/core/Synchronization.cpp
spathiwa: Okeydoke.

source/io/module_eggShell.js
spathiwa: I'm not sure; the OccurEvents entry point is generic and will also be used to fire events on VIs (mouse down, move, etc.) when they are supported, and could also be used to fire User Events from JS if someone wanted to do that.

source/core/module_controlEvents.js
rajsite: Sounds reasonable to me.
It can also provide errors for someone registering the functions late. ie creating a vireo instance, calling via load, registering functions, calling execute and wondering why the functions never fire.

source/io/module_eggShell.js
rajsite: Then maybe the module should be renamed to something more generic, module_eventHelpers or something

source/core/module_controlEvents.js
spathiwa: Ok, done.

source/io/module_eggShell.js
spathiwa: I'll leave this to Sid or Jared if they want to make this change, as they wrote the JS portions of this PR.

source/include/TypeAndDataManager.h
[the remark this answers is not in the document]
spathiwa: Filed issue #458 to track that.

source/core/RefNum.cpp
```diff
@@ -94,6 +94,11 @@ RefNum RefNumStorageBase::CloneRefNum(RefNum refnum) {
     return newRefNum;
 }
 
+RefNum RefNumStorageBase::RefNumFromIndexAndExistingHeader(UInt32 index, const RefNumCommonHeader *header) {
```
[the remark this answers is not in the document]
spathiwa: Ok, fixed.

source/io/module_eggShell.js
rajsite: Just so they are aware @siddhukrs @segaljared

source/core/module_controlEvents.js
spathiwa: I'm getting an 'unexpected empty function' from eslint.  Can you tell me how to fix?
(Also, I can't run eslint locally; npm run lint hangs indefinitely for me and I don't know how to debug, so I have to rely on CI, so making any JS changes is really tedious for me.)

test-it/test.js
```diff
@@ -238,6 +238,14 @@
         vireo.eggShell.setPrintFunction(function (text) {
             testOutput += text + '\n';
         });
+        // Vireo unit tests can't fire JS events, so register no-op registration functions
+        vireo.controlEvents.setRegisterForControlEventsFunction(function () {
```
gleonoliva: I think that if we change the internal functions in `module_controlEvents` to not throw, then we wouldn't need to write these lines and avoid the linting rules. Unless we want our clients to explicitly opt-out of this functionality, as it is right now, then the other question would be; do we want the same opt-out behavior for the property node?

source/core/module_controlEvents.js
rajsite: You can try running `npm install` locally in-case your dependencies are out of date.
There are also eslint plugins for vscode and visual studio that are pretty helpful during development.

You can drop a comment in a function to describe why it is being left blank like the `// Dummy no-op function.` above.

test-it/test.js
```diff
@@ -240,8 +240,11 @@
         });
         // Vireo unit tests can't fire JS events, so register no-op registration functions
         vireo.controlEvents.setRegisterForControlEventsFunction(function () {
+            // No-op.  Wait, didn't you just say above this function was a no-op?
+            // Why yes, yes I did, but this comment is here to defeat stupid eslint rules.
```
rajsite: The tool is only as smart as its contributors! You can see the rule definition here: https://eslint.org/docs/rules/no-empty-function

Maybe make a pull request to add an option to have line before comments as an additional option. That would be pretty cool :)

Hopefully we can get your local dev environment set-up so you don't have to got back and forth with the CI for running lint tests. That really is a silly and inefficient way to run your linter :P

rajsite: Also just for future reference the linter gives a handy token in the error message: https://travis-ci.org/ni/VireoSDK/builds/396619087#L833

like `no-empty-function` which makes it pretty quick to look-up and see a description of the rules 👍

source/core/ControlRef.cpp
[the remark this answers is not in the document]
sanmut: The way TypedArrayCore and String are implemented now, I don't understand how accessing a static method using an object gives any robustness to changes in the class hierarchy. 

If the intent is to keep Delete() as static in TypedArrayCore and have another static/non-static method by the same name "Delete()" in the derived class, that would be a strong code smell.

OTOH, derived class itself can be passed as a template argument to base class and the base class method can in turn call the derived class static implementation. Though it would work, unless there is a strong reason not to make the method in base class virtual, it is not likely a good idea to do (and nothing of that sort is done here anyway currently).

In C# (not sure but likely Java too), it is a compiler error to access a static method using object.

source/core/module_controlEvents.js
spathiwa: Turns out it was taking forever not because it was misconfigured but because I had a webvi saved in my test-it directory I had been debugging at some point, and it had a ni-webvi-resource-v0 folder containing a large number of js files.

I think make rules should NEVER use wildcards to descend folders but always refer to explicit file lists, so  'eslint .' should be reworked.

source/core/ControlRef.cpp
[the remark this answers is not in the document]
spathiwa: It's more robust because it uses the (declared, not virtual) type of the class to locate the static method to call.  If String was refactored to have a more efficient implementation which did not depend on TypedArrayCore but still conformed to the same API, this call would not need to change.
This is not C#.

source/include/VirtualInstrument.h
```diff
@@ -42,10 +44,31 @@ class VIClump;
...
+};
+
+struct EventControlInfo {
+    EventOracleIndex eventOracleIndex;
+    RefNum refnum;
+    explicit EventControlInfo(EventOracleIndex eoIdx = 0, RefNum ref = 0) : eventOracleIndex(eoIdx), refnum(ref) { }
+};
+
+struct EventInfo {
+    typedef std::map<EventControlUID, EventControlInfo> ControlIDInfoMap;
```
sanmut: Hi Craig, Can we change EventControlUID from Integer to string? ControlRefInfo.ControlTag is already a string. Also, there are plans to change the ControlID in source model to be GUIDs / 64-bit integers.

spathiwa: We talked about this in person.  Instead of resolving the controlRef to a controlID in the EventSpec storage, we'll leave it as a controlRef and match on that. The controlID/GUID string will stay in the RefNumStorage as a string, which it already is.
We will remove the controlID from the OccurEvent JS CEntryPoint, since it can be looked up from the eventOracleIdx and we don't want to have to pass a string.

This will happen in a follow-on PR and doesn't need to block this merge.

PR #456, opened 2018-06-22
Implementation of findSubValueRef

gleonoliva:
Takes an existing ValueRef as the starting point and looks up and returns a new ValueRef corresponding to the described path from the starting point.

### DIFF
--- a/source/core/CEntryPoints.cpp
+++ b/source/core/CEntryPoints.cpp
@@ -132,6 +132,9 @@ VIREO_EXPORT EggShellResult EggShell_FindValue(TypeManagerRef tm, const char* vi
 //! Get a reference to the type pointer and data for a sub element
 VIREO_EXPORT EggShellResult EggShell_FindSubValue(TypeManagerRef tm, TypeRef typeRef, const char* eltName, TypeRef* typeRefLocation, void** dataRefLocation)
 {
+    if (typeRef == nullptr || !typeRef->IsValid() || tm != typeRef->TheTypeManager())
+        return kEggShellResult_UnexpectedObjectType;
+
     SubString path(eltName);
     void* pData = typeRef->Begin(kPARead);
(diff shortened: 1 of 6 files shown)


source/core/CEntryPoints.cpp
```diff
@@ -129,6 +129,21 @@ VIREO_EXPORT EggShellResult EggShell_FindValue(TypeManagerRef tm, const char* vi
     return kEggShellResult_Success;
 }
 //------------------------------------------------------------
+//! Get a reference to the type pointer and data for a sub element
+VIREO_EXPORT EggShellResult EggShell_FindSubValue(TypeManagerRef tm, TypeRef typeRef, const char* eltName, TypeRef* typeRefLocation, void** dataRefLocation)
+{
+    if (typeRef == nullptr || !typeRef->IsValid() || tm != typeRef->TheTypeManager())
+        return kEggShellResult_UnexpectedObjectType;
```
rajsite: For the first two cases, UnexpectedObjectType is for when you only accept booleans but a string was passed. Maybe instead we can add a kEggShellResult_InvalidTypeRef.

Instead of the last case we should be calling `TypeManagerScope scope(tm);` with the passed in TypeManagerRef. i.e. we don't need to assert that the typemanager passed is the current type manager, we need to set the current type manager to the one passed in.

gleonoliva: Yes! I'll add the InvalidTypeRef. Regarding the TypeManagerScope that makes way more sense, I'll change it.

test-it/karma/publicapi/DoubleInAggregate.Test.js
```diff
@@ -0,0 +1,75 @@
+describe('The Vireo EggShell Double api can', function () {
```
rajsite: Change the top-level describe for the new test category

PR #460, opened 2018-06-26
Style cleanup and remove unused parameter

mehagar:


source/core/Array.cpp
```diff
@@ -1014,8 +1014,15 @@ InstructionCore* EmitSortInstruction(ClumpParseState* pInstructionBuilder)
 
     ClumpParseState snippetBuilder(pInstructionBuilder);
     pInstructionBuilder->BeginEmitSubSnippet(&snippetBuilder, sortOp, snippetArgId);
-    snippetBuilder.EmitInstruction(&LTName, 3, elementType, (void*)nullptr, elementType,
-                                   (void*)nullptr, booleanType, (void*)nullptr);
+    snippetBuilder.EmitInstruction(
+        &LTName,
```
spathiwa: I don't think this is an improvement.

source/core/Array.cpp
```diff
@@ -1014,8 +1014,15 @@ InstructionCore* EmitSortInstruction(ClumpParseState* pInstructionBuilder)
...
+        3,
+        elementType,
```
spathiwa: Ditto.

source/core/GenericFunctions.cpp
```diff
@@ -31,25 +31,38 @@ ConstCStr CopyProcName(void *pSource, void *pDest, TypeRef elemType)
...
-    switch (aqSize) {  // copy sizes go to 32 bytes just for efficiency
-        case 1:     return "Copy1";   break;
-        case 2:     return "Copy2";   break;
-        case 4:     return "Copy4";   break;
-        case 8:     return "Copy8";   break;
-        case 16:    return "Copy16";  break;
-        case 32:    return "Copy32";  break;
-        default:    return nullptr;      break;
+
+    switch (aqSize) {  // copy sizes go to 32 bytes for efficiency
```
spathiwa: I think the original was compact but still readable.  OK with removing the breaks.

source/core/GenericFunctions.cpp
```diff
@@ -31,25 +31,38 @@ ConstCStr CopyProcName(void *pSource, void *pDest, TypeRef elemType)
...
-{
-    switch (aqSize) {  // these are bytes not bits.  Enums are max UInt64 sized.
-        case 1:     return "CopyEnum1";   break;
-        case 2:     return "CopyEnum2";   break;
-        case 4:     return "CopyEnum4";   break;
-        case 8:     return "CopyEnum8";   break;
-        default:    return nullptr;      break;
+ConstCStr GetCopyEnumOpName(Int32 aqSize)
+{
+    switch (aqSize) {  // these are bytes not bits. Enums are max UInt64 sized.
```
spathiwa: Ditto.

source/core/GenericFunctions.cpp
```diff
@@ -67,22 +79,23 @@ InstructionCore* EmitGenericCopyInstruction(ClumpParseState* pInstructionBuilder
...
+    InstructionCore* pInstruction = nullptr;
     if (sourceType->IsA(destType, true) || destType->IsA(sourceType, true)) {
         void* pSource = pInstructionBuilder->_argPointers[0];
         void* pDest = pInstructionBuilder->_argPointers[1];
         void* extraParam = nullptr;
         ConstCStr copyOpName = nullptr;
         if (sourceType->IsFlat() || originalCopyOp.CompareCStr("CopyTop")) {
             if (destType->IsEnum()) {
-                copyOpName = CopyEnumProcName(sourceType->TopAQSize());
+                copyOpName = GetCopyEnumOpName(sourceType->TopAQSize());
```
spathiwa: Good change.

source/core/GenericFunctions.cpp
```diff
@@ -304,15 +333,32 @@ InstructionCore* EmitGenericBinOpInstruction(ClumpParseState* pInstructionBuilde
...
-                                                            destType, destArg, destType, destArg);
+                        pInstructionBuilder->EmitInstruction(
+                            &operationName,
+                            3,
+                            sourceXType,
+                            savedArg,
+                            destType,
+                            destArg,
+                            destType,
+                            destArg);
```
spathiwa: How about:

`pInstructionBuilder->EmitInstruction(&savedOperation, 3, `
`        sourceXType, savedArg,  `
`        destType, destArg,  `
`        destType, destArg);  `
Keeping types and args they go with together, at least.

source/core/GenericFunctions.cpp
```diff
@@ -1048,9 +1146,23 @@ InstructionCore* EmitGenericInRangeAndCoerceInstruction(ClumpParseState* pInstru
...
+                hiEltType,
+                nullptr,
+                booleanType,
+                nullptr,
+                booleanType,
+                nullptr,
+                coercedEltType,
+                nullptr,
+                destEltType,
+                nullptr);
```
spathiwa: Oh, c'mon.
The surrounding code is somewhat complex, and it's much more beneficial to be able to see it all on the screen at once then to break these out on new lines  What is the point?

spathiwa:
Please remove all of the changes that break parameters into one line per parameter.
That is not a style we have adopted, or that would I approve of.

source/core/GenericFunctions.cpp
mehagar: I see what you mean. If this function signature ever changed we would have to go back and updated all of the callers just for spacing, so I'm just undoing this for now.

PR #462, opened 2018-06-26
Use control id instead of data item id for variable names

mehagar:
I am running a lvjs now with this change and my NXG changes.


mehagar:
@spathiwa , I don't think so. If all we have is the .via file then there is no chance of getting a control id mixed up with a data item id, because at that piont it's just a string, right?

spathiwa:
The strings are still misleading, saying they are dataItems when they are not, no?

rajsite:
@cglzaguilar is ooo this week, overriding merge

PR #463, opened 2018-06-26
Read value ref object

gleonoliva:
Implementation of function readValueRefObject.

This function would be used for getting all different fields and valueRefs in a Cluster or Cluster-like structures (AnalogWaveform, Complex, Timestamp)

### DIFF
--- a/source/core/module_typeHelpers.js
+++ b/source/core/module_typeHelpers.js
@@ -293,6 +293,8 @@
                 dispatcher: dispatchVisitArray
             },
             {
+                // Cluster is evaluated last, since internally Complex, AnalogWaveform and Timestamps
+                // are also clusters.
                 typeChecker: Module.typeHelpers.isCluster,
                 dispatcher: dispatchVisitCluster
             }
--- a/source/core/module_typeHelpers.js
+++ b/source/core/module_typeHelpers.js
@@ -25,11 +25,13 @@
     var assignTypeHelpers = function (Module) {
         // Disable new-cap for the cwrap functions so the names can be the same in C and JS
         /* eslint 'new-cap': ['error', {'capIsNewExceptions': [
-            'TypeRef_Name'
+            'TypeRef_Name',
+            'TypeRef_ElementName'
         ]}], */
         Module.typeHelpers = {};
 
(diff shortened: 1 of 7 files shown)


source/include/CEntryPoints.h
```diff
@@ -76,6 +76,7 @@ VIREO_EXPORT Boolean TypeRef_HasCustomDefault(TypeRef typeRef);
 VIREO_EXPORT EncodingEnum TypeRef_BitEncoding(TypeRef typeRef);
 VIREO_EXPORT Int32 TypeRef_Alignment(TypeRef typeRef);
 VIREO_EXPORT const char* TypeRef_Name(TypeManagerRef typeManager, TypeRef typeRef);
+VIREO_EXPORT const char* TypeRef_ElementName(TypeManagerRef typeManager, TypeRef typeRef);
```
gleonoliva: I wonder if we should change the signature of these to return an `EggShellResult` and stop using `cwrap`ing them.

rajsite: We can avoid cwrapping them after we enable reading SubString support so we aren't quite there yet

source/core/module_typeHelpers.js
```diff
@@ -287,6 +293,8 @@
                 dispatcher: dispatchVisitArray
             },
             {
+                // Cluster is evaluated last, since internally Complex, AnalogWaveform and Timestamps
```
rajsite: We should add comments for the other types above where order is significant like checking String before Array, etc

test-it/karma/publicapi/ReadValueRefObject.Test.js
```diff
@@ -0,0 +1,133 @@
...
+
+        it('returns an object with keys "real" and "imaginary" and valueRefs as values', function () {
+            var objectValueRef = vireo.eggShell.readValueRefObject(valueRef);
+            expectValidValueRef(objectValueRef.real);
+            expectValidValueRef(objectValueRef.imaginary);
+        });
+    });
+
+    it('throws for any other type', function () {
+        var tryReadValueRefObject = function (path) {
```
rajsite: nice helper 👍

PR #467, opened 2018-06-27
Refactored readJSON and writeJSON

gleonoliva:


source/core/CEntryPoints.cpp
```diff
@@ -168,39 +168,33 @@ VIREO_EXPORT NIError EggShell_ReadDouble(TypeManagerRef tm, const TypeRef actual
...
-    if (actualType == nullptr)
-        return;
+    if (typeRef == nullptr || !typeRef->IsValid())
+        return kEggShellResult_InvalidTypeRef;
 
     EventLog log(EventLog::DevNull);
     SubString formatss(format);
     TDViaParser parser(tm, &valueString, &log, 1, &formatss, true, true, true);
-    parser.ParseData(actualType, pData);
+    Int32 error = parser.ParseData(typeRef, data);
```
rajsite: ParseData gives some pretty good error codes back in the form of LVError + JSONLVError: https://github.com/ni/VireoSDK/blob/master/source/core/TDCodecVia.cpp#L1079

Maybe we should pass that through like how readDouble returns NIError instead of EggShellResult: https://github.com/ni/VireoSDK/blob/api-type-refactor/source/io/module_eggShell.js#L296

Alternatively maybe readDouble should switch to returning EggShellResult. Thoughts?

gleonoliva: I agree with switching readDouble and writeDouble to EggShellResult. I think that we can use NIError internally for vireo. If we think is needed to preserve more context about the error, we could add more EggShellResult options (which I don't think is needed right now)

rajsite: Sounds good, I'll do that as part of the writeDouble change

cglzaguilar: Agree. Make the API the same. Use EggShellResult consistently. If we need more specific information add error codes to EggShellResult that are more specific.

PR #469, opened 2018-06-28
Aded visitEnum[8|16|32] to visitor reflection API

gleonoliva:


gleonoliva:
To be honest I still think that the functions should be named `visitEnum8`, `visitEnum16` and `visitEnum32` just because they are shorter.

rajsite:
Two minor comments: go for enum8,16,32 and add a test to verify it fails on unsupported enum sizes

PR #474, opened 2018-06-30
 Minor refactor of event matching so static control refs match on the ref

spathiwa:
...and not the controlID, so this doesn't have to change if the controlID
becomes a larger GUID (64-bit or string).
    
This does not yet remove the controlID from OccurEvent or
jsRegister/UnregisterControlEvent; we can't do that until we're actually ready to
change the type to something else throughout the stack;
jsRegister/UnregisterControlEvent does need to pass the controlID to
associate it with the eventOracleIndex, even if it's a string or larger GUID.

### DIFF
--- a/source/core/Events.cpp
+++ b/source/core/Events.cpp
@@ -324,6 +324,7 @@ bool EventOracle::RegisterForEvent(EventQueueID qID, EventSource eSource, EventT
         }
         if (idx < size) {  // we found an unused index, or one already for this control ID
             _eventReg[idx]._controlUID = controlUID;
+            _eventReg[idx]._controlRef = ref;
         } else {
             EventOracleObj eo(controlUID, ref);
             _eventReg.push_back(eo);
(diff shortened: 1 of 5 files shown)


source/core/Events.cpp
```diff
@@ -568,7 +569,6 @@ void RegisterForStaticEvents(VirtualInstrument *vi) {
                     viName->AppendSubString(&encodedSubstr);
                     char *tagBegin = reinterpret_cast<char*>(tag->Begin()), *tagEnd = reinterpret_cast<char*>(tag->End());
                     EventControlUID controlID = Int32(strtol(tagBegin, &tagEnd, 10));
```
sanmut: > EventControlUID controlID = Int32(strtol(tagBegin, &tagEnd, 10));

1) I thought we should not parse control ID string to an integer as it can be GUID ?
2) EventControlUID should not be an integer too.
3) Why not use the term controlTag as used in ControlRefInfo (instead of UID) as we are referring to the same.  Or, change controlTag in ControlRefInfo also.

spathiwa: 1 and 2. I can't remove that until jsRegisterForControlEvent is refactored to use a larger (or string) GUID.  When that happens, we just change the argument typeIt's only remembered to pass for that argument though.
3. I'l get rid of EventControlUID altogether, rather than renaming it, and just use the controlTag in the refnum, when the control ID->GUID change happens.

rajsite:
Have we run an LVjs with this change? Does it require DFIR changes?

PR #475, opened 2018-07-02
Fix bug in ArrayConcatenate caused by not copying data in reverse order.

spathiwa:
Also fixes leaks caused by not deflating data before copying over it.


source/core/GenericFunctions.cpp
```diff
@@ -1614,11 +1614,27 @@ AQBlock1* ArrayToArrayCopyHelper(TypeRef elementType, AQBlock1* pDest, IntIndex*
...
-            return nullptr;
-        Int32 copiedLength = sourceDimLengths[0] * elementType->TopAQSize();
-        if (copiedLength < destSlabLengths[1]) {
-            memset(pDest + copiedLength, 0, destSlabLengths[1] - copiedLength);
+        Int32 topAQSize = elementType->TopAQSize();
+        Int32 copyElems = sourceDimLengths[0];
+        Int32 copyLength = copyElems * elementType->TopAQSize();
+        pDest += copyLength;
+        pSource += copyLength;
+        while (copyElems > 0) {
```
mehagar: Why not use a for loop that runs copyElems times?

spathiwa: Because this is exactly equivalent, and the initialization section of the for loop would be empty (or have two statements with a comma operator in it).

source/core/GenericFunctions.cpp
```diff
@@ -1614,11 +1614,27 @@ AQBlock1* ArrayToArrayCopyHelper(TypeRef elementType, AQBlock1* pDest, IntIndex*
...
     AQBlock1 *pSource, IntIndex* sourceDimLengths, IntIndex* sourceSlabLengths, Int32 destRank, Int32 sourceRank) {
     if (sourceRank == 1) {
-        if (elementType->CopyData(pSource, pDest, sourceDimLengths[0]))
-            return nullptr;
-        Int32 copiedLength = sourceDimLengths[0] * elementType->TopAQSize();
-        if (copiedLength < destSlabLengths[1]) {
-            memset(pDest + copiedLength, 0, destSlabLengths[1] - copiedLength);
+        Int32 topAQSize = elementType->TopAQSize();
+        Int32 copyElems = sourceDimLengths[0];
+        Int32 copyLength = copyElems * elementType->TopAQSize();
```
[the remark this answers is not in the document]
spathiwa: Yeah, I meant to, missed the edit.  Fixed.

test-it/ViaTests/ArrayConcatBug2D.via
```diff
@@ -0,0 +1,91 @@
+// Autogenerated Vireo assembly file
+// Generated on 07/02/2018 11:27:51
+//---------------------------------------------------
```
ggumen: It would be nice to have the CAR number here, so we can track this ArrayConcatBug2D.via to the source of issue.

spathiwa: Ok, I went ahead and mentioned the CAR in the test comment.  I guess it doesn't hurt anything, but isn't useful to anyone outside NI.

ggumen: I thought the issue was due to the 2D array having elements of different sizes where the last row had more elements than the previous rows or something like that. I think that would be useful to have in the description. And you can drop the CAR number.

PR #476, opened 2018-07-11
Fix ScanValue for empty format string

mehagar:
CG seemed to be doing the equivalent of auto formatting for an empty string, and the documentation for CG doesn't explicitly mention and empty format string. Does anyone know for sure if that's the case?


source/core/NumericString.cpp
```diff
@@ -1691,10 +1700,9 @@ Int32 FormatScan(SubString *input, SubString *format, Int32 argCount, StaticType
                 case '%': {    //%%
                     input->ReadRawChar(&inputChar);
                     ++offsetPastScan;
-                    if (inputChar == '%') {
-                        } else {
-                            canScan = false;
-                        }
+                    if (inputChar != '%') {
```
mehagar: This is just a refactoring.

source/core/NumericString.cpp
```diff
@@ -1603,6 +1603,15 @@ Int32 FormatScan(SubString *input, SubString *format, Int32 argCount, StaticType
     SubString f(format);
     const Utf8Char* origBegin = input->Begin();
     UInt32 offsetPastScan = 0;
+
+    // if the format string is empty, use the auto format option (%g)
+    SubString withoutWhitespaces(&f);
+    withoutWhitespaces.EatLeadingSpaces();
```
sanmut: It looks like NXG uses %g only if the format string is an empty string. If there are only whitespaces in the format string, NXG appears to return the input string in remainder string. Is this different from CG? You should add a couple of test cases with format string just with an empty string and just with whitespaces and compare output value and remainder string in both cases and test it against Desktop in NXG and CG.

I am not a big fan of this prim without an error out and does something undocumented. However, if we are fixing, we might fix it to achieve parity with other platforms.

sanmut: Also, the same function being modified is being used by "Scan Value" and "Scan From String" primitives. I would recommend writing VI tests that use both for similar cases to make sure fixing one does not break the other.

mehagar: That behavior you described about when there are whitespaces is correct for both NXG and CG. I'll make it consistent here (not strip whitespaces) and add a test.

PR #477, opened 2018-07-12
Remove warning

mehagar:
This warning did occur when you had an array with a default value of null, as in when you create an array control. This is not a warning for desktop, so it should not be one for Vireo.


source/core/TDCodecVia.cpp
```diff
@@ -959,7 +959,7 @@ Int32 TDViaParser::ParseArrayData(TypedArrayCoreRef pArray, void* pFirstEltInSli
                 _string.EatChar(Fmt()._itemSeparator);
                 void* pElement = elementCount < length ? pEltData : nullptr;
                 Int32 subErr;
-                if (pElement == nullptr) {
+                if (elementCount >= length) {
```
mehagar: What was happening is that pEltData was also nullptr for empty arrays, so even if elementCount was < length, we were treating it as if there were extra initializers. I change the check here which only occurs if there really are extra initializers and not just an empty array.

test-it/ExpectedResults/MultiDimensionArrayDefaults.vtr
```diff
@@ -1,18 +1,12 @@
 (Line 4 Warning "Ignoring extra array initializer elements.")
 (Line 4 Warning "Ignoring extra array initializer elements.")
```
sanmut: Why are we printing the same warning twice? If you don't fix it, create a github issue and submit this.

mehagar: Good question. I'll create a github issue.

PR #478, opened 2018-07-13
Dynamic variable allocation

gleonoliva:
Exposing 2 functions that will allow JS to allocate variables based on a Type declared in the .via code and of course deallocate the memory.

The workflow goes as follow:
```javascript

// In update service
var valueRef = vireo.eggShell.findValueRef(viName, 'speciallyNamedLocalOrGlobalForEventType'); // Look for a variable
// valueRef contains a pointer to the type and the data for that variable

var writeCallback = function (valueRef) {
    // uses poker to write new data
}
vireo.eventHelpers.occurEvent (valueRef, writeCallback, eventData, other_event_stuff);

// In vireo module_eggShell
publicAPI.eggShell.occurEvent () {
    // allocateData only cares about the type.
    // This will allocate enough memory for that type
    var allocatedDataValueRef = Module.eggShell.allocateData(valueRef.typeRef);
    // Let the update service manipulate the new allocated space
    writeCallback(allocatedDataValueRef)
    // Invoke the actual occur event
    OCCUR_EVENT_C_INTERFACE(stuff);
    // Once done with that data, e.g. after an event occurs, you can deallocate it.
    Module.eggShell.deallocateData(allocatedDataValueRef);
}
```

You can take a look to `AllocateTypes.Test.js` too.

### DIFF
--- a/source/core/CEntryPoints.cpp
+++ b/source/core/CEntryPoints.cpp
@@ -150,6 +150,7 @@ VIREO_EXPORT EggShellResult EggShell_DeallocateData(TypeManagerRef tm, const Typ
     }
 
     NIError error = typeRef->ClearData(dataRef);
+    THREAD_TADM()->Free(dataRef);
     if (error != kNIError_Success) {
         return kEggShellResult_UnableToDeallocateData;
     }
--- a/source/core/CEntryPoints.cpp
+++ b/source/core/CEntryPoints.cpp
@@ -118,10 +118,14 @@ VIREO_EXPORT Int32 EggShell_PokeMemory(TypeManagerRef tm,
     }
 }
 //------------------------------------------------------------
-//! 
+//! Allocates enough memory to fit a new object of TypeRef
 VIREO_EXPORT EggShellResult EggShell_AllocateData(TypeManagerRef tm, const TypeRef typeRef, void** dataRefLocation)
 {
     TypeManagerScope scope(tm);
+    if (typeRef == nullptr || !typeRef->IsValid()) {
+        return kEggShellResult_InvalidTypeRef;
+    }
+
     *dataRefLocation = nullptr;
     Int32 topSize = typeRef->TopAQSize();
     void* pData = THREAD_TADM()->Malloc(topSize);
@@ -133,15 +137,23 @@ VIREO_EXPORT EggShellResult EggShell_AllocateData(TypeManagerRef tm, const TypeR
     return kEggShellResult_Success;
 }
 //------------------------------------------------------------
-//! 
-VIREO_EXPORT EggShellResult EggShell_DeallocateData(TypeManagerRef tm, const TypeRef typeRef, void* dataRefLocation)
+//! Deallocates data and frees up memory in dataRef described by typeRef
+VIREO_EXPORT EggShellResult EggShell_DeallocateData(TypeManagerRef tm, const TypeRef typeRef, void* dataRef)
 {
     TypeManagerScope scope(tm);
+    if (typeRef == nullptr || !typeRef->IsValid()) {
+        return kEggShellResult_InvalidTypeRef;
+    }
 
-    NIError error = typeRef->ClearData(dataRefLocation);
+    if (dataRef == nullptr) {
+        return kEggShellResult_NullDataPointer;
+    }
+
+    NIError error = typeRef->ClearData(dataRef);
     if (error != kNIError_Success) {
         return kEggShellResult_UnableToDeallocateData;
     }
+
     return kEggShellResult_Success;
 }
 //------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


source/core/CEntryPoints.cpp
```diff
@@ -118,6 +118,46 @@ VIREO_EXPORT Int32 EggShell_PokeMemory(TypeManagerRef tm,
...
+    TypeManagerScope scope(tm);
+    if (typeRef == nullptr || !typeRef->IsValid()) {
+        return kEggShellResult_InvalidTypeRef;
+    }
+
+    if (dataRef == nullptr) {
+        return kEggShellResult_InvalidDataPointer;
+    }
+
+    NIError error = typeRef->ClearData(dataRef);
```
mehagar: Do we really need to clear the data before we free it?

test-it/karma/publicapi/AllocateTypes.Test.js
```diff
@@ -0,0 +1,254 @@
...
+        describe('for allocateData throws', function () {
+            it('when typeRef is invalid', function () {
+                var invalidTypeRef = 1234;
+
+                expect(tryAllocate(invalidTypeRef)).toThrowError(/InvalidTypeRef/);
+            });
+        });
+
+        describe('for deallocateData throws', function () {
+            it('when typeRef is invalid', function () {
```
mehagar: From what I understand of this kind of testing framework, the reason the function is called "it" is so the test reads like an english sentence. In that case, please use strings here that read like a sentence, e.g. it('throws an error when given an invalid value ref.')

source/core/CEntryPoints.cpp
spathiwa: Yes.  Freeing only frees the top level, and knows nothing about the type.
ClearData deallocates any "hair" allocations in the type. For example, if a cluster contains a string, it would deallocate the string, and the free would deallocate the cluster top allocaiton.

test-it/karma/publicapi/AllocateTypes.Test.js
gleonoliva: I agree with you. I was more worried about how the whole test name ends up in the output and not so much with the readability of each `it()`. I'll update this. Thanks for pointing it out.

source/core/CEntryPoints.cpp
```diff
@@ -118,6 +118,46 @@ VIREO_EXPORT Int32 EggShell_PokeMemory(TypeManagerRef tm,
...
 }
 //------------------------------------------------------------
+//! Allocates enough memory to fit a new object of TypeRef
+VIREO_EXPORT EggShellResult EggShell_AllocateData(TypeManagerRef tm, const TypeRef typeRef, void** dataRefLocation)
+{
+    TypeManagerScope scope(tm);
+    if (typeRef == nullptr || !typeRef->IsValid()) {
+        return kEggShellResult_InvalidTypeRef;
+    }
+
```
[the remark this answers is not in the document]
gleonoliva: I added a check for that too.

test-it/karma/fixtures/publicapi/AllocateTypes.via
```diff
@@ -0,0 +1,57 @@
...
+
+        e(a(.Boolean 2 2) fixedBooleans)
+        e(.AnalogWaveform<.Double> wave_Double)
+        e(.NIPath nipath)
+        e(dv(Enum16 (zero one two three four) 3) enum16numbers)
+    )
+    clump (1
+    )
+)))
+
```
sanmut: It is a good idea to test globals. We could add a few tests for "ce" in "Locals" as that's what DFIR emits now for constants. It is probably a good idea to add tests for multi-dimensional arrays but can be done later and this need not be blocked.

sanmut:
I like the fact that the JS APIs are not public. Looks good. Thanks.

test-it/karma/fixtures/publicapi/AllocateTypes.via
gleonoliva: I added a few cases for `ce` in locals and a 3D array of single values.

source/core/CEntryPoints.cpp
```diff
@@ -118,6 +118,50 @@ VIREO_EXPORT Int32 EggShell_PokeMemory(TypeManagerRef tm,
...
+    if (dataRefLocation == nullptr) {
+        return kEggShellResult_InvalidDataPointer;
+    }
+
+    *dataRefLocation = nullptr;
+    Int32 topSize = typeRef->TopAQSize();
+    void* pData = THREAD_TADM()->Malloc(topSize);
+    NIError error = typeRef->InitData(pData, (TypeRef)nullptr);
```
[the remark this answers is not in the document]
gleonoliva: I made this change too.

PR #471, opened 2018-06-28
Vireo EggShell Refactor for Supporting More Types

rajsite:
Implementation of "Vireo API Changes for Supporting More Types Design Doc" - [NITalk](https://nitalk.jiveon.com/docs/DOC-483168) - [PDF](https://github.com/ni/VireoSDK/files/2147222/DOC-483168.pdf)

This change encapsulates the breaking API changes associated with the Vireo API refactor. There is still additional work to be done out of master after the change is merged consisting of:

1. Making error checks and error handling consistent across all of the new api functions: https://github.com/ni/VireoSDK/issues/466

2. Migrating the following JavaScript users of the now deprecated API functions to the new API functions: HttpClient, JavaScriptInvoke, Core (FPSync, LV Error Handling), Events.

3. Removing the deprecated API functionality

*LVjs Build*: Still in-progress testing changes in Bajor/Main

PR #479, opened 2018-07-16
Update PropertyNode JS Callbacks

gleonoliva:
After #471 made it to master, the concept of a valueRef was available in our vireo eggshell API. In this branch we take advantage of this new object and modify the property read/write JS callbacks to use this instead of the temporary variable VI name and its path.
This results in:
- Less code, deleted function `FindVINameAndPropertyPathForValue` in PropertyNode.cpp and a few `STACK_VAR`s.
- A couple of parameters less passed to JS, see `module_PropertyNode.js`
- Improved performance since we are no longer reading the temporary variable VIName, Type name and Path as strings every time.

I refactored all related tests too, improving a little bit their readability. 
I still need to run a LVjsCI once I have that we can merge this branch to master.

### DIFF
--- a/source/io/PropertyNode.cpp
+++ b/source/io/PropertyNode.cpp
@@ -29,10 +29,10 @@ enum { kNIError_ObjectReferenceIsInvalid = 1055 };  // TODO(spathiwa) move to co
 #if kVireoOS_emscripten
 extern "C" {
     // JavaScript function prototypes
-    // Parameters: controlRefVIName, controlId, propertyName, propertyTypeName, tempVariableVIName, tempVariablePath, errorStatus*, errorCode*, errorSource*
-    extern void jsPropertyNodeWrite(StringRef, StringRef, StringRef, StringRef, StringRef, StringRef, Boolean *, Int32 *, StringRef);
-    // Parameters: controlRefVIName, controlId, propertyName, propertyTypeName, tempVariableVIName, tempVariablePath, errorStatus*, errorCode*, errorSource*
-    extern void jsPropertyNodeRead(StringRef, StringRef, StringRef, StringRef, StringRef, StringRef, Boolean *, Int32 *, StringRef);
+    // Parameters: controlRefVIName, dataItemName, propertyName, tempVariableType, tempVariableDataPtr, errorStatus*, errorCode*, errorSource*
+    extern void jsPropertyNodeWrite(StringRef, StringRef, StringRef, TypeRef, void *, Boolean *, Int32 *, StringRef);
+    // Parameters: controlRefVIName, dataItemName, propertyName, tempVariableType, tempVariableDataPtr, errorStatus*, errorCode*, errorSource*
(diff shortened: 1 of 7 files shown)


source/io/PropertyNode.cpp
```diff
@@ -29,10 +29,10 @@ enum { kNIError_ObjectReferenceIsInvalid = 1055 };  // TODO(spathiwa) move to co
 #if kVireoOS_emscripten
 extern "C" {
     // JavaScript function prototypes
-    // Parameters: controlRefVIName, controlId, propertyName, propertyTypeName, tempVariableVIName, tempVariablePath, errorStatus*, errorCode*, errorSource*
-    extern void jsPropertyNodeWrite(StringRef, StringRef, StringRef, StringRef, StringRef, StringRef, Boolean *, Int32 *, StringRef);
-    // Parameters: controlRefVIName, controlId, propertyName, propertyTypeName, tempVariableVIName, tempVariablePath, errorStatus*, errorCode*, errorSource*
-    extern void jsPropertyNodeRead(StringRef, StringRef, StringRef, StringRef, StringRef, StringRef, Boolean *, Int32 *, StringRef);
+    // Parameters: controlRefVIName, dataItemName, propertyName, tempVariableType, tempVariableDataPtr, errorStatus*, errorCode*, errorSource*
```
rajsite: accidentally re-introduced dataItemName, change back to controlName

gleonoliva: 😆 good catch! I'll change it back.

source/io/PropertyNode.cpp
```diff
@@ -29,10 +29,10 @@ enum { kNIError_ObjectReferenceIsInvalid = 1055 };  // TODO(spathiwa) move to co
 #if kVireoOS_emscripten
 extern "C" {
     // JavaScript function prototypes
-    // Parameters: controlRefVIName, controlId, propertyName, propertyTypeName, tempVariableVIName, tempVariablePath, errorStatus*, errorCode*, errorSource*
-    extern void jsPropertyNodeWrite(StringRef, StringRef, StringRef, StringRef, StringRef, StringRef, Boolean *, Int32 *, StringRef);
-    // Parameters: controlRefVIName, controlId, propertyName, propertyTypeName, tempVariableVIName, tempVariablePath, errorStatus*, errorCode*, errorSource*
-    extern void jsPropertyNodeRead(StringRef, StringRef, StringRef, StringRef, StringRef, StringRef, Boolean *, Int32 *, StringRef);
+    // Parameters: controlRefVIName, controlId, propertyName, tempVariableType, tempVariableDataPtr, errorStatus*, errorCode*, errorSource*
```
sanmut: Please delete the comment line // Parameters and name the parameters in the function declarations below. Will be much easier to read and maintain.

gleonoliva: I agree. I'll address this in a different PR.

gleonoliva:
A LVjsCI just came back and we are good to go. I'll wait for required reviewers.

PR #481, opened 2018-07-18
Allowing data to be passed into occurEvent for control events

segaljared:
These changes allow the update service to pass event data into occurEvent. I've add the TypeRef and data parameters to the OccurEvent C interface. I've also modified occurEvent method in eventHelpers to utilize the new allocateData and deallocateData methods to create space in Vireo's memory for the event data before using a callback method to write the data into the value reference.

### DIFF
--- a/source/core/module_eventHelpers.js
+++ b/source/core/module_eventHelpers.js
@@ -36,9 +36,6 @@
         var unRegisterForControlEvent = function () {
             throw new Error('No event un-registration callback was supplied');
         };
-        var writeEventData = function () {
-            throw new Error('No event data write callback was supplied');
-        };
 
         Module.eventHelpers.jsRegisterForControlEvent = function (
             viNamePointer,
@@ -76,18 +73,10 @@
             unRegisterForControlEvent = fn;
         };
 
-        publicAPI.eventHelpers.setWriteEventDataFunction = Module.eventHelpers.setWriteEventDataFunction = function (fn) {
-            if (typeof fn !== 'function') {
-                throw new Error('WriteEventData must be a callable function');
-            }
-
-            writeEventData = fn;
-        };
-
-        publicAPI.eventHelpers.occurEvent = Module.eventHelpers.occurEvent = function (eventOracleIndex, controlId, eventType, eventDataTypeValueRef, eventData) {
+        publicAPI.eventHelpers.occurEvent = Module.eventHelpers.occurEvent = function (eventOracleIndex, controlId, eventType, typeVisitor, eventDataTypeValueRef, eventData) {
             // Allocate space for the event data using the type information passed in to occurEvent
             var allocatedDataValueRef = Module.eggShell.allocateData(eventDataTypeValueRef.typeRef);
-            writeEventData(allocatedDataValueRef, eventData);
+            Module.eggShell.reflectOnValueRef(typeVisitor, allocatedDataValueRef, eventData);
             Module._OccurEvent(eventOracleIndex, controlId, eventType, allocatedDataValueRef.typeRef, allocatedDataValueRef.dataRef);
             // Now that the data has been passed to Vireo, which should copy it, deallocate the memory
             Module.eggShell.deallocateData(allocatedDataValueRef);
--- a/source/core/module_eventHelpers.js
+++ b/source/core/module_eventHelpers.js
@@ -73,10 +73,10 @@
             unRegisterForControlEvent = fn;
         };
 
-        publicAPI.eventHelpers.occurEvent = Module.eventHelpers.occurEvent = function (eventOracleIndex, controlId, eventType, typeVisitor, eventDataTypeValueRef, eventData) {
+        publicAPI.eventHelpers.occurEvent = Module.eventHelpers.occurEvent = function (eventOracleIndex, controlId, eventType, writeCallback, eventDataTypeValueRef, eventData) {
             // Allocate space for the event data using the type information passed in to occurEvent
             var allocatedDataValueRef = Module.eggShell.allocateData(eventDataTypeValueRef.typeRef);
-            Module.eggShell.reflectOnValueRef(typeVisitor, allocatedDataValueRef, eventData);
+            writeCallback(allocatedDataValueRef, eventData);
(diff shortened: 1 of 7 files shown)


source/core/module_eventHelpers.js
```diff
@@ -27,22 +27,18 @@
...
 
-        var OccurEvent = Module.cwrap('OccurEvent', 'void', ['number', 'number', 'number']);
-
         var registerForControlEvent = function () {
             throw new Error('No event registration callback was supplied');
         };
         var unRegisterForControlEvent = function () {
             throw new Error('No event un-registration callback was supplied');
         };
+        var writeEventData = function () {
```
segaljared: This callback is used to write the data. In the PR for the dynamic variable allocation, this was a method passed occurEvent, but I believe all of the callbacks would be the same. Any event could just use the Vireo poker to generically write JS objects into the allocated memory (VIREO_POKER.pokeValueRef(vireo, valueRef, eventData). Therefore, I'm making this callback something that can be set once for Vireo like the register and unregister methods.

rajsite: I think it should be something that is specified per occurEvent call. There can be other code paths and event types that may occur an event besides value properties and I don't think we necessarily should force them down a single function registration.

One example is that we have discussed implementing WebSockets using Events which would receive strings over the network. While strings that go through the peeker and poker are always assumed to be JS strings, we can actually get a UTF-8 array buffer directly from a websocket and have a more optimized copy. Forcing one write function means we have to twist it to support all possible use cases for events occurring.

I would actually be in favor of switching from a function being passed to a visitor being passed. That would keep the api interface consistent between other api functions we added like `reflectOnValueRef (typeVisitor, valueRef, data)`

If we were to do that we can change the occurEvent signature to be:
`occurEvent = function (eventOracleIndex, controlId, eventType, typeVisitor, eventDataTypeValueRef, eventData)`

And in the occurEvent function do:
``` 
occurEvent = function {
var allocatedDataValueRef = ...
Module.eggShell.reflectOnValueRef(typeVisitor, valueRef, data);
Module._occurEvent(...)
Module.eggShell.deallocateData(...)
}
```

I say we either switch to passing in a writeFunction every time or passing in a typeVisitor every time, but I'm not for one writefunction / typeVisistor being used for all calls.

test-it/karma/events/ControlEvents.Test.js
```diff
@@ -64,7 +71,12 @@ describe('ValueChanged event tests', function () {
         var viPathParser = vireoRunner.createVIPathParser(vireo, 'UpdateBooleanOnValueChangeEvent');
 
         setTimeout(function () {
```
rajsite: not sure how i missed these setTimeouts but they probably need to be eliminated, definitely look like they can cause race conditions

rajsite: The settimeouts don't need to be addressed as part of this review, seems to be used elsewhere unexpectedly as well

source/core/module_eventHelpers.js
segaljared: I've updated it to take a visitor and use reflectOnValueRef

test-it/karma/events/ControlEvents.Test.js
segaljared: I added a new test that follows the same pattern. Is there a better way to do this other than setTimeout? I could update my new test and the others to follow that.

source/core/module_eventHelpers.js
```diff
@@ -80,7 +73,14 @@
             unRegisterForControlEvent = fn;
         };
 
-        publicAPI.eventHelpers.occurEvent = Module.eventHelpers.occurEvent = OccurEvent;
+        publicAPI.eventHelpers.occurEvent = Module.eventHelpers.occurEvent = function (eventOracleIndex, controlId, eventType, typeVisitor, eventDataTypeValueRef, eventData) {
+            // Allocate space for the event data using the type information passed in to occurEvent
+            var allocatedDataValueRef = Module.eggShell.allocateData(eventDataTypeValueRef.typeRef);
+            Module.eggShell.reflectOnValueRef(typeVisitor, allocatedDataValueRef, eventData);
```
gleonoliva: I thought we were going to go with the writeCallback function.

segaljared: @rajsite suggested using a visitor instead of using a writeCallback.

test-it/karma/events/ControlEvents.Test.js
rajsite: We have a pattern that allows you to run JS immediately after the viaCode is loaded but before clumps have executed and another one after all clumps have executed. I think the first is likely too early (especially longer term when registration will be done with emitted JSLI instructions) and the latter is too late.

We would need to come up with a new pattern that is reliable. I'm thinking it would be polling periodically for a boolean flag that the via code sets and executing some JS after that time.
Or even the via test executes the JS via the JSLI.

But yea we would need to think of a new pattern, I don't think there is any good precedent.

source/core/module_eventHelpers.js
rajsite: Memo, Jared, and I synced up and decided to use a writeCallback(allocatedValueRef) function that is referenced in each invocation of occurEvent

spathiwa:
It looks like you did a merge instead of a rebase and got a lot of duplicate commits here.

I would just create a new branch off of master and cherry-pick your fix commits into it instead of trying to fix it here.

test-it/karma/events/ControlEvents.Test.js
```diff
@@ -119,4 +164,29 @@ describe('ValueChanged event tests', function () {
             done();
         });
     });
+
+    it('Verify event occurrence passes correct old and new value', function (done) {
```
mehagar: Please name this test as per BDD (See the first paragraph here: https://jasmine.github.io/tutorials/your_first_suite)

segaljared:
@cglzaguilar just to be certain: the release built off of these changes will require my VSTS changes in order for them to work, so we'll need to wait on merging them in until the VSTS PR is ready

test-it/karma/events/ControlEvents.Test.js
segaljared: I've updated all the names in the file to reflect the naming convention as best as I could

PR #482, opened 2018-07-18
Fix alignment issues in copying event data.

spathiwa:
If the event data in a user event is, say, a cluster of an Int32 and
Double, then platform alignment causes there to be 4 bytes of padding
between the 2 fields so the Double can be 8-byte aligned.
However, in the event data node cluster on the diagram (and associated
VIA cluster), there are common event data fields (source, type, index,
timestamp, ref) that aren't explicitly part of the user-supplied user
event data.. When the User Event data is appended to this common
data, the resulting cluster does NOT need padding between the last Int32
and the Double, because the Int32 was already 20 bytes offset into the
cluster, so the next field, at 24 bytes offset, is already 8
byte-aligned.  This discrepancy is the source of this bug; the code
assumed you only had to worry about alignment at the beginning of the
copy, and then the clusters could be copied in one operation.

Solution is to copy the clusters one field at a time, in which case we
don't even need to explicitly handle alignment, because cluster types
know their elements' individual offsets. This is less performant than
copying the whole cluster, but we needed to do this eventually anyway
to handle shared event cases between events with some but not all
data elements in common.


source/core/Events.cpp
```diff
@@ -1064,14 +1064,45 @@ VIREO_FUNCTION_SIGNATUREV(WaitForEventsAndDispatch, WaitForEventsParamBlock)
...
+                        // common fields, and also to handle shared event cases which only expose event data common
+                        // to all sources.
+                        // Note, although this looks like an O(N^2) nested loop, it's actually O(N+M), the sum of
+                        // the number of fields in the source (fired event data) and dest (cluster).
+                        while (destElementIndex < numDestElements) {
+                            // (TODO(spathiwa) We could use a custom emit proc to verify types and generate
+                            // custom snippets for each event data node cluster so it doesn't have to check at run-time.)
+                            TypeRef destElemType = esEventDataNodeType->GetSubElement(destElementIndex);
+                            IntIndex destOffset = destElemType->ElementOffset();
+                            while (sourceElementIndex < numSourceElements) {
```
segaljared: Do we always assume that the order of data in the source and destination must be the same? That way we don't have to reset sourceElementIndex before starting this loop?

spathiwa: Yes, this is required. NCG does the same thing.

PR #483, opened 2018-07-31
Update devDependencies to latest

rajsite:
Karma had an indirect dependecy on a package with a known vulnerability described here: https://nvd.nist.gov/vuln/detail/CVE-2017-16113

This update migrates all packages to the latest released semantic compatible version. In addition, the following packages required a major version change to remove the vulberable dependency: jasmine-core, karma, and karma-jasmine-html-reporter.

A minor change was required to the karma vtr test runner to perform an expect during the test as required by newer versions of jasmine.


rajsite:
This change will require users to run `npm install` locally after syncing to update their local dev system.

rajsite:
Added @gleonoliva and @mehagar who expressed interest in Vireo build behavior

gleonoliva:
How did you figure out which version of karma to upgrade to?

rajsite:
@gleonoliva Looking at the package-lock.json and tracing up through the file who had the dependency. Ultimately it was socket.io used by karma that had the bad dependency. Then looked at karma's history for socket.io upgrades.

Not sure if there are better approaches but it also didn't take very long.

rajsite:
Closing until I can investigate the errors more closely

PR #484, opened 2018-08-01
Cluster binary op fix

segaljared:
This is fixing an issue that was originally filed on just Quotient & Remainder, but I found on all binary math operations involving clusters where the operator or elements in the cluster need to be converted. For example, take a cluster with 3 doubles and subtract a I8 from it. The I8 must be converted to double in order to do the element-wise subtraction. This works on the first element of the cluster, but fails on subsequent elements.

It looks like the ClusterBinaryOp code tries to offset the parameter to the conversion instruction by a value in the actual operation instruction (Sub, Quotient, Remainder, etc.) which results in it accessing the wrong memory after the first element, which has no offset. This will be a little more clear in the code (I'll leave comments there)


source/core/GenericFunctions.cpp
```diff
@@ -1848,11 +1848,11 @@ VIREO_FUNCTION_SIGNATURET(ClusterBinaryOp, AggregateBinOpInstruction)
             int whichConvertArg = -int(intptr_t(convertSnippet->_p1));
             snippet = (Instruction4<AQBlock1, AQBlock1, AQBlock1, AQBlock1>*)convertSnippet->Next();
             if (whichConvertArg == 1) {
-                convertSnippet->_p0 = snippet->_p0 + (size_t)_ParamPointer(SX);
```
segaljared: snippet->_p0 corresponds to the offset in the actual instruction we're trying to perform (e.g. Sub or Remainder) not the conversion instruction. However, in [Generics.cpp#303](https://github.com/segaljared/VireoSDK/blob/3562cf61af93e76ce1b967cc45213b5ac630224b/source/core/GenericFunctions.cpp#L303), for example,  we've already put the destination offset in that parameter, so we'll end up using the wrong offset. The right offset has already been put into the conversion instruction, so we just need to add _ParamPointer(SX) like we would otherwise.

test-it/ViaTests/ClusterBinaryOpBug.via
```diff
@@ -0,0 +1,55 @@
+define(clusterBinaryOpBug dv(.VirtualInstrument (
```
sanmut: Please add a similar test for the following cases:

- Array of clusters
- Unary Op (for both the cluster case you have and the array of clusters case)

 array of clusters.

PR #485, opened 2018-08-03
Call CreateSourceFromMessage for JSLI errors

mehagar:


source/io/module_javaScriptInvoke.js
```diff
@@ -388,7 +388,8 @@
             } catch (ex) {
                 var newErrorStatus = true;
                 var newErrorCode = ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke.CODE;
-                var newErrorSource = Module.coreHelpers.formatMessageWithException(ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke.MESSAGE + '\'' + functionName + '\'.', ex);
+                var errorMessage = Module.coreHelpers.formatMessageWithException(ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke.MESSAGE + '\'' + functionName + '\'.', ex);
```
rajsite: Can we change from
` + '\'' + functionName + '\'.'`

To
` + '\nfunction: ' + functionName`

Following the pattern in the http client here: https://github.com/ni/VireoSDK/blob/master/source/io/module_httpClient.js#L632

Same for the other usages in the file

PR #486, opened 2018-08-03
Enqueue front fix

gleonoliva:
Fixing an issue with QueueCore. I found a case where queue would stop working by following the operations:
1. Obtain an unbounded queue
2. Enqueue an element
3. Dequeue element
4. EnqueueFront an element <- Vireo would  crash here.

The issue was that by enqueuing an element to an unbounded queue would increase the size of the internal array that holds the elements, and also increase `_insert` and `_count` by 1.
However by doing a Dequeue, `_count` is decreased but `_insert` is not modified. Next EnqueueFront first gets the index where to push to by calling `RemoveIndex()` which returns `_insert - _count` (1), after calling `TryMakeRoom()` the variable `_insert` has been updated to reflect the fact that the queue is actually empty and insertion index should be 0, which at this point it is. The fix is to get `front` again after `_insert` has been updated.

Adding a test that can catch the issue.


source/core/Synchronization.cpp
```diff
@@ -521,6 +521,7 @@ Boolean QueueCore::PushFront(void* pData)
     IntIndex front = RemoveIndex();
     if (!TryMakeRoom(1, front))
         return false;
+    front = RemoveIndex();  // front may have changed after making room
```
gleonoliva: Is worth noting that this step is not necessary in the case of `Enqueue` because we are always using the variable `_insert` directly instead of a computed one like `front`.

test-it/ViaTests/QueueRefNumEnqueueFrontBug.via
```diff
@@ -0,0 +1,44 @@
...
+        e(.Boolean badAliasingTestPass)
+        e(a(.Double *) remainingElts)
+        e(.ErrorCluster err)
+    )
+
+    clump (
+        ObtainQueue(q * * * * err)
+        Enqueue(q 0.0 1000 timedOut err)
+        Printf("Enqueue with empty unbounded queue\n")
+        GetQueueStatus(q true maxsize * * * ne remainingElts err)
```
sanmut: I think we should print err (at least, err.status and err.code).

sanmut: Even with queue of fixed size, following case fail where last Deque return the incorrect value.
1) Enqueue, Deque, EnqueueFront, Dequeue
2) EnqueueFront, Deque, EnqueueFront, Dequeue
Please add more tests including the above. I would also use "PeekQueue" before Dequeue.  Adding More tests in VI test suite are acceptable. However, we should make sure those all pass for the proposed fix.

sanmut:
I would also update the commit message if you also find out that the problem exists regardless of whether the queue is bounded or not.

PR #487, opened 2018-08-03
Multi register events

segaljared:
Fixing an issue where we would call into jsRegisterForControlEvent multiple times for the same event on the same control. In order to fix this, I've modified RegisterForEvents so that it returns an enum indicating whether this was a registration for a new event on a control, a new event queue looking at an already registered event, or if somehow no change occurred. Only in the case of a new event being registered on a control should we call into jsRegisterForControlEvent.

### DIFF
--- a/source/core/Events.cpp
+++ b/source/core/Events.cpp
@@ -178,6 +178,8 @@ class EventOracle {
     static EventOracle _s_singleton;
 
  public:
+    typedef enum { kNewRegistration, kAddedToNewQueue, kNoChange } EventInsertStatus;
+
     size_t size(EventQueueID qID) const { return _qObject[qID].size(); }
     void OccurEvent(EventOracleIndex eventOracleIdx, const EventData &eData);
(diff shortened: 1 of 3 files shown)


source/core/Events.cpp
```diff
@@ -246,8 +248,9 @@ bool EventOracle::GetControlInfoForEventOracleIndex(EventOracleIndex eventOracle
 }
 
 // EventListInsert -- add registration entry at given eventOracleIndex for event source/type/ref, watching the given QueueID
-bool EventOracle::EventListInsert(EventOracleIndex eventOracleIndex, EventRegQueueID eventRegQueueID, EventSource eSource, EventType eType) {
-    bool added = false;
+EventOracle::EventInsertStatus EventOracle::EventListInsert(EventOracleIndex eventOracleIndex, EventRegQueueID eventRegQueueID, EventSource eSource,
+                                   EventType eType) {
+    EventInsertStatus added = kNoChange;
```
cglzaguilar: Rename added to status. It doesn't convey a boolean meaning anymore.

PR #488, opened 2018-08-07
Update error strings

mehagar:


source/io/module_javaScriptInvoke.js
```diff
@@ -34,32 +34,32 @@
 
         kNIUnableToInvokeAJavaScriptFunction: {
             CODE: 44300,
-            MESSAGE: 'An exception occurred within the external JavaScript function called by a JavaScript Library Interface Node. Verify the values you wired to the node. Exception thrown when calling: '
+            MESSAGE: 'An exception occurred within the external JavaScript function called by a JavaScript Library Interface node. Verify your JavaScript code is valid. Exception thrown when calling: '
```
rajsite: So we should be removing the "Exception throw when calling: " correct?
As we talked about the error messages should be standalone. We now use the additional information facility to take on more info.

PR #490, opened 2018-08-16
Adding Internal functions to JSInvoke

ggumen:
Adding the concept of internal functions to JavaScriptInvoke calls, by re-using an unused parameter errorCheckingEnabled.

### DIFF
--- a/source/io/JavaScriptInvoke.cpp
+++ b/source/io/JavaScriptInvoke.cpp
@@ -24,8 +24,8 @@ namespace Vireo {
 #if kVireoOS_emscripten
 extern "C" {
     // JavaScript function prototypes
-    // Parameters: occurrence, functionName, returnValue, parameters*, parametersCount, isInternalFunction, errorStatus*, errorCode*, errorSource*
-    extern void jsJavaScriptInvoke(OccurrenceRef, StringRef, void *, void *, Int32, Boolean, Boolean *, Int32 *, StringRef);
+    extern void jsJavaScriptInvoke(OccurrenceRef occurrence, StringRef functionName, void *returnValue, void *parameters, Int32 parametersCount, 
+                                    Boolean isInternalFunction, Boolean *errorStatus, Int32 *errorCode, StringRef errorSource);
 }
 #endif
 
--- a/source/io/JavaScriptInvoke.cpp
+++ b/source/io/JavaScriptInvoke.cpp
@@ -24,8 +24,7 @@ namespace Vireo {
 #if kVireoOS_emscripten
 extern "C" {
     // JavaScript function prototypes
-    // Parameters: occurrence, functionName, returnValue, parameters*, parametersCount, errorCheckingEnabled, errorStatus*, errorCode*, errorSource*
-    // Remark : errorCheckingEnabled is currently unused. It was added to support an optional error checking feature just like the SLI does.
+    // Parameters: occurrence, functionName, returnValue, parameters*, parametersCount, isInternalFunction, errorStatus*, errorCode*, errorSource*
     extern void jsJavaScriptInvoke(OccurrenceRef, StringRef, void *, void *, Int32, Boolean, Boolean *, Int32 *, StringRef);
 }
 #endif
@@ -105,11 +104,11 @@ VIREO_EXPORT void* JavaScriptInvoke_GetParameterPointer(StaticTypeAndData *param
 }
 
 //------------------------------------------------------------
-// arguments: occurrence, errorCheckingEnabled, errorCluster, functionName, returnValue, then variable number of inputs that can be nullptr or any type
+// arguments: occurrence, isInternalFunction, errorCluster, functionName, returnValue, then variable number of inputs that can be nullptr or any type
 struct JavaScriptInvokeParamBlock : public VarArgInstruction
 {
     _ParamDef(OccurrenceRef, occurrence);
-    _ParamDef(Boolean, errorCheckingEnabled);
+    _ParamDef(Boolean, isInternalFunction);
     _ParamDef(ErrorCluster, errorCluster);
     _ParamDef(StringRef, functionName);
     _ParamImmediateDef(StaticTypeAndData, returnValue[1]);
@@ -131,8 +130,8 @@ VIREO_FUNCTION_SIGNATUREV(JavaScriptInvoke, JavaScriptInvokeParamBlock)
     Observer* pObserver = clump->GetObservationStates(2);
     if (!pObserver) {
         StringRef functionName = _Param(functionName);
-        Boolean errorCheckingEnabled = _Param(errorCheckingEnabled);
-        const Int32 configurationParameters = 4;  // occurrence, errorCheckingEnabled, errorCluster and functionName
+        Boolean isInternalFunction = _Param(isInternalFunction);
+        const Int32 configurationParameters = 4;  // occurrence, isInternalFunction, errorCluster and functionName
         const Int32 staticTypeAndDataParameters = 2;  // Two parameters are inserted, one for type another for data. See StaticTypeAndData definition.
         Int32 userParametersCount = (_ParamVarArgCount() - configurationParameters - staticTypeAndDataParameters) / staticTypeAndDataParameters;
         StaticTypeAndData *returnValuePtr = _ParamImmediate(returnValue);
@@ -147,7 +146,7 @@ VIREO_FUNCTION_SIGNATUREV(JavaScriptInvoke, JavaScriptInvokeParamBlock)
                 returnValuePtr,
                 parametersPtr,
                 userParametersCount,
-                errorCheckingEnabled,
+                isInternalFunction,
                 &errorClusterPtr->status,
                 &errorClusterPtr->code,
                 errorClusterPtr->source);
@@ -170,7 +169,7 @@ VIREO_FUNCTION_SIGNATUREV(JavaScriptInvoke, JavaScriptInvokeParamBlock)
 //------------------------------------------------------------
 DEFINE_VIREO_BEGIN(JavaScriptInvoke)
     DEFINE_VIREO_REQUIRE(Synchronization)
-    DEFINE_VIREO_FUNCTION(JavaScriptInvoke, "p(i(VarArgCount argumentCount) i(Occurrence occurrence) i(Boolean errorCheckingEnabled)"
+    DEFINE_VIREO_FUNCTION(JavaScriptInvoke, "p(i(VarArgCount argumentCount) i(Occurrence occurrence) i(Boolean isInternalFunction)"
         "io(ErrorCluster errorCluster) i(String functionName) o(StaticTypeAndData returnValue) io(StaticTypeAndData functionParameters))")
 DEFINE_VIREO_END()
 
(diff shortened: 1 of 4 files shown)


source/io/JavaScriptInvoke.cpp
```diff
@@ -24,8 +24,7 @@ namespace Vireo {
 #if kVireoOS_emscripten
 extern "C" {
     // JavaScript function prototypes
-    // Parameters: occurrence, functionName, returnValue, parameters*, parametersCount, errorCheckingEnabled, errorStatus*, errorCode*, errorSource*
-    // Remark : errorCheckingEnabled is currently unused. It was added to support an optional error checking feature just like the SLI does.
+    // Parameters: occurrence, functionName, returnValue, parameters*, parametersCount, isInternalFunction, errorStatus*, errorCode*, errorSource*
```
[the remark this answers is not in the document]
ggumen: I agree that would be more readable.
Looks like most of our extern declarations are written without parameter names. Is there a reason why we do this? 
If this is something we want to move away from, then I can update this case.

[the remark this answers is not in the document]
sanmut: No particular reason why/how someone wrote without params to begin with.

source/io/module_javaScriptInvoke.js
```diff
@@ -432,18 +442,33 @@
                     throw new Error('The API being accessed for ' + functionName + ' is not valid anymore.');
                 }
                 completionCallbackStatus.retrievalState = completionCallbackRetrievalEnum.RETRIEVED;
-                return generateCompletionCallback(occurrencePointer, functionName, returnTypeName, returnValuePointer, errorStatusPointer, errorCodePointer, errorSourcePointer, completionCallbackStatus);
+                return generateCompletionCallback(occurrencePointer, functionName, returnTypeName, returnValuePointer, errorStatusPointer, errorCodePointer, errorSourcePointer, completionCallbackStatus, isInternalFunction);
             };
+
+            if (isInternalFunction) {
+                api.setLabVIEWError = function (statusCode, errorCode, errorMessage) {
```
rajsite: Should keep the parameter names close to the lv error representation: `status`, `code`, `source`.

rajsite: longer term now that we have good apis for writing to clusters we will be able to reduce the number of references we pass around to just the error cluster instead of all the componenets

source/io/module_javaScriptInvoke.js
```diff
@@ -432,18 +442,33 @@
...
+            }
             return api;
         };
 
+        publicAPI.javaScriptInvoke.registerInternalFunctions = function (functionsToAdd) {
+            Object.keys(functionsToAdd).forEach(function (name) {
+                if (internalFunctionsMap.has(name)) {
+                    throw new Error('Private function already registered for name:' + name);
+                }
+                internalFunctionsMap.set(name, functionsToAdd[name]);
```
rajsite: should also check that typeof functionsToAdd[name] === 'function'

source/io/JavaScriptInvoke.cpp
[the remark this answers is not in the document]
ggumen: Updated the parameters.

source/io/module_javaScriptInvoke.js
ggumen: done.

source/io/module_javaScriptInvoke.js
ggumen: done.

source/io/module_javaScriptInvoke.js
```diff
@@ -411,6 +418,9 @@
                 if (!(returnValue instanceof Error)) {
                     tryUpdateReturnValue(functionName, returnTypeName, returnValuePointer, returnValue, errorStatusPointer, errorCodePointer, errorSourcePointer, completionCallbackStatus);
```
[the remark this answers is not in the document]
ggumen: done.

source/io/module_javaScriptInvoke.js
```diff
@@ -432,18 +442,33 @@
...
+                    Module.coreHelpers.mergeErrors(statusCode, errorCode, errorMessage, errorStatusPointer, errorCodePointer, errorSourcePointer);
+                };
+            }
             return api;
         };
 
+        publicAPI.javaScriptInvoke.registerInternalFunctions = function (functionsToAdd) {
+            Object.keys(functionsToAdd).forEach(function (name) {
+                if (internalFunctionsMap.has(name)) {
+                    throw new Error('Private function already registered for name:' + name);
```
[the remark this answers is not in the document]
ggumen: done.

source/io/module_javaScriptInvoke.js
```diff
@@ -432,18 +443,36 @@
...
+            }
             return api;
         };
 
+        publicAPI.javaScriptInvoke.registerInternalFunctions = function (functionsToAdd) {
+            Object.keys(functionsToAdd).forEach(function (name) {
+                if (internalFunctionsMap.has(name)) {
+                    throw new Error('Internal function already registered for name:' + name);
+                }
+                if (functionsToAdd[name] === 'function') {
```
cglzaguilar: Revisit this code, we want to throw if it is not a function.

source/io/module_javaScriptInvoke.js
```diff
@@ -400,17 +407,21 @@
...
                 }
                 if (completionCallbackStatus.invocationState === completionCallbackInvocationEnum.REJECTED) {
                     throw new Error('The call to ' + functionName + ' threw an error, so this callback cannot be invoked.');
                 }
-                if (!(returnValue instanceof Error)) {
+                var errorStatus = Module.eggShell.dataReadBoolean(errorStatusPointer);
+                if (!(returnValue instanceof Error) && !errorStatus) {
                     tryUpdateReturnValue(functionName, returnTypeName, returnValuePointer, returnValue, errorStatusPointer, errorCodePointer, errorSourcePointer, completionCallbackStatus);
                 } else {
+                    if (isInternalFunction) {
```
cglzaguilar: I don't think we want to throw in this case if errorStatus is true.

cglzaguilar: But we also don't want to create a new errorMessage if errorStatus is true.

rajsite: For an internal function with error status true we should re-throw the exception.

internal functions should not throw exceptions, if they do they need to propagate up so they fail tests, etc.

This case is subtle, it may be worth adding a comment describing the case here

source/io/module_javaScriptInvoke.js
```diff
@@ -400,29 +407,35 @@
...
                 }
-                if (!(returnValue instanceof Error)) {
-                    tryUpdateReturnValue(functionName, returnTypeName, returnValuePointer, returnValue, errorStatusPointer, errorCodePointer, errorSourcePointer, completionCallbackStatus);
-                } else {
-                    var newErrorStatus = true;
-                    var newErrorCode = ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke.CODE;
-                    var errorMessage = Module.coreHelpers.formatMessageWithException(ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke.MESSAGE + '\nfunction: ' + functionName, returnValue);
-                    var newErrorSource = Module.coreHelpers.createSourceFromMessage(errorMessage);
-                    Module.coreHelpers.mergeErrors(newErrorStatus, newErrorCode, newErrorSource, errorStatusPointer, errorCodePointer, errorSourcePointer);
+                var errorStatus = Module.eggShell.dataReadBoolean(errorStatusPointer);
```
rajsite: Why was this check added? What case is it trying to guard against?

source/io/module_javaScriptInvoke.js
```diff
@@ -516,7 +550,10 @@
 
             // assume synchronous invocation since the completion callback was never retrieved
             if (completionCallbackStatus.retrievalState === completionCallbackRetrievalEnum.AVAILABLE) {
-                tryUpdateReturnValue(functionName, returnTypeName, returnValuePointer, returnValue, errorStatusPointer, errorCodePointer, errorSourcePointer, completionCallbackStatus);
+                var errorStatus = Module.eggShell.dataReadBoolean(errorStatusPointer);
+                if (!errorStatus) {
```
rajsite: why was this case added?

So one thing we tried to do is represent the possible states of the jsli call using completionCallbackRetrievalEnum and completionCallbackInvocationEnum because there are a lot of possible configurations of the state of the JSLI function call.

What is the case where we are trying to look at Vireo's state of the error cluster? WE should try and capture it in the state enums or have the condition resolved sooner. IE if the input to the JSLI call is an error we should handle that much earlier.

source/io/module_javaScriptInvoke.js
```diff
@@ -432,18 +445,36 @@
...
+                api.setLabVIEWError = function (status, code, source) {
+                    Module.coreHelpers.mergeErrors(status, code, source, errorStatusPointer, errorCodePointer, errorSourcePointer);
+                };
+            }
             return api;
         };
 
+        publicAPI.javaScriptInvoke.registerInternalFunctions = function (functionsToAdd) {
+            Object.keys(functionsToAdd).forEach(function (name) {
+                if (internalFunctionsMap.has(name)) {
```
rajsite: Can we add one more assert please? :)

typeof name === 'string' && name.length > 0

test-it/karma/javascriptinvoke/InternalFunction.Test.js
```diff
@@ -0,0 +1,77 @@
...
+            }
+        });
+
+        runSlicesAsync(function (rawPrint, rawPrintError) {
+            expect(rawPrint).toBeEmptyString();
+            expect(rawPrintError).toBeEmptyString();
+            expect(viPathParser('error2.status')).toBeTrue();
+            expect(viPathParser('error2.code')).toBe(777);
+            expect(viPathParser('error2.source')).toContain('this is the error message');
+            expect(viPathParser('returnValue2')).toBe(0);
```
rajsite: should be 1 :)

source/io/module_javaScriptInvoke.js
```diff
@@ -496,13 +524,16 @@
...
             if (functionToCall.length === parameters.length + 1) {
-                api = generateAPI(occurrencePointer, functionName, returnTypeName, returnValuePointer, errorStatusPointer, errorCodePointer, errorSourcePointer, completionCallbackStatus);
+                api = generateAPI(occurrencePointer, functionName, returnTypeName, returnValuePointer, errorStatusPointer, errorCodePointer, errorSourcePointer, completionCallbackStatus, isInternalFunction);
                 parameters.push(api);
             }
 
             try {
                 returnValue = functionToCall.apply(context, parameters);
             } catch (ex) {
+                if (isInternalFunction) {
```
mehagar: Why does this throw again if it is an internal function?

rajsite: In user JS code we don't want to crash the runtime. It should be caught and reported as a JS error back on the wire.

Our internal functions should allow the exception through because it means there is a failure in code we have written. If internal functions need to report an error on the wire they should use the jsapi for setting LabVIEW style errors, not throwing an exception. As such exceptions are actually exceptions and need to propagate upwards to fail tests or cause the editor to receive an error, etc.

PR #492, opened 2018-08-28
add notifications for ms teams webhook

mehagar:


rajsite:
Will this webhook post in it's own channel in teams?

Can we filter the notification messages? Seems like passed builds for tagged releases and failed builds are the most interesting. Pull request builds in particular would be noisy.

PR #493, opened 2018-08-28
Create nuget package for vireo executable for tests.

mehagar:
The .nuspec file is what defines the nuget package. As part of the appveyor build I get the `npm version` of vireo and use that to set the nuget version. The "provider" section in the appveyor.yml then knows how to publish the nuget package to Nuget.org.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -47,4 +47,4 @@ deploy:
   provider: NuGet
   artifact: '/.*\.nupkg/'
   api_key:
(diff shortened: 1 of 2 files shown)


mehagar:
See https://www.nuget.org/packages/vireo/11.0.0.134 for an example of what it generates. Note that we are generating a vireo nuget package on every successful build now. I wasn't sure how to only create a nuget on a github release tag like we do for the zipped executable.

mehagar:
I am about to run a lvjs now with the vireo nuget I created.

PR #498, opened 2018-09-01
Update dev tooling

rajsite:
Karma had an indirect dependecy on a package with a known vulnerability described here: https://nvd.nist.gov/vuln/detail/CVE-2017-16113

This update migrates all packages to the latest versions and in the process removes the vulnerable dependency.

A change was required to the karma vtr test runner as the test required specific ordering of describe blocks and newer versions of karma seem to be able to run describe blocks in different orders. This is a reasonable change as tests should not have dependencies between describe blocks.

After merging into mainline **devs will need to run npm install** to get the latest packages and **should rebase existing active branches with master**.


rajsite:
Added memo and michael who are interested in build infra

test-it/karma/utilities/MainAmd.js
```diff
@@ -4,7 +4,7 @@
     var tests = [];
     for (var file in window.__karma__.files) {
         if (window.__karma__.files.hasOwnProperty(file)) {
-            if (/test\.js$/.test(file)) {
+            if ((/test\.js$/).test(file)) {
```
mehagar: What is the reason behind this change? Is this protecting against a vulnerability?

rajsite: Nah it was an eslint rule that changed in eslint 5 to make regular expressions unambiguous by wrapping them in parens

PR #499, opened 2018-09-04
Fix scan from string for booleans

mehagar:
The values I'm using to check for true or false come from http://zone.ni.com/reference/en-XX/help/371361N-01/glang/scan_from_string/


test-it/ExpectedResults/StringScanBoolean.vtr
```diff
@@ -0,0 +1,20 @@
...
+offset_past:<5>, remaining:()
+:<false> err:<(false 0 '')>
+offset_past:<5>, remaining:()
+:<false> err:<(false 0 '')>
+offset_past:<2>, remaining:()
+:<false> err:<(false 0 '')>
+offset_past:<3>, remaining:()
+:<false> err:<(false 0 '')>
+offset_past:<5>, remaining:()
+:<false> err:<(false 0 '')>
```
spathiwa: Add a test for correct behavior when an error occurs (the expected Boolean text isn't there, or end of string).

source/core/NumericString.cpp
```diff
@@ -1291,6 +1293,39 @@ IntMax ScanIntValues(char formatChar, char* beginPointer, char** endPointer)
...
 
+std::string validTrueValues[5] = { "y", "t", "true", "on", "yes" };
+std::string validFalseValues[4] = { "f", "false", "off", "no" };
+
+Boolean ScanBooleanValue(char* beginPointer, char** endPointer)
+{
+    std::string toLowered = std::string(beginPointer);
+    *endPointer = beginPointer;
+
+    std::transform(toLowered.begin(), toLowered.end(), toLowered.begin(), ::tolower);
```
spathiwa: I actually don't think it's a good idea to be constructing std::strings in a run-time operation for performance reasons.  None of the other string parsing functions do this.

mehagar:
There's another case I didn't implement - that of using a float like 0.5 and passing in a format string like %f, which would return a false boolean. Is it OK if I don't implement that behavior in this change our would you like me to also implement that?

sanmut:
Your commits are confusing. I think you should rebase your local branch changes first before raising a PR. (git remote update -p   ; git rebase -p origin/master). This will create a PR with clean history of relevant commits for the change.

source/core/NumericString.cpp
sanmut: +1. Why not use one of the SubString::Compare() variants that can be called with ignoring case.

source/core/NumericString.cpp
```diff
@@ -1291,6 +1293,39 @@ IntMax ScanIntValues(char formatChar, char* beginPointer, char** endPointer)
     return intValue;
 }
 
+std::string validTrueValues[5] = { "y", "t", "true", "on", "yes" };
+std::string validFalseValues[4] = { "f", "false", "off", "no" };
```
spathiwa: CG/NCG also matches "n".  The documentation page you linked incorrectly omits this.

spathiwa:
Also okay with not addressing float here, but you should also, in the bug report, mention that it needs to handle the other format codes as well (%d,%e,%g,%x,%o).

source/core/NumericString.cpp
sanmut: ah, good find. Please file a documentation CAR

mehagar:
I ran a lvjs and it seemed to pass correctly. I say seem because there were some existing failures.

source/core/NumericString.cpp
```diff
@@ -1291,6 +1293,38 @@ IntMax ScanIntValues(char formatChar, char* beginPointer, char** endPointer)
     return intValue;
 }
 
+char validTrueValues[5][10] = { "y", "t", "true", "on", "yes" };
+char validFalseValues[5][10] = { "n", "f", "false", "off", "no" };
```
sanmut: Since the size of array of strings and length of each string is unknown, we should initialize like the following:
static const char *validTrueValues[] = { "y", "t", "true", "on", "yes" };
static const char *validFalseValues[] = { "n", "f", "false", "off", "no" };
Making it "static" and "const" implies the intentions also.

source/core/NumericString.cpp
```diff
@@ -1490,15 +1527,15 @@ Boolean TypedScanString(SubString* inputString, IntIndex* endToken, const Format
                             // reach the end of the input
                             i--;
                         }
-                        if (formatOptions->MinimumFieldWidth > 0 && i+1-stringStart > formatOptions->MinimumFieldWidth) {
+                        if (formatOptions->MinimumFieldWidth > 0 && i + 1 - stringStart > formatOptions->MinimumFieldWidth) {
```
cglzaguilar: Please create another PR with the refactoring changes. We are going to cherry-pick this PR and create a Hot-fix for it. I want to cherry-pick the code that is strictly necessary.

source/core/NumericString.cpp
```diff
@@ -1291,6 +1293,38 @@ IntMax ScanIntValues(char formatChar, char* beginPointer, char** endPointer)
...
+    }
+
+    for (const char* falseValue : validFalseValues) {
+        if (s.CompareCStrIgnoreCase(falseValue)) {
+            *endPointer += strlen(falseValue);
+            return false;
+        }
+    }
+
+    return false;
```
cglzaguilar: Is this how CG and NG behave, returning false when none of the "true", "false" values are found? Do we have a test for this?

mehagar: I tested it and it is based on the default value wired in, so what I have here is incorrect. I should be returning the default value.

source/core/NumericString.cpp
```diff
@@ -1291,6 +1291,44 @@ IntMax ScanIntValues(char formatChar, char* beginPointer, char** endPointer)
     return intValue;
 }
 
+static const char *validTrueValues[] = { "y", "t", "true", "on", "yes" };
+static const char *validFalseValues[] = { "n", "f", "false", "off", "no" };
+
+Boolean TryScanBooleanValue(char* beginPointer, char** endPointer, Boolean* value)
```
spathiwa: The return of a Boolean success flag seems an insufficient reason to me for putting the word "Try" in the function name.  We don't have this kind of convention anywhere else.

cglzaguilar:
@mehagar, also fix the cpp linting issues that appear in your PR.

PR #504, opened 2018-09-07
Fix scan from string for booleans (#499)

cglzaguilar:
* fix scan from string for booleans

* add test for errors and if so return default value

* revert refactorings

* remove one more refactoring

* remove one more refactoring 2

* remove try part

* return false from function


cglzaguilar:
This is a straight merge for ec64f3296e5ee858dbe088768f0ff4fb4afad221

PR #506, opened 2018-09-12
Add docs for using local vireo esh.exe

mehagar:
So contributors can test their vireo changes on the CI, I added documentation for how to do so.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,12 +6,14 @@ It covers setting up your own fork with continuous integration (CI) server conne
 # Getting started
 
 ## Create a fork
+
 Vireo relies on the [fork-pull model](https://help.github.com/articles/about-collaborative-development-models/) for developer contributions.
 To begin development on Vireo make a GitHub account and [fork the ni/VireoSDK repository](https://help.github.com/fork-a-repo/).
 
 Your fork will be available online at github.com/YOUR_USER_NAME/VireoSDK
 
 ## Connect fork to CI servers
+
 Vireo uses Travis and AppVeyor to build and test the Vireo source code automatically when changes are made.
 You are encouraged to push changes to your fork when you have changes and to make branches for experiments.
 The fork is your playground for development and experimentation.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -196,6 +196,15 @@ This allows you to make changes to your local vireo and have them be instantly a
 
 **WARNING** Since the local VireoSDK is symlinked to the JavaScript application be aware of tooling that could accidently delete or modify your local development files unexpectedly.
 
+If you have a .NET application that depends on Vireo you can do the following (you must have nuget.exe available)
+- In the directory with the `VireoSDK.nuspec` file run:
+  ```console
+  nuget pack VireoSDK.nuspec -properties version="[version]"
+  ```
+  
+  replacing \[version\] with some version.
```
rajsite: Instead of slash escaping you can use the back tick \`\[version\]\` so it renders as an identifer `[version]`

CONTRIBUTING.md
```diff
@@ -196,6 +196,15 @@ This allows you to make changes to your local vireo and have them be instantly a
...
+- See [installing a nuget package](https://docs.microsoft.com/en-us/nuget/consume-packages/ways-to-install-a-package) for the various ways to install the created package to your .NET application.
```
rajsite: Can you mention the expected output of the command, ie a vireo.nupkg file (?) is generated in the folder the command is run (?)

CONTRIBUTING.md
```diff
@@ -196,6 +196,15 @@ This allows you to make changes to your local vireo and have them be instantly a
...
+
 ### Testing on a different machine / CI
 If you want to test your local vireo changes on a separate machine from your local machine (a CI or test machine) without publishing to the npm registry you can do the following:
```
rajsite: So if you create a vireo.nupkg file locally, how do you create a build that points to it on the CI? Please add instructions to this section as well

rajsite:
@mehagar please provide descriptions in the pull requests

CONTRIBUTING.md
```diff
@@ -176,64 +189,97 @@ It is encouraged before submission to rebase your changes on top of master and t
...
+    ```
+    replacing `[version]` with some version.
+
+    This command will generate a `.nupkg` file in the folder the command is run from.
+3. See [installing a nuget package](https://docs.microsoft.com/en-us/nuget/consume-packages/ways-to-install-a-package) for the various ways to install the created package to your .NET application.
+
+### Testing esh.exe on a different machine / CI
+
+1. Follow the same steps above for testing on your local machine.
+2. Once you have installed the nuget package, you must add a package source for that package so the remote machine can download the package. See [publishing a package](https://docs.microsoft.com/en-us/nuget/create-packages/publish-a-package) for information how to do this.
```
rajsite: Is it not possible to reference a nupkg file that you add to the shelveset for testing? Requiring the package be published for testing is a bit extreme.

mehagar: I see what you mean. At least with visual studio, the sources for a nuget are defined in the NuGet.config file (https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file). Currently we have nuget.org and the LabVIEW TFS packages referenced here, but that would require we publish to one of those feeds. I thought I could add a reference to the "Tests.HtmlVI" project, but then I realized that the path is absolute in the examples and I don't know what the path would be on the CI machine.

We can talk about this in person if you would like. My plan for our squad is to just install the package locally on your machine, and then add the esh.exe directly to the shelveset instead of trying to reference some local package. Or, perhaps we could write a script that publishes a test package to our internal ni feed so as to not pollute the public nuget.org vireo feed.

As for this document though, I wanted to be as general as possible so others would know how to actually publish a test package if that's what they wanted to do.

mehagar:
Whoops, didn't mean to close this. Ugh, I accidentally deleted the remote branch. Restored and reopened.

rajsite:
Closing until revisited

CONTRIBUTING.md
mehagar: I found another way to do this without having to publishing to a feed. Also, there is a way to reference relative paths in the NuGet.config file.

PR #509, opened 2018-09-14
add webhook notification

mehagar:
This will replace the appveyor website settings so we have the notification settings in source control.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -68,3 +68,38 @@ deploy:
       secure: O5sZ6nRPlF2WbhTUSvrSTou7s6Pdg7G4PClfzvlGTU2XGwErDYRNyayAsUzjFsTl
     on:	
       appveyor_repo_tag: true
+notifications:
+  - provider: Webhook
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -68,3 +68,38 @@ deploy:
       secure: O5sZ6nRPlF2WbhTUSvrSTou7s6Pdg7G4PClfzvlGTU2XGwErDYRNyayAsUzjFsTl
     on:	
       appveyor_repo_tag: true
+notifications:
+  - provider: Webhook
+    url: https://outlook.office.com/webhook/6ee4efaa-986e-4b8e-8d6b-cfa47c89f4b0@87ba1f9a-44cd-43a6-b008-6fdb45a5204e/IncomingWebhook/48bf1f46af764a2ea8cb38acd029bb5a/0ad415e3-d66a-425c-9914-e9e52f2b81e1
```
rajsite: We should coordinate to make sure and encrypt this url and token for both AppVeyor and Travis

Since both tokens have been exposed publicly they should both be invalidated and we should create new ones that we make sure to encrypt

mehagar: Good idea. Is this what I should use for encrypting both urls and tokens?
https://ci.appveyor.com/tools/encrypt

rajsite:
Closing until revisited

PR #510, opened 2018-09-20
add callchain for propertynode if internal

mehagar:
The functionName will be something like PropertyNode_PropertyRead or PropertyNode_PropertyWrite, which we want to show the user instead of "JavaScriptInvoke" if they were indeed using a property node.

### DIFF
--- a/source/io/JavaScriptInvoke.cpp
+++ b/source/io/JavaScriptInvoke.cpp
@@ -150,7 +150,13 @@ VIREO_FUNCTION_SIGNATUREV(JavaScriptInvoke, JavaScriptInvokeParamBlock)
                 &errorClusterPtr->status,
                 &errorClusterPtr->code,
                 errorClusterPtr->source);
-            AddCallChainToSourceIfErrorPresent(errorClusterPtr, "JavaScriptInvoke");
+
+            if (isInternalFunction) {
+                AddCallChainToSourceIfErrorPresent(errorClusterPtr, (const char*)functionName->Begin());
+            } else {
+                AddCallChainToSourceIfErrorPresent(errorClusterPtr, "JavaScriptInvoke");
+            }
+
             InstructionCore* instructionCorePtr = clump->WaitOnObservableObject(_this);
             return instructionCorePtr;
         }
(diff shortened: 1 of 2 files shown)


cglzaguilar:
Please add a VIA test that exercises this code path. Since internal functions for property nodes are really defined in another system, in this case, the panel code, we can't test them here, but what we can test is calling an "external function" that we define for the test that returns an error so we can exercise the new code that was added.

mehagar:
@cglzaguilar , I believe @ggumen already added such a test as part of #490. See the file test-it/karma/javascriptinvoke/InternalFunction.Test.js.

ggumen:
> 
> 
> @cglzaguilar , I believe @ggumen already added such a test as part of #490. See the file test-it/karma/javascriptinvoke/InternalFunction.Test.js.

Yes, it does exist. Can you update that test to also check for the new function name you're appending to the error message?

test-it/karma/javascriptinvoke/InternalFunction.Test.js
```diff
@@ -62,6 +62,7 @@ describe('A JavaScript function invoke', function () {
             expect(viPathParser('error2.status')).toBeTrue();
             expect(viPathParser('error2.code')).toBe(777);
             expect(viPathParser('error2.source')).toContain('this is the error message');
+            expect(viPathParser('error2.source')).toContain('in MyVI');
```
cglzaguilar: I think we need to check for "NI_InternalFunctionSetsError" to be contained on the error string. That is the purpose of the change you're making.

source/io/JavaScriptInvoke.cpp
```diff
@@ -150,7 +150,13 @@ VIREO_FUNCTION_SIGNATUREV(JavaScriptInvoke, JavaScriptInvokeParamBlock)
                 &errorClusterPtr->status,
                 &errorClusterPtr->code,
                 errorClusterPtr->source);
-            AddCallChainToSourceIfErrorPresent(errorClusterPtr, "JavaScriptInvoke");
+
+            if (isInternalFunction) {
+                AddCallChainToSourceIfErrorPresent(errorClusterPtr, (const char*)functionName->Begin());
```
sanmut: It would be nice to get the name of the property in the message.

cglzaguilar: Good point. But if I'm not mistaken that will be part of the error message that is generated in the actual code dealing with the property node call.

rajsite: That would be the responsibility of the ASW JS implementing the property node behavior, not the JavaScriptInvoke node itself.

PR #511, opened 2018-09-21
JavaScriptInvoke Optional ErrorCluster

gleonoliva:
For fixing this we could've gone with the approach taken for [PropertyNode.cpp](https://github.com/ni/VireoSDK/pull/431/files#diff-deae6f923a6d21720b2ce0a2c601a084)

However, since this is the second time we have found this kind of issue and we now have an API that allows us to write an ErrorCluster directly. I took the responsibility of refactoring `mergeErrors` to take an error cluster ValueRef and a JS object with the same properties in an error cluster (status, code, source).

After refactoring I managed to get all tests passing again. I had to add `writeStringFromArray` to handle responses for HttpClient.


source/core/module_coreHelpers.js
```diff
@@ -303,37 +303,40 @@
...
-            Module.eggShell.dataWriteInt32(existingErrorCodePointer, newErrorCode);
-            Module.eggShell.dataWriteString(exisitingErrorSourcePointer, newErrorSource);
+        var writeNewError = function (errorValueRef, newError) {
+            Module.eggShell.writeJSON(errorValueRef, JSON.stringify(newError));
         };
 
-        Module.coreHelpers.mergeErrors = function (newErrorStatus, newErrorCode, newErrorSource, existingErrorStatusPointer, existingErrorCodePointer, exisitingErrorSourcePointer) {
+        Module.coreHelpers.mergeErrors = function (errorValueRef, newError) {
             // Follows behavior of merge errors function: https://zone.ni.com/reference/en-XX/help/371361N-01/glang/merge_errors_function/
+            if (errorValueRef === undefined) {
```
gleonoliva: No wired error, do nothing. This is true because `createValueRef` now returns `undefined` if the passed `typeRef` or `dataRef` are invalid.

source/core/module_coreHelpers.js
```diff
@@ -303,37 +303,40 @@
...
+                return;
+            }
 
-            var existingErrorStatus = Module.eggShell.dataReadBoolean(existingErrorStatusPointer);
+            var errorStatusValueRef = Module.eggShell.findSubValueRef(errorValueRef, 'status');
+            var existingErrorStatus = Module.eggShell.readDouble(errorStatusValueRef) !== 0;
```
sanmut: I think the variable existingErrorStatus is redundant. isExistingError  can be directly used.

source/core/module_coreHelpers.js
```diff
@@ -303,37 +303,40 @@
...
             var isExistingError = existingErrorStatus;
```
sanmut: You could declare variables as necessary always in this method. You seem to be doing in some places but not others. Move "var isNewError = newError.status;" below "if (isExistingError)"

source/core/module_coreHelpers.js
gleonoliva: hahaha you are right. I'll remove it.

test-it/karma/fixtures/javascriptinvoke/InternalFunctions.via
```diff
@@ -12,14 +12,22 @@ define(MyVI dv(VirtualInstrument (
...
+        e(dv(.Int32 0) returnValue4)
         e(.Occurrence occurrence1)
         e(.Occurrence occurrence2)
+        e(.Occurrence occurrence3)
+        e(.Occurrence occurrence4)
     )
     clump (
         JavaScriptInvoke(occurrence1 true error1 'NI_InternalFunction' returnValue1 value1)
         JavaScriptInvoke(occurrence2 true error2 'NI_InternalFunctionSetsError' returnValue2 value2)
+        JavaScriptInvoke(occurrence3 true * 'NI_InternalFunctionNoCluster' returnValue3 value3)
```
sanmut: Rename to "NI_InternalFunctionNoErrorCluster"

source/core/module_coreHelpers.js
```diff
@@ -303,37 +303,40 @@
             return currentFPS;
         };
 
-        var writeNewError = function (newErrorStatus, newErrorCode, newErrorSource, existingErrorStatusPointer, existingErrorCodePointer, exisitingErrorSourcePointer) {
-            Module.eggShell.dataWriteBoolean(existingErrorStatusPointer, newErrorStatus);
-            Module.eggShell.dataWriteInt32(existingErrorCodePointer, newErrorCode);
-            Module.eggShell.dataWriteString(exisitingErrorSourcePointer, newErrorSource);
+        var writeNewError = function (errorValueRef, newError) {
+            Module.eggShell.writeJSON(errorValueRef, JSON.stringify(newError));
```
rajsite: Since we know the structure of an error can we do use the more optimized eggShell functions instead?
findSubValueRef for the different components (status, code, source) and write to them using the correct eggShell function (writeDouble, writeDouble, writeString)

source/core/module_coreHelpers.js
gleonoliva: I'll remove it.

source/core/module_coreHelpers.js
sanmut: +1

source/io/module_eggShell.js
```diff
@@ -479,6 +483,27 @@
             }
         };
 
+        Module.eggShell.writeStringFromArray = function (valueRef, inputArray) {
```
rajsite: Talked with memo and instead we will create a `writeTypedArray` function that is more generic for writing arbitrary TypedArrays to memory and that it will support String types.

Will include a comment saying that it can be expanded for more types in the future as needed

source/io/module_httpClient.js
```diff
@@ -760,24 +870,35 @@
...
-            var httpClient = findhttpClientOrWriteError(handle, errorStatusPointer, errorCodePointer, errorSourcePointer);
+        Module.httpClient.jsHttpClientConfigCORS = function (
+            handleTypeRef, handleDataRef,
+            includeCredentialsDuringCORSTypeRef, includeCredentialsDuringCORSDataRef,
+            errorTypeRef, errorDataRef) {
+            var handleValueRef = Module.eggShell.createValueRef(handleTypeRef, handleDataRef);
+            var includeCredentialsDuringCORSValueRef = Module.eggShell.createValueRef(includeCredentialsDuringCORSTypeRef, includeCredentialsDuringCORSDataRef);
+            var errorValueRef = Module.eggShell.createValueRef(errorTypeRef, errorDataRef);
+            var handle = Module.eggShell.readDouble(handleValueRef);
+            var includeCredentialsDuringCORS = Module.eggShell.readDouble(includeCredentialsDuringCORSValueRef);
```
rajsite: Can we move the !== FALSE check up here? That way includeCredentialsDuringCORS can just be a boolean

sanmut:
BTW, great job on refactoring and cleaning up. This is looking great.

source/core/module_coreHelpers.js
rajsite: I had organized it so all the error checks are done in a group followed by all the warning checks (get info needed and then perform logic).

I thought mixing the acquiring of needed state and logic using the state was more confusing. Also var isNewError = newError.status is very little work, so it did not really add overhead.

It's just a stylistic choice to me, @sanmut if it's still clear and you prefer that style that's fine.

source/io/HttpClient.cpp
```diff
@@ -88,16 +135,17 @@ VIREO_FUNCTION_SIGNATURE6(HttpClientOpen, StringRef, StringRef, StringRef, UInt3
         return THREAD_EXEC()->Stop();
     }
 
+    TypeRef typeRefUInt32 = TypeManagerScope::Current()->FindType("UInt32");
+    TypeRef typeRefErrorCluster = TypeManagerScope::Current()->FindType("ErrorCluster");
```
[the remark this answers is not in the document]
cglzaguilar: I was looking at "TypeManager::New" and "TypeDefiner::DefineStandardTypes", perhaps we can come up with a caching mechanism to access the most commonly used type definitions. But I think that would be part of another pull request and not this one.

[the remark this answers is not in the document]
cglzaguilar: Also we can use StaticTypeAndData type on the Vireo function definition. That will give you access to both the type and a pointer to the data. You might want to do one function to see how it will look. The problem with that is that the Vireo definition for the instruction will allow more types than intended. Again this is more likely a refactor to do in another pull request.

source/io/module_httpClient.js
gleonoliva: ah yeah, I'll change it.

source/io/HttpClient.cpp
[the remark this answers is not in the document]
gleonoliva: I agree we can continue refactoring but I think this is a good chunk of work for this PR.

source/core/module_coreHelpers.js
```diff
@@ -303,37 +303,40 @@
...
+                writeNewError(errorValueRef, newError);
                 return;
             }
 
-            var existingErrorCode = Module.eggShell.dataReadInt32(existingErrorCodePointer);
+            var existingErrorCodeValueRef = Module.eggShell.findSubValueRef(errorValueRef, 'code');
+            var existingErrorCode = Module.eggShell.readDouble(existingErrorCodeValueRef);
             var isExistingWarning = existingErrorCode !== CODES.NO_ERROR;
-            var isNewWarning = newErrorCode !== CODES.NO_ERROR;
+            var isNewWarning = newError.code !== CODES.NO_ERROR;
```
rajsite: To keep consistency move this down to right before the isNewWarning check

source/core/module_coreHelpers.js
sanmut: Not a biggie. It is generally a good convention to declare variables as needed. If it helps with clarity,  then exceptions are ok.

PR #512, opened 2018-09-24
Switching Vireo to Wasm from asm.js

rajsite:
Major changes:
- Updating emscripten version
- Switching to ES6 Module format for JavaScript source
- Switching from a synchronous constructor (`new Vireo()`) to asynchronous (`vireoHelpers.createInstance()`)
- Change `executeSlicesUntilClumpsFinished` from being callback-based to being Promise-based

Note: It looks like a lot of files changes buts that mostly because two lines had to change in every JS test.
It was unavoidable due WebAssembly requiring async construction.

The actual changes are pretty well localized to the build scripts and the loader js files.


test-it/karma/helloworld/HelloWorld.Test.js
```diff
@@ -28,9 +28,26 @@ describe('Vireo loaded as a global in the browser', function () {
         expect(result).toBe('Hello, sky. I can fly.\n');
     });
 
-    it('can run HelloWorld Async', function (done) {
-        var Vireo = window.NationalInstruments.Vireo.Vireo;
-        var vireo = new Vireo();
+    it('can run HelloWorld async using a callback', async function () {
```
gleonoliva: Cannot? Or perhaps test says that now it throws?

karma.coverage.conf.js
```diff
@@ -1,37 +1,31 @@
...
         ];
 
         sharedConfig.reporters = [
             'dots',
             'coverage'
         ];
 
-        sharedConfig.preprocessors['source/core/*.js'] = ['coverage'];
-        sharedConfig.preprocessors['source/io/*.js'] = ['coverage'];
+        sharedConfig.preprocessors['dist/wasm32-unknown-emscripten/release/vireo.js'] = ['coverage'];
```
mehagar: Does this mean we're generating a code coverage report for the karma tests?

package.json
```diff
@@ -14,22 +13,25 @@
...
+    "terser": "^3.8.2",
     "xhr2": "^0.1.4"
   },
   "scripts": {
+    "rollup": "cd make-it && rollup",
+    "terser": "cd make-it && terser",
     "test": "karma start --browsers ChromeHeadlessNoSandbox,FirefoxHeadless",
+    "test-min": "karma start karma.min.conf.js --browsers ChromeHeadlessNoSandbox,FirefoxHeadless",
+    "test-debug": "karma start karma.debug.conf.js --browsers ChromeHeadlessNoSandbox,FirefoxHeadless",
     "test-dev": "karma start --browsers ChromeHeadlessNoSandbox --skip-tags Slow",
```
mehagar: Should this have FirefoxHeadless like the others?

test-it/ManualLoaderTests/loaderglobal_currentbrowserfps.js
```diff
@@ -11,9 +11,10 @@
 
     var runTest = function () {
         var viaCode = document.getElementById('viacode').textContent;
-        var vireo = new window.NationalInstruments.Vireo.Vireo();
-        vireo.eggShell.loadVia(viaCode);
-        vireo.eggShell.executeSlicesUntilClumpsFinished(function () {
+        window.vireoHelpers.createInstance().then(function (vireo) {
```
mehagar: Can we use the "await" keyword here to avoid having to string "then"s together?

test-it/karma/utilities/TestHelpers.HttpParser.js
```diff
@@ -35,7 +35,7 @@
     var parseUrlElement = document.createElement('a');
 
     var parseUrl = function (url) {
-        // Add and remove the element to workaround IE https://connect.microsoft.com/IE/Feedback/Details/1002846
+        // Add and remove the element to workaround behavior in legacy browsers, no longer required
```
mehagar: Could you add a vireo issue to remove the instances where you said a section is no longer required.

karma.coverage.conf.js
rajsite: Only if you run the `npm run test-coverage` command

package.json
rajsite: nah, `npm run test-dev` is intended for local development time as a quick way to run a subset of tests. It intentionally skips slow tests and runs only in one browser.

test-it/karma/utilities/TestHelpers.HttpParser.js
rajsite: I don't think we need a specific issue to address these. None of the spots where I made the change are incorrect, just add extra checks, etc, but in the future if someone is curious why some behavior is there it is documented.

test-it/ViaTests/HttpBinGet.via
```diff
@@ -18,12 +18,13 @@ define(MyVI dv(VirtualInstrument (
...
             ) args)
         ) clusterRepresentation)
+        e(a(.String *) path)
     )
     clump (
         HttpClientGet(handle url outputFile timeout headers body statusCode error)
 
         // string lvtype pathIntoJSON lvExtensionsBool strictValidationBool defaultNullBool errorCluster
-        UnflattenFromJSON(body clusterRepresentation * false false false error)
+        UnflattenFromJSON(body clusterRepresentation path false false false error)
```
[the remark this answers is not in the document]
rajsite: Tested in the editor and DFIR seems to emit an empty array for unwired inputs. Created the following issue: https://github.com/ni/VireoSDK/issues/513

test-it/ManualLoaderTests/loaderglobal_currentbrowserfps.js
rajsite: done, definitely cleaner

test-it/karma/helloworld/HelloWorld.Test.js
rajsite: good catch! updated the test name

test-it/ManualLoaderTests/loaderglobal_httpget.html
```diff
@@ -0,0 +1,34 @@
...
+    clump (
+        HttpClientGet(handle url outputFile timeout headers body statusCode error)
+    )
+) ) )
+
+enqueue(MyVI)
+//Finished!! :D
+</script>
+</body>
+</html>
```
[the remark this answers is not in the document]
rajsite: Updated

spathiwa:
This looks great!
But is there a good reason to get rid of the asm.js target completely and move to wasm in one go?
It seems like keeping both targets would be good at least initially; if we find performance or behavior changes it might be useful to build the asm.js version and compare.

source/io/module_eggShell.js
```diff
@@ -1019,7 +1020,12 @@
                 }
             };
 
-            runExecuteSlicesAsync();
+            // Queue a microtask for starting execution
+            Promise.resolve().then(function () {
```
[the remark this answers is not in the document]
rajsite: The microtask will get called at the end of the current JS callstack, so roughly soon after executeSlicesUntilClumpsFinished is finished.

I ran runExecuteSlicesAsync as a microtask because that function has the ability to complete and resolve synchronously. That would make two code paths possible, one where it resolves synchronously in the current JS stack or when asynchronously in a new stack.

I made the change so no one can accidently rely on synchronous execution behavior from executeSlicesUntilClumpsFinished

rajsite:
@spathiwa Initially had tried to support both asm.js and wasm with the old UMD module system but that was very complex. After switching ES6 modules it simplified a lot of pieces and it wasn't too bad to have both wasm and asm.js builds supported so both are now enabled.

The asm.js build also revealed a problem that existed in the tests and that mainline relied on and is now cleaned up so it was a nice bit that was uncovered.

I'm resetting @cglzaguilar and @gleonoliva to buddy that api behavior change.

sanmut:
> The asm.js build also revealed a problem that existed in the tests and that mainline relied on and is now cleaned up so it was a nice bit that was uncovered.

This is awesome. Thanks for building both.

make-it/EmMakefile
```diff
@@ -87,7 +89,7 @@ ifeq ($(TARGET), wasm32-unknown-emscripten)
 endif
 
 # NO_DYNAMIC_EXECUTION=1 Allows Vireo to be used in CSP environments where eval() is forbidden: https://github.com/kripken/emscripten/blob/master/src/settings.js#L563
-EM_OPT = $(EM_OPTFLAGS) $(WASM_FLAG) -s NO_EXIT_RUNTIME=1 -std=c++14 -fno-exceptions --memory-init-file 0 -s NO_DYNAMIC_EXECUTION=1
+EM_OPT = $(EM_OPTFLAGS) $(WASM_FLAG) -s NO_EXIT_RUNTIME=1 -std=c++14 -fno-exceptions --memory-init-file 0 -s NO_DYNAMIC_EXECUTION=1 -s MODULARIZE=1 -s EXPORT_NAME=VireoCreateCoreModule -s EXPORT_ES6=1 --minify 0 -s BINARYEN_TRAP_MODE=js -s ALLOW_MEMORY_GROWTH=1
```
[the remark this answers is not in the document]
rajsite: I tried to use clamp mode in this commit: https://github.com/ni/VireoSDK/pull/512/commits/888c4dda238085c8994b3a25b56089bfd1bf7f41

It did cause tests to fail. It's probably worth looking into more in the future when we look closer at perf.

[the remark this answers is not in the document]
rajsite: The default starting memory is 16MB and I did not override that or make it explicit.

I did get to test it and got a gviweb to allocate 2GB of space before crashing (wasm is 32bit)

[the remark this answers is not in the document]
sanmut: Got it, thanks.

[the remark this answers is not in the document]
sanmut: Sweet :+1:

gleonoliva:
Also I just noticed that this is the 2^9 pull request to VireoSDK 🙌 🎉

rajsite:
Ran the tests suggested by @sanmut just waiting on ASW PR

PR #514, opened 2018-09-25
Expanded type support for `writeTypedArray`

gleonoliva:
In #511 I added `writeTypedArray` to vireo's module and didn't expose it to the public API.
In this PR I'm adding the rest of supported TypedArrays.
This work will be needed for refactoring JSInvoke part that @sanmut is working on.

### DIFF
--- a/source/io/module_eggShell.js
+++ b/source/io/module_eggShell.js
@@ -573,18 +573,16 @@
             return typedArray;
         };
 
-        Module.eggShell.writeTypedArray = function (valueRef, typedArrayValue) {
+        Module.eggShell.writeTypedArray = publicAPI.eggShell.writeTypedArray = function (valueRef, typedArrayValue) {
             var TypedArrayConstructor = findCompatibleTypedArrayConstructor(valueRef.typeRef);
-
-            // TODO: (gleon) For now this function will only work for writing to strings
-            // we can support more types by modifying this function moving forward.
-            if (TypedArrayConstructor === undefined || Module.typeHelpers.isString(valueRef.typeRef) === false) {
+            if (TypedArrayConstructor === undefined || !(typedArrayValue instanceof TypedArrayConstructor)) {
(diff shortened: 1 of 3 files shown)


source/io/module_eggShell.js
```diff
@@ -573,18 +573,16 @@
...
-            // we can support more types by modifying this function moving forward.
-            if (TypedArrayConstructor === undefined || Module.typeHelpers.isString(valueRef.typeRef) === false) {
+            if (TypedArrayConstructor === undefined || !(typedArrayValue instanceof TypedArrayConstructor)) {
                 throw new Error('Performing writeTypedArray failed for the following reason: ' + eggShellResultEnum[EGGSHELL_RESULT.UNEXPECTED_OBJECT_TYPE] +
                     ' (error code: ' + EGGSHELL_RESULT.UNEXPECTED_OBJECT_TYPE + ')' +
                     ' (typeRef: ' + valueRef.typeRef + ')' +
                     ' (dataRef: ' + valueRef.dataRef + ')');
             }
 
+            // TODO: (gleon) For now this function will only work for 1D arrays
```
rajsite: Here is one proposal:
I think the correct way for someone to use writeTypedArray is that they first called Module.eggShell.resizeArray to set the dimensionality and then call writeTypedArray to set the value of the whole array on one go.

I don't think writeTypedArray should have to worry about dimensionality (because JS TypedArrays don't have dimensionality, that is a LV thing in this context).

So writeTypedArray should not call resizeArray, instead it should just validate that the array being written to is the exact correct size and throw an error if it is not.

rajsite: And that should be a pretty quick tweak since we have a helper to get the full array size already: https://github.com/ni/VireoSDK/blob/4cf223972ebdcfce9fb92565265527a550e66083/source/io/module_eggShell.js#L571

gleonoliva: Aaaah, yeah I like that. I'll change it.

PR #515, opened 2018-09-26
Adding support for JavaScriptRefNum to events

segaljared:
Adding support for JavaScriptRefNums to the events code. Right now it doesn't remove support for using ControlRefNums, which means they can both be used depending on which feature is enabled in LV *and* this can be used without the changes VIA generation, since it won't break existing functionality.

Currently, I haven't create tests on the GitHub side for this, but I will be adding them as the review goes on.

I will add an example of what the VIA would like as a comment.

### DIFF
--- a/source/core/Events.cpp
+++ b/source/core/Events.cpp
@@ -549,6 +549,7 @@ static inline EventSource GetEventSourceForEventType(EventType eType) {
     return eSource;
 }
 
+// TODO(segaljared): Move this to a more convenient location for general use in the future
 void GetVIName(VirtualInstrument *vi, StringRef viName) {
     PercentEncodedSubString encodedStr(vi->VIName(), true, false);
     SubString encodedSubstr = encodedStr.GetSubString();
--- a/source/core/Events.cpp
+++ b/source/core/Events.cpp
@@ -608,7 +608,7 @@ void RegisterForStaticEvents(VirtualInstrument *vi) {
     vi->SetEventInfo(eventInfo);
 }
 
-void ConfigureEventSpecForJSRef(VirtualInstrument *vi, UInt32 eventStructIndex, UInt32 eventSpecIndex, JavaScriptRefNum jsReference) {
+void ConfigureEventSpecForJSRef(VirtualInstrument *vi, Int32 eventStructIndex, Int32 eventSpecIndex, JavaScriptRefNum jsReference) {
     TypedObjectRef eventStructSpecsRef = vi->EventSpecs();
     Int32 numEventStructs = eventStructSpecsRef->ElementType()->SubElementCount();
     EventInfo *eventInfo = vi->GetEventInfo();
@@ -1210,10 +1210,10 @@ VIREO_FUNCTION_SIGNATURE3(_OccurEvent, RefNumVal, UInt32, UInt32)
     return _NextInstruction();
 }
 
-VIREO_FUNCTION_SIGNATURE3(ConfigureEventSpecJSRef, UInt32, UInt32, JavaScriptRefNum)
+VIREO_FUNCTION_SIGNATURE3(ConfigureEventSpecJSRef, Int32, Int32, JavaScriptRefNum)
 {
-    UInt32 eventStructIndex = _Param(0);
-    UInt32 eventSpecIndex = _Param(1);
+    Int32 eventStructIndex = _Param(0);
+    Int32 eventSpecIndex = _Param(1);
     JavaScriptRefNum jsRefNum = _Param(2);
     VirtualInstrument *owningVI = THREAD_CLUMP()->OwningVI();
     ConfigureEventSpecForJSRef(owningVI, eventStructIndex, eventSpecIndex, jsRefNum);
@@ -1294,8 +1294,8 @@ DEFINE_VIREO_BEGIN(Events)
     DEFINE_VIREO_FUNCTION(WaitForEventsAndDispatch, "p(i(VarArgCount) i(Int32 timeOut) i(StaticTypeAndData ref) i(Int32 esIndex) "
                           "i(VarArgRepeat) i(Int32 specIndex)i(StaticTypeAndData)i(BranchTarget))")
 
-    DEFINE_VIREO_FUNCTION(_OccurEvent, "p(i(JavaScriptRefNum controlRef) i(UInt32 eSource) i(UInt32 eType))")
-    DEFINE_VIREO_FUNCTION(ConfigureEventSpecJSRef, "p(i(UInt32 eStructIndex) i(UInt32 eSpecIndex) i(JavaScriptRefNum jsReference))")
+    DEFINE_VIREO_FUNCTION(_OccurEvent, "p(i(ControlRefNum controlRef) i(UInt32 eSource) i(UInt32 eType))")
+    DEFINE_VIREO_FUNCTION(ConfigureEventSpecJSRef, "p(i(Int32 eStructIndex) i(Int32 eSpecIndex) i(JavaScriptRefNum jsReference))")
     DEFINE_VIREO_FUNCTION(RegisterForJSEvent, "p(i(JavaScriptRefNum jsReference) i(UInt32 eType))")
 
     DEFINE_VIREO_FUNCTION_CUSTOM(IsNotANumPathRefnum, IsNotAUserEventRefnum, "p(i(UserEventRefNum) o(Boolean))")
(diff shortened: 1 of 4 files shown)


segaljared:
Here is an example VIA file that I hand edited(The key additions are the new ConfigureEventSpecJSRef and RegisterForJSEvent):

// Autogenerated Vireo assembly file
// Generated on 09/26/2018 10:06:00
//VI Definition: ::Web Server::Interactive::Application::Function.gviweb
define (Application%3A%3AFunction%2Egviweb dv(.VirtualInstrument (
    Events: c(
        e(c( // Event Struct 1
            e(dv(c(
                e(.UInt32 eventSource)
                e(.UInt32 eventType)
                e(.UInt32 controlUID)
                e(.UInt32 dynamicIndex)
            ) (1000 2 2 0 ))local4) // passing in an int here let's us get the control ID, since we can't get it from the JavaScriptRefNum
        ))
    )
    Locals: c(   // Data Space
        e(dv(.Double 0)dataItem_Numeric)
        e(.Double local6)
        e(c(
            e(.Double Old%20Value)
            e(.Double New%20Value)
        ) %7BF3226E31-E688-4C58-AE1E-7ED6718409A1%7D_2)
        e(.Int32 local10)
        e(c(
            e(.UInt32 Source)
            e(.UInt32 Type)
            e(.UInt32 Time)
            e(.UInt32 Index)
            e(.JavaScriptRefNum Control%20Reference)
            e(.Double Old%20Value)
            e(.Double New%20Value)
        ) local18)
        e(c(
            e(.UInt32 Source)
            e(.UInt32 Type)
            e(.UInt32 Time)
            e(.UInt32 Index)
            e(.JavaScriptRefNum Control%20Reference)
            e(.Double Old%20Value)
            e(.Double New%20Value)
        ) local26)
        e(.JavaScriptRefNum local27)
        e(.Double local28)
        e(.JavaScriptRefNum local29)
        e(.Boolean local30)
        e(dv(.Boolean false)dataItem_XY)
        e(.Boolean local32)
        ce(dv(.Boolean true)c33)
        e(dv(.Boolean false)dataItem_Done)
        e(.Boolean local35)
        ce(dv(.Int32 0)c36)
        ce(dv(.Int32 0)c37)
		e(dv(.String 'Application::EventStructureLoop.gviweb')viName)
		e(dv(.Int32 2)controlID)
        e(.Occurrence occurrence)
        e(.ErrorCluster error)
    )
        clump(1
        JavaScriptInvoke(occurrence true error 'ControlReference_GetControlObject' local29 viName controlID)
        ConfigureEventSpecJSRef(0 0 local29)
        RegisterForJSEvent(local29 2)
        WaitForEventsAndDispatch(* * c36 c37 local18 1)
        Branch(0)
        Perch(1)
        Copy(local18 local26)
        Copy(local26.Control%20Reference local27)
        Copy(local26.New%20Value local28)
        IsEQ(local29 local27 local30)
        Copy(local30 dataItem_XY)
        Copy(c33 dataItem_Done)
        Branch(0)
        Perch(0)
/* Clump Ended. */    )
)))
enqueue (Application%3A%3AFunction%2Egviweb)
//Finished!! :D

source/core/Events.cpp
```diff
@@ -548,6 +549,30 @@ static inline EventSource GetEventSourceForEventType(EventType eType) {
     return eSource;
 }
 
+void GetVIName(VirtualInstrument *vi, StringRef viName) {
```
segaljared: Adding a helper since we do this in a few different places in the file

source/core/Events.cpp
```diff
@@ -548,6 +549,30 @@ static inline EventSource GetEventSourceForEventType(EventType eType) {
...
+    PercentEncodedSubString encodedStr(vi->VIName(), true, false);
+    SubString encodedSubstr = encodedStr.GetSubString();
+    viName->AppendSubString(&encodedSubstr);
+}
+
+void RegisterForControlEvent(EventInfo *eventInfo, EventSpec eventSpec, StringRef viName, EventQueueID qID, UInt32 controlID, RefNum reference, bool registerControlEvent) {
```
segaljared: Similar helper as above. This will end up calling jsRegisterForcontrolEvent for ControlRefNums, but not JavaScriptRefNums. We may be able to get away without the RegisterForJSEvent instruction if we just always attempt to register here. It should still only do this once like for ControlRefNums.

source/core/Events.cpp
```diff
@@ -548,6 +549,31 @@ static inline EventSource GetEventSourceForEventType(EventType eType) {
     return eSource;
 }
 
+void GetVIName(VirtualInstrument *vi, StringRef viName) {
+    PercentEncodedSubString encodedStr(vi->VIName(), true, false);
+    SubString encodedSubstr = encodedStr.GetSubString();
+    viName->AppendSubString(&encodedSubstr);
+}
+
+void RegisterForControlEvent(EventInfo *eventInfo, EventSpec eventSpec, StringRef viName, EventQueueID qID, UInt32 controlID, RefNum reference,
```
segaljared: Adding another helper for registering the events since it's used in multiple places. For ControlRefNums, this will end up calling jsRegisterForControlEvent. We may be able to do this for JavaScriptRefNum too and avoid having a separate RegisterForJSEvent instruction call.

source/core/Events.cpp
```diff
@@ -597,6 +608,30 @@ void RegisterForStaticEvents(VirtualInstrument *vi) {
     vi->SetEventInfo(eventInfo);
```
gleonoliva: Do we need to free eventInfo before exiting the method?

gleonoliva: Nevermind, I see that the owner becomes the VI.

source/core/Events.cpp
```diff
@@ -1223,6 +1295,8 @@ DEFINE_VIREO_BEGIN(Events)
                           "i(VarArgRepeat) i(Int32 specIndex)i(StaticTypeAndData)i(BranchTarget))")
 
     DEFINE_VIREO_FUNCTION(_OccurEvent, "p(i(ControlRefNum controlRef) i(UInt32 eSource) i(UInt32 eType))")
+    DEFINE_VIREO_FUNCTION(ConfigureEventSpecJSRef, "p(i(Int32 eStructIndex) i(Int32 eSpecIndex) i(JavaScriptRefNum jsReference))")
```
cglzaguilar: I was thinking about ConfigureEventSpec and RegisterForEvent to make them more generic and possibly useful in other targets, but both functions take a JavaScriptRefNum, so maybe not a good idea. What do you think?

sanmut:
Talked to Jared about this. Waiting for Jared to add tests here or post a PR in the ASW side with the generated VIA output from Gandi's that shows events pass with JSRefnum/JSInvoke.

source/core/Events.cpp
segaljared: Right now it doesn't look we don't allow using just RefNum in these Vireo function definitions, but I can try adding that and see if it's okay taking a generic RefNum

segaljared: The other option there is to add other versions of this definition here that take other RefNum types when we want to support other RefNum types here

spathiwa: I think adding other definitions as necessary is preferable; it's likely there would be subtle differences in implementation based on the refnum type anyway.

segaljared:
I've added a test for JavaScriptRefNum events to our karma test for control events. I wanted to have the test run the VIA file twice in a row, but have been having issues with getting that to work. I don't want to hold this up too long on adding that test functionality, so I've put in the single run version of the test.

spathiwa:
For the record, I'm strongly against retiring ControlRefnum; we might have other backends beside JS i n the future that could still use this infrastructure, and I think the architecture is cleaner to have a separate refnum type for control refnums than conflating this with jsinvoke refnums.

sanmut:
> For the record, I'm strongly against retiring ControlRefnum; we might have other backends beside JS i n the future that could still use this infrastructure, and I think the architecture is cleaner to have a separate refnum type for control refnums than conflating this with jsinvoke refnums.

Then, I think we should use # if directives for supporting ControlRefnum and JavascriptRefnum separately and enable the right ones only in the builds that use.

source/core/Events.cpp
```diff
@@ -548,6 +549,31 @@ static inline EventSource GetEventSourceForEventType(EventType eType) {
     return eSource;
 }
 
+void GetVIName(VirtualInstrument *vi, StringRef viName) {
```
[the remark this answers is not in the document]
segaljared: Since this is holding a merge back and can probably be addressed later since it would only be useful once someone else wants to use a similar helper, I'm going to add a TODO for now about moving this.

PR #517, opened 2018-09-29
Use JSinvoke for propertynode

sanmut:


source/io/module_eggShell.js
```diff
@@ -205,7 +214,7 @@
 
         Module.eggShell.createValueRef = function (typeRef, dataRef) {
             if (typeof typeRef !== 'number' || typeof dataRef !== 'number' ||
-                typeRef <= 0 || dataRef <= 0) {
+                (typeRef <= 0 && dataRef <= 0)) {
```
sanmut: @gleonoliva @rajsite : The reason I changed this is the following: There are valid situations where typeRef is not null and valueRef can be null (for example, when wild card argument is passed for a function in VIA). After createValueRef() is called, I would like to still be able to use valueRef.typeRef in those situations without exceptions thrown.

sanmut: OTOH, if dataRef is nulll in Vireo (0 here in JS), we could set dataRef to undefined and leave typeRef to contain valid value. Would you rather prefer that as opposed to leaving dataRef as 0?

cglzaguilar: Sankara, both appveyor and travis-ci failed. We need to look into these errors.

make-it/EmMakefile
```diff
@@ -166,7 +166,10 @@ EM_EXPORTS = -s EXPORTED_FUNCTIONS="[\
     '_TypeRef_IsTimestamp',\
     '_TypeRef_IsComplex',\
     '_TypeRef_IsAnalogWaveform',\
-    '_JavaScriptInvoke_GetParameterType',\
+    '_TypeRef_IsJavascriptRefnum',\
+    '_JavaScriptInvoke_GetParameterTypeName',\
```
rajsite: JavaScriptInvoke_GetParameterTypeName is no longer needed correct? we should remove it.

source/io/JavaScriptInvoke.cpp
```diff
@@ -46,7 +46,7 @@ StringRef AllocateReturnBuffer()
 
 //------------------------------------------------------------
 //! Return parameter type name for the given element(index) in the parameters array
-VIREO_EXPORT const char* JavaScriptInvoke_GetParameterType(StaticTypeAndData *parameters, Int32 index)
+VIREO_EXPORT const char* JavaScriptInvoke_GetParameterTypeName(StaticTypeAndData *parameters, Int32 index)
```
rajsite: JavaScriptInvoke_GetParameterTypeName is no longer needed correct? we should remove it.

source/io/module_javaScriptInvoke.js
```diff
@@ -68,7 +68,9 @@
         // Disable new-cap for the cwrap functions so the names can be the same in C and JS
         /* eslint 'new-cap': ['error', {'capIsNewExceptions': [
             'JavaScriptInvoke_GetParameterPointer',
-            'JavaScriptInvoke_GetParameterType',
+            'JavaScriptInvoke_GetParameterTypeName',
```
rajsite: JavaScriptInvoke_GetParameterTypeName is no longer needed correct? we should remove it.

source/io/module_javaScriptInvoke.js
```diff
@@ -79,28 +81,26 @@
 
         // Private Instance Variables (per vireo instance)
         var internalFunctionsMap = new Map();
-        var JavaScriptInvoke_GetParameterType = Module.cwrap('JavaScriptInvoke_GetParameterType', 'number', ['number', 'number']);
-        var JavaScriptInvoke_GetParameterPointer = Module.cwrap('JavaScriptInvoke_GetParameterPointer', 'number', ['number', 'number']);
-        var JavaScriptInvoke_GetArrayElementType = Module.cwrap('JavaScriptInvoke_GetArrayElementType', 'number', ['number']);
+        var JavaScriptInvoke_GetParameterTypeRef = Module.cwrap('JavaScriptInvoke_GetParameterTypeRef', 'number', ['number', 'number']);
```
rajsite: Because these are only numerics in and out, we can avoid the overhead of using cwrap. You can call JavaScriptInvoke_GetParameterTypeRef by doing:

```
var resultNum = Module._JavaScriptInvoke_GetParameterTypeRef(num1, num2);
```

Note the c function name is prefixed with an underscore when accessed directly from the Module object.

source/io/module_javaScriptInvoke.js
```diff
@@ -79,28 +81,26 @@
...
+        var JavaScriptInvoke_GetParameterDataRef = Module.cwrap('JavaScriptInvoke_GetParameterDataRef', 'number', ['number', 'number']);
```
rajsite: Same as above, at the callsite use Module._JavaScriptInvoke_GetParameterDataRef instead

source/io/module_javaScriptInvoke.js
rajsite: Similar to how you did the TypeRef call: https://github.com/ni/VireoSDK/blob/aa4422ca3aa849cefdb1442fd09f55f016545845/source/core/module_typeHelpers.js#L288

source/core/module_typeHelpers.js
```diff
@@ -278,7 +284,15 @@
             return Module._TypeRef_IsAnalogWaveform(typeRef) !== 0;
         };
 
+        Module.typeHelpers.isJSObjectRefnum = function (typeRef) {
+            return Module._TypeRef_IsJavascriptRefnum(typeRef) !== 0;
+        };
+
         var typeHandlers = [
+            {
```
rajsite: some items in this list have significance in ordering, is there any significance in ordering for isJSObjectRefnum? If so can we document that

rajsite: @gleonoliva Maybe we should document each entry? Just from looking it is not obvious where ordering is significant.

source/io/module_eggShell.js
```diff
@@ -70,6 +70,15 @@
         publicAPI.eggShell = {};
 
         // Private Instance Variables (per vireo instance)
+
+        Module.eggShell.readJavaScriptRefNum = function (valueRef) {
```
rajsite: This should also be exposed on publicAPI.eggShell.readJavaScriptRefNum and publicAPI.eggShell.writeJavaScriptRefNum so the ASW side can update those kind of valueRefs

source/io/module_eggShell.js
rajsite: Of the options described the second one sounds better. Hard check that typeRef <=0 but when dataRef <= 0 set the valueRef.dataRef to undefined.

If we don't leak DOM "null" values in ASW we shouldn't leak Vireo null values (zeroes) in Travis. Coerce nulls to undefined.

---

Out of curiosity, what is the situation where you need to know the type of a unwired input? Why does this need to loosen assertions to account for this case?

source/io/module_javaScriptInvoke.js
```diff
@@ -155,158 +155,144 @@
...
-                    return value instanceof Uint32Array;
-                }
-            },
-            ArraySingle: {
-                reader: Module.eggShell.dataReadTypedArray,
-                writer: Module.eggShell.dataWriteTypedArray,
-                isValidReturnType: function (value) {
-                    return value instanceof Float32Array;
+        var jsInvokePeeker = function () {
+            var PeekVisitor = function () {
```
rajsite: There are some unnecessary gymnastics going on here. Go through the effort of making a prototype chain just to make a singleton is kinda weird. The constructor doesn't even do anything.

Can we just do:

```
var createPeeker = function () {
  var visitNumeric = ...

  return {
    visitBoolean = function (valueRef) {
    },
    visitInt8: visitNumeric,
    visitInt16: visitNumeric
    ...
  };
};
```

A simple module pattern

source/io/JavaScriptInvoke.cpp
cglzaguilar: Consider deleting JavaScriptInvoke_GetParameterTypeName and JavaScriptInvoke_GetParameterPointer. They appear not to be used anymore.

source/io/module_javaScriptInvoke.js
```diff
@@ -338,6 +324,19 @@
             };
         };
 
+        var createJavaScriptParametersArray = function (isInternalFunction, parametersPointer, parametersCount) {
+            var parameters = new Array(parametersCount);
+            for (var index = 0; index < parametersCount; index += 1) {
+                var parameterValueRef = createValueRefFromPointerArray(parametersPointer, index);
+                if (!isInternalFunction) {
```
cglzaguilar: I was wondering if we need to pass a ValueRef to all internal functions present and future. I think it is OK. It will make for a more efficient API for internal functions.

source/io/module_javaScriptInvoke.js
```diff
@@ -349,65 +348,26 @@
...
-            var typeConfig = typeFunctions[returnTypeName];
-            var newError = {
-                status: true
+            var valueAndError = {
+                userValue: returnUserValue,
+                errorValueRef: errorValueRef,
+                functionName: functionName
             };
-
-            if (typeConfig === undefined) {
```
cglzaguilar: What is the jsInvokePeeker behavior when dealing for a type if doesn't know? This is the behavior this validation is doing. Can we incorporate it in the jsInvokePeeker?
Note: This is mostly for handwritten VIA. DFIR doesn't generate types that we don't know.

source/io/module_javaScriptInvoke.js
rajsite: Similar discussion for the poker

source/io/module_javaScriptInvoke.js
```diff
@@ -155,158 +155,144 @@
...
-            for (var index = 0; index < parametersCount; index += 1) {
-                var typeName = getParameterTypeString(parametersPointer, index);
-                var parameterPointer = JavaScriptInvoke_GetParameterPointer(parametersPointer, index);
-                if (typeName === 'Array') {
-                    typeName += getArrayElementTypeString(parameterPointer);
+            proto.visitArray = function (valueRef, data) {
+                try {
+                    Module.eggShell.resizeArray(valueRef, [data.userValue.length]);
+                    Module.eggShell.writeTypedArray(valueRef, data.userValue);
+                    return;
```
rajsite: unneccessary return? I guess just make it consistent with visitJSObjectRefnum

source/io/module_javaScriptInvoke.js
```diff
@@ -338,6 +324,19 @@
             };
         };
 
+        var createJavaScriptParametersArray = function (isInternalFunction, parametersPointer, parametersCount) {
+            var parameters = new Array(parametersCount);
```
rajsite: just do `var parameters = [];`

Using the Array constructor does not actually allocate any space. It creates a sparse array and is ultimately more difficult for JS engines to optimize than growing a normal array.

source/core/module_typeHelpers.js
gleonoliva: I'm not sure if it necessary to add description to every single entry, it would be nice though.

source/io/module_eggShell.js
```diff
@@ -70,6 +70,15 @@
         publicAPI.eggShell = {};
 
         // Private Instance Variables (per vireo instance)
```
gleonoliva: This comment should go with variable `NULL`, `POINTER_SIZE` and `DOUBLE_SIZE`

source/io/module_javaScriptInvoke.js
```diff
@@ -155,158 +155,144 @@
...
-                reader: Module.eggShell.dataReadTypedArray,
-                writer: Module.eggShell.dataWriteTypedArray,
-                isValidReturnType: function (value) {
-                    return value instanceof Float64Array;
+                Module.eggShell.writeDouble(valueRef, data.userValue ? 1 : 0);
+            };
+
+            var visitNumeric = function (valueRef, data) {
+                // TODO-san anything to do for this?
+                // Looks like the data-grid ends up pushing the string value "-Infinity"
```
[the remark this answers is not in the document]
gleonoliva: I agree I don't think this function should care about how data-grid is pushing values to vireo. Sounds like something to be fixed on the other side (ASW)

source/io/module_javaScriptInvoke.js
```diff
@@ -349,65 +348,26 @@
...
-        var updateReturnValue = function (
-            functionName,
-            returnTypeName,
-            returnValuePointer,
-            returnUserValue,
-            errorValueRef) {
-            if (returnTypeName === 'StaticTypeAndData') {
-                // User doesn't want return value. If we're passing '*' for the return in VIA code, we get StaticTypeAndData
+        var updateReturnValue = function (functionName, returnValueRef, returnUserValue, errorValueRef) {
+            if (returnValueRef.dataRef === 0) {
```
gleonoliva: Ah I get it. So this would be true if the user configures the JSLI call to not return. So we say, even if ultimately the invoked function returned something in `returnUserValue` we are just going to ignore it, right?
Please use the variable `NULL` here to indicate that comes from C++ or if you decide to change the behavior of `createValueRef` to return `undefined` then don't forget to update this one.

source/io/module_javaScriptInvoke.js
```diff
@@ -474,6 +426,10 @@
             });
         };
 
+        var shouldUpdateReturnValue = function (isInternalFunction, returnTypeRef) {
+            return !isInternalFunction || Module.typeHelpers.isJSObjectRefnum(returnTypeRef);
```
rajsite: I'm a little confused here. If the function is internal we should not update the return value.

For internal functions the user is passed a bunch of valueRefs. They should use those valueRefs to update the return value. They should call vireo.eggShell.writeJavaScriptRefNum and vireo.eggShell.readJavaScriptRefNum.

If anything we should throw an error if a return value is provided (not undefined) and say they should use the valueRef and the API functions to update the return value.

source/io/module_javaScriptInvoke.js
```diff
@@ -484,24 +440,19 @@
             errorTypeRef,
             errorDataRef) {
             var errorValueRef = Module.eggShell.createValueRef(errorTypeRef, errorDataRef);
+            var functionName = Module.eggShell.dataReadString(functionNamePointer);
+            var parameters = new Array(0);
```
rajsite: prefer `var parameters = [];`

source/io/module_javaScriptInvoke.js
```diff
@@ -484,24 +440,19 @@
...
 
-            var newError = {
-                status: false,
-                code: ERRORS.NO_ERROR.CODE,
-                source: ERRORS.NO_ERROR.MESSAGE
-            };
+            var returnValueRef = createValueRefFromPointerArray(returnPointer, 0);
+            var shouldReturnValueBeUpdated = shouldUpdateReturnValue(isInternalFunction, returnValueRef.typeRef);
+            if (!shouldReturnValueBeUpdated) {
+                parameters = parameters.concat(returnValueRef);
```
rajsite: This will vary the JS function signature (number of parameters) depending on if the return value is wired in which is a bit strange.

Instead, if the function is internal we should always prepend the returnValueRef. And if the returnValueRef is unwired then we should prepend `undefined`. That would be direct createValueRef returned undefined for this case as it did previously.

source/io/module_eggShell.js
rajsite: I think I found the situation below, and my comments down there make me think we should not loosen the check here.

source/io/module_javaScriptInvoke.js
rajsite: And like Carlos mentioned we can get rid of JavaScriptInvoke_GetParameterPointer and we can probably also get rid of JavaScriptInvoke_GetArrayElementType

source/io/module_javaScriptInvoke.js
```diff
@@ -139,13 +139,13 @@
             return jsRefNumCookieCounter;
         };
 
-        var dataReadJavaScriptRefNum = function (javaScriptRefNumPointer) {
-            var cookie = Data_ReadJavaScriptRefNum(javaScriptRefNumPointer);
+        Module.javaScriptInvoke.dataReadJavaScriptRefNum = function (javaScriptValueRef) {
```
gleonoliva: It was my understanding that functions prefixed with data are functions that directly interact with data pointers instead of valueRef's. By the way, we marked all functions in `Module.eggShell` that are prefixed with data as deprecated. Maybe we should remove the data prefix from these two?

rajsite: Agreed that we should remove the data prefix, eggShell should just be an alias to this function.

rajsite: I think we could remove the Data_ReadJavaScriptRefNum and Data_WriteJavaScriptRefNum calls as well. 

What they do is turn a labview type (JSRefNum) into a JS numeric representation and back. We already have a function that turns things into numeric representations and back, the readDouble and writeDouble functions. Those should instead be updated to support this type in addition to the other types they support that can be represented as JS numerics.

I don't think I'll push for that change here, but hopefully overtime we can use this philosophy to shrink the API surface area. It's definitely gotten too verbose.

source/io/module_javaScriptInvoke.js
rajsite: Can we remove the unnecessary not check? ie both cases are defined so no need for inversion

rajsite: I hope eventually the concept for internal functions only means that we look at a private list of functions and does not involve a change in behavior like it does now.

Eventually each parameter should be individually configurable as being passed as a valueRef and the user should have access to all the advanced jsapi functions.

source/io/module_javaScriptInvoke.js
rajsite: We should talk about this more but I think it is equally valid for returnValueRef to just be undefined. Having access to the type does not actually add value here.

source/io/module_javaScriptInvoke.js
sanmut: Good point. Will do.

source/io/module_javaScriptInvoke.js
sanmut: ah, I see. good to know. Done.

source/io/module_javaScriptInvoke.js
sanmut: Will do.

source/io/module_eggShell.js
gleonoliva: I think I disagree here. If we keep the behavior in `createValueRef` to return `undefined`. We can rewrite those two other places like:
```
var updateReturnValue = function (functionName, returnValueRef, returnUserValue, errorValueRef) {
if (returnValueRef === undefined) {
//............
}
```

and

```
Module.javaScriptInvoke.jsJavaScriptInvoke = function (
...............
var shouldReturnValueBeUpdated = shouldUpdateReturnValue(isInternalFunction, returnValueRef);
}
// Where ...

var shouldUpdateReturnValue = function (isInternalFunction, returnValueRef) {
if (returnValueRef === undefined) return false;
...
```
So far we've been guarding us against those cases where one of them is valid and the other is not. 
I don't see a strong reason to loosen that check in `createValueRef`. Am I missing something?

source/io/module_javaScriptInvoke.js
```diff
@@ -79,28 +81,26 @@
...
-            return typeName;
-        };
-
-        var getArrayElementTypeString = function (arrayPointer) {
-            var typeNamePointer = JavaScriptInvoke_GetArrayElementType(arrayPointer);
-            var responseLength = Module.coreHelpers.findCStringLength(Module.HEAPU8, typeNamePointer);
-            var typeName = Module.coreHelpers.sizedUtf8ArrayToJSString(Module.HEAPU8, typeNamePointer, responseLength);
-            return typeName;
+        var mergeNewError = function (errorValueRef, functionName, errorToSet, exception) {
+            var newError = {
```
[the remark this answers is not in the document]
sanmut: Good point. Will do.

source/core/module_typeHelpers.js
sanmut: Will document. Basically, JavascriptRefnum should be first because it should not be treated as Intger.

source/io/module_eggShell.js
sanmut: Good idea. Will do.

test-it/karma/javascriptinvoke/InternalFunction.Test.js
```diff
@@ -45,14 +47,16 @@ describe('A JavaScript function invoke', function () {
         });
     });
 
-    it('internal function successfully sets error and return value is unset', function (done) {
+    it('internal function successfully sets error and return value is unset', function (done) { // TODO
```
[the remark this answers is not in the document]
sanmut: My bad. Done, thanks.

source/io/module_javaScriptInvoke.js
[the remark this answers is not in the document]
sanmut: Removed TODO and comments.

source/io/JavaScriptInvoke.cpp
sanmut: Good catch. Will do.

source/io/module_javaScriptInvoke.js
```diff
@@ -155,158 +155,144 @@
...
+                    mergeNewError(data.errorValueRef, data.functionName, ERRORS.kNITypeMismatchForReturnTypeInJavaScriptInvoke);
+                    return;
                 }
-            },
-            JavaScriptRefNum: {
-                reader: dataReadJavaScriptRefNum,
-                writer: dataWriteJavaScriptRefNum,
-                isValidReturnType: function () {
-                    return true;
+                var dataNum = parseFloat(data.userValue);
```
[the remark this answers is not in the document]
sanmut: Agreed. Removed it.

source/io/module_javaScriptInvoke.js
sanmut: Done.

source/io/module_eggShell.js
sanmut: I cleaned up this area of code after exposing read/write Javascript in eggshell, so what I was doing was no longer required.

source/io/module_javaScriptInvoke.js
sanmut: Removed "data" prefix in the dataReadJavaScriptRefNum, dataWriteJavaScriptRefNum APIs.

source/io/module_javaScriptInvoke.js
sanmut: Good point. Done.

source/io/module_javaScriptInvoke.js
sanmut: Good to know. Done.

source/io/module_javaScriptInvoke.js
sanmut: Done.

source/io/module_javaScriptInvoke.js
sanmut: Explained in other place. This check is changed.

source/io/module_javaScriptInvoke.js
sanmut: I cleaned up this code and got rid of shouldUpdateReturnValue business.

source/io/module_javaScriptInvoke.js
```diff
@@ -484,24 +440,19 @@
             errorTypeRef,
             errorDataRef) {
             var errorValueRef = Module.eggShell.createValueRef(errorTypeRef, errorDataRef);
+            var functionName = Module.eggShell.dataReadString(functionNamePointer);
```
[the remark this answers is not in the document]
sanmut: Good catch. Done.

source/io/module_javaScriptInvoke.js
sanmut: Done.

source/io/module_javaScriptInvoke.js
sanmut: Done.

source/io/module_eggShell.js
sanmut: Oops. done.

source/io/module_javaScriptInvoke.js
sanmut: The reflect API will throw exception which will turn into an unsupported error.

source/io/JavaScriptInvoke.cpp
```diff
@@ -131,13 +110,14 @@ VIREO_FUNCTION_SIGNATUREV(JavaScriptInvoke, JavaScriptInvokeParamBlock)
...
     if (!pObserver) {
-        StringRef functionName = _Param(functionName);
+        StringRef functionName = _Param(FunctionName);
         Boolean isInternalFunction = _Param(isInternalFunction);
         const Int32 configurationParameters = 4;  // occurrence, isInternalFunction, errorCluster and functionName
         const Int32 staticTypeAndDataParameters = 2;  // Two parameters are inserted, one for type another for data. See StaticTypeAndData definition.
         Int32 userParametersCount = (_ParamVarArgCount() - configurationParameters - staticTypeAndDataParameters) / staticTypeAndDataParameters;
         StaticTypeAndData *returnValuePtr = _ParamImmediate(returnValue);
         StaticTypeAndData *parametersPtr = _ParamImmediate(parameters);
+        TypeRef typeRefString = TypeManagerScope::Current()->FindType(tsStringType);
```
rajsite: Is this still needed?

source/io/module_javaScriptInvoke.js
```diff
@@ -131,156 +138,123 @@ var assignJavaScriptInvoke;
...
+
+            var visitNumeric64 = function (valueRef) {
+                return JSON.parse(Module.eggShell.readJSON(valueRef));
+            };
+
+            return {
+                visitInt8: visitNumeric,
+                visitInt16: visitNumeric,
+                visitInt32: visitNumeric,
+                visitInt64: visitNumeric64,
```
rajsite: Only include types that were supported previously (we should add more types when supported in the editor)

source/io/module_javaScriptInvoke.js
```diff
@@ -131,156 +138,123 @@ var assignJavaScriptInvoke;
...
+
+                visitString: function (valueRef) {
+                    return Module.eggShell.readString(valueRef);
+                },
+
+                visitArray: function (valueRef) {
+                    return Module.eggShell.readTypedArray(valueRef);
+                },
+
+                visitJSObjectRefnum: function (valueRef) {
```
rajsite: Should remove visitJSObjectRefnum from this peeker and poker like discussed

source/io/module_javaScriptInvoke.js
rajsite: ie remove the 64bits, the enums, and jsObjectrefnums

source/io/module_javaScriptInvoke.js
```diff
@@ -131,156 +138,123 @@ var assignJavaScriptInvoke;
...
+                visitString: function (valueRef, data) {
+                    if (typeof data.userValue !== 'string') {
+                        mergeNewError(data.errorValueRef, data.functionName, ERRORS.kNITypeMismatchForReturnTypeInJavaScriptInvoke);
+                        return;
+                    }
+                    Module.eggShell.writeString(valueRef, data.userValue);
+                },
+
+                visitArray: function (valueRef, data) {
+                    try {
```
rajsite: Update to check type of value to return better errors

source/io/module_javaScriptInvoke.js
```diff
@@ -131,156 +138,123 @@ var assignJavaScriptInvoke;
...
+                visitEnum8: visitNumeric,
+                visitEnum16: visitNumeric,
+                visitEnum32: visitNumeric,
+
+                visitBoolean: function (valueRef, data) {
+                    if (typeof data.userValue !== 'boolean') {
+                        mergeNewError(data.errorValueRef, data.functionName, ERRORS.kNITypeMismatchForReturnTypeInJavaScriptInvoke);
+                        return;
+                    }
+                    Module.eggShell.writeDouble(valueRef, data.userValue ? 1 : 0);
```
rajsite: These writes should be wrapped in try catch to be able to report the "Unable to set return value" errors

source/io/module_javaScriptInvoke.js
```diff
@@ -452,32 +392,28 @@ var assignJavaScriptInvoke;
...
+            var parameters = [];
 
-            var newError = {
-                status: false,
-                code: ERRORS.NO_ERROR.CODE,
-                source: ERRORS.NO_ERROR.MESSAGE
-            };
+            var returnValueRef = createValueRefFromPointerArray(returnPointer, 0);
+            if (isInternalFunction) {
+                parameters = parameters.concat(returnValueRef);
```
rajsite: should switch to `parameters.push(returnValueRef)` to avoid extra array allocations like we mentioned in the book club :)

source/io/module_javaScriptInvoke.js
```diff
@@ -452,32 +392,28 @@ var assignJavaScriptInvoke;
...
+            }
 
-            var errorMessage = '';
-            var functionName = Module.eggShell.dataReadString(functionNamePointer);
-            var parameters = undefined;
             try {
-                parameters = createJavaScriptParametersArray(parametersPointer, parametersCount);
+                parameters = parameters.concat(createJavaScriptParametersArray(isInternalFunction, parametersPointer, parametersCount));
```
rajsite: can do `Array.prototype.push.apply(parameters, createJavaScriptParametersArray(isInternalFunction, parametersPointer, parametersCount))` to avoid extra array copies

test-it/karma/javascriptinvoke/ReturnDataTypes.Test.js
```diff
@@ -413,8 +415,8 @@ describe('A JavaScript function invoke', function () {
         vireo.eggShell.loadVia('enqueue(ExceptionInUpdateReturnValue)');
         runSlicesAsync(function () {
             expect(viPathParser('error.status')).toBeTrue();
-            expect(viPathParser('error.code')).toBe(44303);
-            expect(viPathParser('error.source')).toMatch(/Unable to set return value/);
+            expect(viPathParser('error.code')).toBe(44306);
```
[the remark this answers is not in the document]
rajsite: Discussed with sankara and this change should not be necessary after we add the array type checking

source/io/module_javaScriptInvoke.js
sanmut: I was contemplating about passing error object to reflect APIs and missed this one earlier. Fixing in the next commit as discussed with Milan for now to check for type compatibility for typed arrays.

source/io/module_javaScriptInvoke.js
sanmut: oops. Moved the try catch around reflectOnValueRef in pokeValueRef.

source/io/module_javaScriptInvoke.js
sanmut: Removed 64-bits, enums. We should talk about removing visitJSObjectRefnum from peeker and poker.

source/io/module_javaScriptInvoke.js
sanmut: I am marking this resolved and keeping the one on peeker open. While resolving that, we should address this as well the same way anyway.

source/io/module_javaScriptInvoke.js
```diff
@@ -407,11 +393,12 @@ var assignJavaScriptInvoke;
 
             var returnValueRef = createValueRefFromPointerArray(returnPointer, 0);
             if (isInternalFunction) {
-                parameters = parameters.concat(returnValueRef);
+                parameters.push(returnValueRef);
             }
 
             try {
-                parameters = parameters.concat(createJavaScriptParametersArray(isInternalFunction, parametersPointer, parametersCount));
+                // Array.prototype.push.apply(parameters, addToJavaScriptParametersArray (isInternalFunction, parametersPointer, parametersCount));
```
cglzaguilar: Please remove this commented line. I don't think is adding any value.

source/io/module_javaScriptInvoke.js
rajsite: Talked about removing visitJSObjectRefnum and will remove in a separate PR

test-it/karma/javascriptinvoke/InternalFunction.Test.js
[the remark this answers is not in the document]
rajsite: Might have missed that TODO

PR #518, opened 2018-10-03
Allocation check behind flag and promisify vireoRunner

rajsite:
The major change is modifying the _totalAQAllocated check in Malloc to only occur when the VIREO_TRACK_MEMORY_QUANTITY flag is set.

The reason is that in Release when the flag is unset then _totalAQAllocated appears to represent the number of allocations that have occurred, not the total number of bytes allocated. In Debug builds _totalAQAllocated represents the total number of bytes allocated.

Also modified Malloc to cause the Vireo runtime to stop execution and print an error if a malloc failure occurs (ie malloc returns zero). Could be handy for accidental very large allocations but if truly out of memory it's pointless.

Also modified TestHelpers.VireoRunner.js to be promise-based and handle when Vireo crashes (ie executeSlicesUntilClumpsFinished rejections) which was originally needed, but after adding the graceful exit on failed malloc it wasn't strictly necessary. Still keeping for a change @sanmut is making where we do want to test Vireo failure scenarios with JSLI.


source/core/TypeAndDataManager.cpp
```diff
@@ -128,16 +128,15 @@ size_t gAllocNumWatch = 0;
 void* TypeManager::Malloc(size_t countAQ)
 {
     VIREO_ASSERT(countAQ != 0);
+    size_t allocationCount = 1;
+
+#ifdef VIREO_TRACK_MEMORY_QUANTITY
```
sanmut: I don't think we should use this directive to change the meaning of how much gets allocated only in Malloc(). There are other places that use this directive.  I would recommend adjusting only _allocationLimit in this PR and not change the meaning of what VIREO_TRACK_MEMORY_QUANTITY controls. We can discuss that later.

sanmut: Discussed with Milan. I misunderstood part of the changes. However, the way TrackAllocation and _totalAQAllocated are handled is confusing and worth revisiting later to cleanup.

PR #520, opened 2018-10-11
Adding null check for JavaScriptRefNum in events configuration instructions

segaljared:
Adding null check for JavaScriptRefNum in events configuration instructions. I've also updated the JavaScriptRefNum karma test so that it works correctly.

I have not been able to get a test working for the null check specifically, because other tests start failing when the error occurs in Vireo. It's currently commented out in the test file.

### DIFF
--- a/source/core/Events.cpp
+++ b/source/core/Events.cpp
@@ -1218,7 +1218,7 @@ VIREO_FUNCTION_SIGNATURE3(ConfigureEventSpecJSRef, Int32, Int32, JavaScriptRefNu
     Int32 eventSpecIndex = _Param(1);
     JavaScriptRefNum jsRefNum = _Param(2);
     if (jsRefNum == 0) {
-        THREAD_EXEC()->LogEvent(EventLog::kHardDataError, "JavaScriptRefNum must not be null");
+        THREAD_EXEC()->LogEvent(EventLog::kSoftDataError, "JavaScriptRefNum must not be null");
         return THREAD_EXEC()->Stop();
     }
     VirtualInstrument *owningVI = THREAD_CLUMP()->OwningVI();
@@ -1231,7 +1231,7 @@ VIREO_FUNCTION_SIGNATURE2(RegisterForJSEvent, JavaScriptRefNum, UInt32)
 {
     JavaScriptRefNum jsRefNum = _Param(0);
     if (jsRefNum == 0) {
-        THREAD_EXEC()->LogEvent(EventLog::kHardDataError, "JavaScriptRefNum must not be null");
+        THREAD_EXEC()->LogEvent(EventLog::kSoftDataError, "JavaScriptRefNum must not be null");
         return THREAD_EXEC()->Stop();
     }
     UInt32 eventType = _Param(1);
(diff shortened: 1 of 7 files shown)


source/core/Events.cpp
```diff
@@ -1217,6 +1217,10 @@ VIREO_FUNCTION_SIGNATURE3(ConfigureEventSpecJSRef, Int32, Int32, JavaScriptRefNu
     Int32 eventStructIndex = _Param(0);
     Int32 eventSpecIndex = _Param(1);
     JavaScriptRefNum jsRefNum = _Param(2);
+    if (jsRefNum == 0) {
```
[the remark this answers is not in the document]
segaljared: Not in C++. However, should I be using !jsRefNum instead of ==  0?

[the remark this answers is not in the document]
spathiwa: It's fine as-is.

test-it/karma/fixtures/events/ValueChangeStaticControlEventNullJavaScriptRefNum.via
```diff
@@ -0,0 +1,55 @@
+define (UpdateBooleanOnValueChangeEvent dv(.VirtualInstrument (
+    Events: c(
+        e(c( // Event Struct 0
+           e(c(  // Event spec 0
+               e(dv(UInt32 1000) eventSource)  // event source enum
+               e(dv(UInt32 2) eventType)  // ValueChanged
+               e(dv(UInt32 18) controlUID)  // for static control refs
```
spathiwa: This should be a JavaScriptRefnum, not a controlUID.

Maybe we should actually enforce this, and check the type when you use ConfigureEventSpecjSRef.

segaljared: This won't really be related to ConfigureEventSpecJSRef, since this information isn't actually fed into ConfigureEventSpecJSRef. We need the control ID in the event spec at some point, because otherwise we won't know what the ID actually is, since JavaScriptRefNums are opaque to Vireo.

test-it/karma/events/ControlEvents.Test.js
```diff
@@ -244,9 +246,29 @@ describe('The Vireo Control Event', function () {
...
             expect(rawPrintError).toBeEmptyString();
-            expect(viPathParser('eventOccurred')).toEqual(true);
-            expect(viPathParser('eventTimedOut')).toEqual(false);
+            expect(rawPrint).toBe('Waiting on Events\nEvent occurred\n');
+            done();
+        });
+    });
+
+    /*
+    it('errors when JavaScriptRefNum is left null', function (done) {
```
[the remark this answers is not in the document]
rajsite: Removing the setTimeout and enabling the test is even better

test-it/karma/events/ControlEvents.Test.js
```diff
@@ -244,8 +246,17 @@ describe('The Vireo Control Event', function () {
 
         runSlicesAsync(function (rawPrint, rawPrintError) {
             expect(rawPrintError).toBeEmptyString();
-            expect(viPathParser('eventOccurred')).toEqual(true);
-            expect(viPathParser('eventTimedOut')).toEqual(false);
+            expect(rawPrint).toBe('Waiting on Events\nEvent occurred\n');
```
rajsite: Instead you can do:
```
var result = `Waiting on Events
Event occurred
`;
 expect(rawPrint).toMatchVtrText(result)
```

It cleans up line endings, removes comments, and prints a diff of the changes on failures. Used by the karma vtr runner.

rajsite: Actually, only really valuable for lots of text, so nvm

PR #522, opened 2018-10-11
JavaScript invoke internal function clean-up

rajsite:
The goals of this PR are three things:

1. Make sure internal functions do not rely on returning values. They should instead be using api functions to set the returnValueRef. If values are returned for internal functions we error instead.
2. Enable tests that check if the vireo runtime has an error during execution
3. Remove JSObjectRefnum from the public parameterValueVisitor (formerly peeker) and returnValueVisitor (formerly poker)

In the process I made the following changes:
1. Cleaned up the sync vs async code path so they call three helper functions for processing error and returnValues. This makes it much clearer what are the actual differences in the code paths
2. Make terminology used consistent throughout
3. Increase coverage of locations where internal errors should be reported as exceptions instead of LabVIEW errors. The heuristic is that each location with mergeNewError calls should be preceded with an internal check and throw or justify why one is not needed.

### DIFF
--- a/source/io/module_javaScriptInvoke.js
+++ b/source/io/module_javaScriptInvoke.js
@@ -187,18 +187,14 @@ var assignJavaScriptInvoke;
                     try {
                         fn(valueRef, data);
                     } catch (ex) {
-                        if (data.isInternalFunction) {
-                            throw ex;
-                        }
+                        // Internal check not needed because poker is not used for internal calls
                         mergeNewError(data.errorValueRef, data.functionName, ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke, ex);
                     }
                 };
             };
 
             var reportTypeMismatch = function (data) {
-                if (data.isInternalFunction) {
-                    throw new Error('Type mismatch trying to set return value');
-                }
+                // Internal check not needed because poker is not used for internal calls
                 mergeNewError(data.errorValueRef, data.functionName, ERRORS.kNITypeMismatchForReturnTypeInJavaScriptInvoke);
             };
 
@@ -334,8 +330,7 @@ var assignJavaScriptInvoke;
                 data = {
                     returnValue: returnValue,
                     errorValueRef: errorValueRef,
-                    functionName: functionName,
-                    isInternalFunction: isInternalFunction
+                    functionName: functionName
                 };
                 Module.eggShell.reflectOnValueRef(jsInvokePoker, returnValueRef, data);
             }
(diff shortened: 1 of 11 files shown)


source/io/module_javaScriptInvoke.js
```diff
@@ -423,35 +435,43 @@ var assignJavaScriptInvoke;
...
+                Module.eggShell.setOccurrence(occurrencePointer);
+                return;
+            }
+
+            // for async execution check that user did not provide a return value
+            if (returnValue !== undefined) {
+                if (isInternalFunction) {
+                    throw new Error('Unexpected return value for function requiring asynchronous completion');
                 }
+                mergeNewError(errorValueRef, functionName, ERRORS.kNIUnableToSetReturnValueInJavaScriptInvoke);
```
rajsite: This error happens if a user retrieves the getCompletionCallback() but also gives a return value. We error because the user should not give a return value after getCompletionCallback is called, they should be passing the return value to the getCompletionCallback.

Previously this was silently ignored but now we give the kNIUnableToSetReturnValueInJavaScriptInvoke error, aka:
"Unable to set return value for JavaScript Library Interface node parameter". Should we make a new error code for this scenario or is the unable to set return value error clear enough?

cglzaguilar: I would prefer a new error code. That way we can give precise guidance as to how to fix it or we can use the current one, but we need to update the documentation.

rajsite: That's fair, I'll make that in a separate change.

source/io/module_javaScriptInvoke.js
```diff
@@ -303,21 +290,47 @@ var assignJavaScriptInvoke;
...
 
-        var updateReturnValue = function (functionName, returnValueRef, returnUserValue, errorValueRef) {
-            if (returnValueRef === undefined) {
-                // User doesn't want to return value. We get here, if we're passing '*' for return parameter in VIA code
+        var hasExecutionError = function (returnValue) {
+            return returnValue instanceof Error;
+        };
+
+        var reportExecutionError = function (functionName, returnValue, errorValueRef, completionCallbackStatus, isInternalFunction) {
+            if (!hasExecutionError(returnValue)) {
```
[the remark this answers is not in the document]
rajsite: Yea it is redundant. I actually removed it once because I realized I couldn't cover it in tests, but then added it back to be defensive to changes in the future. I don't feel strongly either way and can remove it if that's preferred, but otherwise leaving it.

PR #523, opened 2018-10-16
Removal of legacy vireo api and some old prototypes

rajsite:
With the recent updates to the JavaScriptInvoke effectively all of the remaining usages of the deprecated api functions can be removed.

This PR does the following:
1. Migrates a few tiny remaining pieces of the JavaScriptInvoke to the new api functions (primarily the IsNotAJavaScriptRefnum instruction)
2. Removes all the deprecated api functions possible (verifying not used in ASW)
3. Enables the NO_FILESYSTEM=1 build flag so the simulated FS infrastructure is not shipped. This is the bulk of the savings.

    It does result in the mandelbrot tests failing in JS because they use stream apis instead of print / printf apis. This is fixed in https://github.com/kripken/emscripten/pull/7255. So instead of disabling the instructions in JS builds we can wait for a release to re-enable those tests. None of that should effect ASW as we don't have print / stream instructions emitted there yet.
4. Adds a transform to the karma test runner so that null and newline characters are both transformed to newline (mitigation to behavior change after switching to NO_FILESYSTEM=1). This does mean we cannot assert in tests accurate output of null padding in flatten functions. As flatten and unflatten are not shipped in ASW that is not a big deal yet. Described use case in the following GitHub issue: https://github.com/kripken/emscripten/issues/5290
5. Fixes the karma VTR runner so that the queue and event tests actually run in the browser. Apparently they never were running before. Had to disable one of the core event tests in browser as it fails in browser but not node.js or native. The test was moved to just run in native and node.js for now.
6. Remove the Canvas prototype and the BrowserFPS apis. All of the tests for them were disabled  / unmaintained. If we were to seriously implement we would use SDL (crossplatform) / JavaScriptInvoke for Canvas these days and JavaScriptInvoke for BrowserFPS. So got rid of the unmaintained code.

The following shows the size saving of the PR:
Before:
vireo.core.wasm: 481 KB
vireo.min.js: 213 KB

After:
vireo.core.wasm: 476 KB
vireo.min.js: 123 KB


karma.shared.js
```diff
@@ -109,7 +109,14 @@
             logLevel: config.LOG_INFO,
 
             // Timeout for jasmine set to 50000 in TestHelpers.ExtendJasmineTimeout.js so warn if test time getting close
-            reportSlowerThan: 40000
+            reportSlowerThan: 40000,
+
+            // Settings to pass to frameworks and plugins
+            client: {
+                jasmine: {
+                    random: true
```
rajsite: This is the default, but just exposing in the config so it is easier to disable when debugging. It's actually a pretty good default, helps shake out bad infrastructure problems.

source/core/JavaScriptRef.cpp
```diff
@@ -20,18 +20,18 @@ namespace Vireo {
...
 
 VIREO_FUNCTION_SIGNATURE2(IsNotAJavaScriptRefnum, JavaScriptRefNum, Boolean)
 {
+    TypeRef typeRefJavaScriptRefNum = TypeManagerScope::Current()->FindType("JavaScriptRefNum");
+    TypeRef typeRefIsNotARefNum = TypeManagerScope::Current()->FindType("Boolean");
     JavaScriptRefNum* refnumPtr = _ParamPointer(0);
     Boolean isNotARefnum = false;
     #if kVireoOS_emscripten
-        jsIsNotAJavaScriptRefnum(&isNotARefnum, refnumPtr);
+        jsIsNotAJavaScriptRefnum(typeRefJavaScriptRefNum, refnumPtr, typeRefIsNotARefNum, &isNotARefnum);
```
rajsite: These parameters were flipped from the instruction signature so made them consistent

sanmut:
> Had to disable one of the core event tests in browser as it fails in browser but not node.js or native. The test was moved to just run in native and node.js for now.

Are you tracking these failures somewhere?

sanmut:
> As flatten and unflatten are not shipped in ASW that is not a big deal yet. Described use case in the following GitHub issue: kripken/emscripten#5290

I am glad you mentioned this in the issue. 

> This does mean we cannot assert in tests accurate output of null padding in flatten functions. 

What do you mean by "flatten" functions and that we don't support in ASW yet?

source/io/module_javaScriptInvoke.js
```diff
@@ -117,12 +97,12 @@ var assignJavaScriptInvoke;
         };
 
         Module.javaScriptInvoke.readJavaScriptRefNum = function (javaScriptValueRef) {
-            var cookie = Data_ReadJavaScriptRefNum(javaScriptValueRef.dataRef);
+            var cookie = Module.eggShell.readDouble(javaScriptValueRef);
```
sanmut: Using readDouble to deal with refnum does not feel right at the API level. It is probably an implementation detail that readDouble works for refnum. OTOH, the API may be generalized for "Refnum" and not specific to "JavascriptRefNum" if that makes sense.

rajsite: Nope, disagree. The major idea of the new api is to focus on reading values out of Vireo's memory as the host memory type. Previously we had hundreds of functions for converting each vireo type to a type supported in JS. The new api flips it on it's head. There are a handful of functions that can be used to can handle lots of vireo types.

The previous function read a JavaScriptRefNum type and resulted in a JavaScript number as a double. That IS what it did just with tons of implicit conversions. If reading as a double felt wrong, too late that is what we were doing, it just wasn't explicit.

This makes it very explicit.

Now the ReadDoubleFromMemory function implements very specifically how the Refnum is represented as a double: https://github.com/ni/VireoSDK/pull/523/files#diff-f2734a9037270e6cd627741ed25d331aR2724

So instead of hiding behind implicit conversions from Int32 to double, this codifies it explicitly.

source/core/JavaScriptRef.cpp
```diff
@@ -20,18 +20,18 @@ namespace Vireo {
...
-    // Parameters: returnValue*, reference
-    extern void jsIsNotAJavaScriptRefnum(Boolean*, JavaScriptRefNum*);
+    extern void jsIsNotAJavaScriptRefnum(TypeRef, JavaScriptRefNum*, TypeRef, Boolean*);
 }
 #endif
 
 VIREO_FUNCTION_SIGNATURE2(IsNotAJavaScriptRefnum, JavaScriptRefNum, Boolean)
 {
+    TypeRef typeRefJavaScriptRefNum = TypeManagerScope::Current()->FindType("JavaScriptRefNum");
+    TypeRef typeRefIsNotARefNum = TypeManagerScope::Current()->FindType("Boolean");
```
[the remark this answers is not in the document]
rajsite: done

rajsite:
@sanmut 

>> Had to disable one of the core event tests in browser as it fails in browser but not node.js or native. The test was moved to just run in native and node.js for now.
>
> Are you tracking these failures somewhere?

Tracking here: https://github.com/ni/VireoSDK/issues/524

>> This does mean we cannot assert in tests accurate output of null padding in flatten functions.
> 
> What do you mean by "flatten" functions and that we don't support in ASW yet?

The FlattenToString and UnflattenFromString nodes are not exposed on pallets

sanmut:
> > > This does mean we cannot assert in tests accurate output of null padding in flatten functions.
> > What do you mean by "flatten" functions and that we don't support in ASW yet?
> The FlattenToString and UnflattenFromString nodes are not exposed on pallets
So, when we need to support, we can't write VTR-type test for null-string cases, right (until emscripten fixes the issue and we upgrade)? That's an acceptable limitation, but just making sure.

make-it/EmMakefile
```diff
@@ -93,11 +93,10 @@ ifeq ($(TARGET), wasm32-unknown-emscripten)
 endif
 
 # NO_DYNAMIC_EXECUTION=1 Allows Vireo to be used in CSP environments where eval() is forbidden: https://github.com/kripken/emscripten/blob/master/src/settings.js#L563
-EM_OPT = $(EM_OPTFLAGS) $(TARGET_FLAG) -s NO_EXIT_RUNTIME=1 -std=c++14 -fno-exceptions --memory-init-file 0 -s NO_DYNAMIC_EXECUTION=1 -s MODULARIZE=1 -s EXPORT_NAME=VireoCreateCoreModule -s EXPORT_ES6=1 --minify 0
```
sanmut: With "ABORTING_MALLOC=0", would we get a different exception later? Why do we do this?

rajsite: Potentially gives Vireo a chance to clean-up since it has those code paths. I think it would be difficult to test out of memory situations that will always successfully clean-up beyond the path we already tested.

source/io/module_javaScriptInvoke.js
rajsite: This discussion did make me realize that if the definition of refnum changed from UInt32 to UInt64 then this code should fail as a UInt64 RefNum cannot be safely represented as a double (other UInts will loose precision which is a choice the user can make but that is not a valid choice for RefNums).

So I made the following change in this PR:
https://github.com/ni/VireoSDK/pull/523/files#diff-f2734a9037270e6cd627741ed25d331aR2722

https://github.com/ni/VireoSDK/pull/523/files#diff-f2734a9037270e6cd627741ed25d331aR2792

source/core/Events.cpp
```diff
@@ -596,8 +596,10 @@ void RegisterForStaticEvents(VirtualInstrument *vi) {
                     STACK_VAR(String, viNameVar);
                     StringRef viName = viNameVar.Value;
                     GetVIName(vi, viName);
-                    char *tagBegin = reinterpret_cast<char*>(tag->Begin()), *tagEnd = reinterpret_cast<char*>(tag->End());
-                    EventControlUID controlID = Int32(strtol(tagBegin, &tagEnd, 10));
+                    SubString tagSubString = tag->MakeSubStringAlias();
+                    TempStackCString tagCString;
+                    tagCString.Append(&tagSubString);
```
sanmut: Good catch :+1: 
I think this might be sufficient:
TempStackCString tagCString(tag->Begin(), tag->Length());

rajsite: done

source/core/Events.cpp
```diff
@@ -596,9 +596,7 @@ void RegisterForStaticEvents(VirtualInstrument *vi) {
                     STACK_VAR(String, viNameVar);
                     StringRef viName = viNameVar.Value;
                     GetVIName(vi, viName);
-                    SubString tagSubString = tag->MakeSubStringAlias();
-                    TempStackCString tagCString;
-                    tagCString.Append(&tagSubString);
+                    TempStackCString tagCString(tag->Begin(), tag->Length());
```
spathiwa: Looks good.

PR #530, opened 2018-10-22
Support Promises for JavaScriptInvoke calls

rajsite:
This PR allows users to return a Promise to perform async behaviors instead of relying on jsapi.getCompletionCallback.

The implementation itself is pretty minor, the bulk of the work was refactoring the Async tests. The tests were refactored so that the same test used for callbacks can be used for promises. This makes sure that we maintain equal coverage of both over time and evaluate error scenarios common to both overtime.

The lines of test code have gone up but the complexity should be greatly reduced. Each test is now relatively standalone and it is clearer to see how assertions in the test match the executing code.

(reopened to trigger builds which failed previously due to GitHub outage)


rajsite:
FYI all webhooks disabled due to github service outage so builds not queuing: https://status.github.com/messages

source/io/module_javaScriptInvoke.js
```diff
@@ -30,6 +30,18 @@ var assignJavaScriptInvoke;
...
+        },
+
+        // TODO needs review and to be added to error code database
+        kNIUnableToHandlePromise: {
+            CODE: 44307,
+            MESSAGE: 'Unable to use Promise. Verify that the getCompletionCallback API function is not used in the external JavaScript function with a Promise return value.'
+        },
+
+        // TODO needs review and to be added to error code database
+        kUnableToAcceptReturnValueDuringAsync: {
```
siddhukrs: Looking at the other errors, looks like there should be an "NI" here.

test-it/karma/javascriptinvoke/Async.Test.js
```diff
@@ -8,277 +8,797 @@ describe('A JavaScript function invoke', function () {
...
+            beforeEach(function () {
+                window.SingleFunction = function (input, jsapi) {
+                    var completionCallback = jsapi.getCompletionCallback();
+                    try {
+                        jsapi.getCompletionCallback();
+                    } catch (ex) {
+                        error = ex;
+                    }
+                    // TODO mraj this allows completing after an exception is thrown, this should not be allowed.
+                    // Instead if we are not resolved we should resolve immediately with an error
```
siddhukrs: Good catch. Can we track this somehow? Maybe as a CAR, because we do not have a US for this work.

siddhukrs:
Looks good! I looked through all the test cases and they do cover everything we previously had (that are still relevant). Sorry that it long.

test-it/karma/javascriptinvoke/Async.Test.js
rajsite: Made CAR 717753 for it titled "JSLI api functions that error asynchronously cause the block diagram to hang"

rajsite:
@siddhukrs Can you look through the test clean-up commits for the changes we discussed

PR #531, opened 2018-10-22
Update devDependencies to remove vulnerability

rajsite:
Update devDependencies with `npm audit fix` so no longer depend on a package with a vulnerability. https://nvd.nist.gov/vuln/detail/CVE-2018-3721

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "7.0.0"
+        "@babel/highlight": "^7.0.0"
       }
     },
     "@babel/highlight": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -19,9 +19,9 @@
       "integrity": "sha512-UFMC4ZeFC48Tpvj7C8UgLvtkaUuovQX+5xNWrsIoMG8o2z+XFKjKaN9iVmS84dPwVN00W4wPmqvYoZF3EGAsfw==",
       "dev": true,
       "requires": {
-        "chalk": "2.4.1",
-        "esutils": "2.0.2",
-        "js-tokens": "4.0.0"
+        "chalk": "^2.0.0",
```
cglzaguilar: I find odd that we are getting older versions instead of recent ones. Is this on purpose?

rajsite: It's not an older version, the lock file format in newer versions of npm just now changed from showing exact versions in this section to showing the version range.

2.4.1 is compatible with ^2.0.0

rajsite: After merged though we probably should have everyone upgrade the npm version

cglzaguilar: All right! Thanks a lot for the clarification.

sanmut:
Can this be merged?

rajsite:
mmm I'm actually going to hold off for a bit. With this change I was going to recommend everyone update to the latest npm but that will put us out of sync with mainline. I want to make sure that won't cause any issues, gonna close this for now.

PR #533, opened 2018-11-01
Fix parsing for VIA files with lots of events (too many arguments in an instruction)

segaljared:
On suggestion from @spathiwa, I've added a helper that will switch to a dynamic array whenever there are more than the "max" number of arguments to an instruction. In this case, the abstraction holds onto a statically defined array and will push any overflow elements into a vector, resizing the vector as needed.


segaljared:
Just to be clear: the issue that is being fixed here has to do with VIA files with so many events that the argument list for the WaitForEventsAndDispatch instruction exceeds the current limit. To fix this, I'm allowing for elements over that limit to be put into a vector. This preserves the performance of using a static array for most instructions, while allowing for extra long argument lists for others

source/core/TDCodecVia.cpp
```diff
@@ -1638,7 +1638,8 @@ void TDViaParser::ParseClump(VIClump* viClump, InstructionAllocator* cia)
     ClumpParseState state(viClump, cia, _pLog);
     SubString  token;
     SubString  instructionNameToken;
-    SubString  argExpressionTokens[ClumpParseState::kMaxArguments];
+    SubString  argExpressionTokenBacker[ClumpParseState::kMaxArguments];
+    ExpandableStaticArray<SubString> argExpressionTokens(argExpressionTokenBacker, ClumpParseState::kMaxArguments);
```
spathiwa: I think it would be much cleaner for the ExpandableStaticArray to own the 'backer' static array as well. There's no advantage to having the caller pass it, and you have to duplicate the size and make sure it's consistent.

source/core/TDCodecVia.cpp
```diff
@@ -1790,6 +1786,7 @@ void TDViaParser::ParseClump(VIClump* viClump, InstructionAllocator* cia)
             if (!instruction) {
                 LOG_EVENTV(kSoftDataError, "Instruction not generated '%.*s'", FMT_LEN_BEGIN(&instructionNameToken));
             }
+            argExpressionTokens.Clear();
```
spathiwa: You should have a destructor that does this, and rely on that rather than doing Clear explicitly.
Otherwise, incorrect usage could leak.

source/include/TDCodecVia.h
```diff
@@ -207,6 +207,31 @@ class TDViaFormatter
...
+    ExpandableStaticArray(T* array, Int32 length) {
+        _staticArray = array;
+        _staticLength = length;
+    }
+    T& operator[] (Int32 index) {
+        if (index < _staticLength)
+            return _staticArray[index];
+        index -= _staticLength;
+        if (index >= _dynamicArray.size())
+            _dynamicArray.resize(index + 1);
```
spathiwa: Consider having a chunk size by which you grow, and call _dynamicArray.reserve(index + kChunkSize) so the array doesn't have to resize and realloc memory on every single addition past the static size.

spathiwa: Also, consider making the API mimic like std::vector.

segaljared: Do you mean by including methods like resize and begin and end? Is the main benefit of wrapping this in our own class over just using std::vector that we can pre-allocate the "usual" amount needed in a stack allocated array?

source/core/TDCodecVia.cpp
segaljared: I will remove this because I don't think it's necessary and will result in more allocations instead of less. The idea wasn't to free memory before destruction, but to free it before we process the next instruction. The object doesn't get destructed in between instructions.

As far as leaks go, shouldn't the std::vector be deallocated automatically since it is a non-pointer member variable of the class?

spathiwa: Yeah, that's true.  Sorry, ignore this comment.

source/include/TDCodecVia.h
spathiwa: It's not so much the names of the API entries that I'm concerned about, it's that the functionality is self-contained and it seems like a usable data structure on its own.   Holding onto a static array but not knowing how many elements are actually used seems half-done.   Making it use something like push_back rather than auto-inflating when indexing the element would be better.  And since it is a generic data structure and not something very Vireo specific, adopting the STL convention rather than using names that makes it look like it could be a Vireo TypedArray seems reasonable to me, but I don't feel strongly about the names.  If you want to keep Clear() and Length(), and add AddElement() (or something) instead of push_back(), that's fine with me.

segaljared: Sounds good. I've updated the helper class to take a size_t template argument and added the push_back. I also changed the []operator to no longer allocate more space, since push_back should be used in that kind of scenario.

source/core/TDCodecVia.cpp
segaljared: Done

source/core/TDCodecVia.cpp
```diff
@@ -1638,7 +1638,7 @@ void TDViaParser::ParseClump(VIClump* viClump, InstructionAllocator* cia)
     ClumpParseState state(viClump, cia, _pLog);
     SubString  token;
     SubString  instructionNameToken;
-    SubString  argExpressionTokens[ClumpParseState::kMaxArguments];
+    ExpandableStaticArray<SubString, ClumpParseState::kMaxArguments> argExpressionTokens;
```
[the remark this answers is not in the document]
segaljared: @spathiwa What do you think of using std::vector here instead of the new helper class?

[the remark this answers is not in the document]
spathiwa: My reason for suggesting creating the class was to avoid the extra dynamic allocation in the common case where there are a limited number of arguments.  The idea of a class which uses static allocations for small/common situations but automatically expands to use dynamic allocation when necessary is a common design pattern and can be very beneficial.
However, in this case, where the same array is being reused every time to parse the arguments of every instruction, I concede that adding a reserve() call would get most of the benefits of the class.  However, since Jared did the work of making the class on my recommendation, and it's a judgement call whether it's a win in this case but could be useful in other areas, I say he should keep it.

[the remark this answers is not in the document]
cglzaguilar: I concur with @sanmut, please use a std::vector instead and reserve ClumpParseState:kMaxArguments for it. That will make it a smaller fix, which is great because we want to patch this if there is another patch for NXG 3.1.

PR #535, opened 2018-11-05
Fixing JS reference being written in a loop

ggumen:
Fixing the case of control reference wire being written in a loop. We should allow this case. Originally we thought we couldn't get to the scenario of changing the value of a JS reference, but we can easily get into this case if the wire was being re-written to a new value in a loop with Reference PropertyRead (ie, label reference).


rajsite:
@ggumen would you be up for renaming these maps and helper functions? https://github.com/ni/VireoSDK/pull/535/commits/51be83d669153c46f284baeabc1137698f1f9057#diff-a7468c287af215525c8ba9f958336bd4R75

All of them say js refnum <-> js value but should really be cookie <-> js value. MAkes the story of jsRefNums being rewritable cookie containers much clearer.

ggumen:
> 
> 
> @ggumen would you be up for renaming these maps and helper functions? [51be83d#diff-a7468c287af215525c8ba9f958336bd4R75](https://github.com/ni/VireoSDK/commit/51be83d669153c46f284baeabc1137698f1f9057#diff-a7468c287af215525c8ba9f958336bd4R75)
> 
> All of them say js refnum <-> js value but should really be cookie <-> js value. Makes the story of jsRefNums being rewritable cookie containers much clearer.

Done.

PR #537, opened 2018-11-06
Fix issues found in native test suite with AddressSanitizer

tchipley:
There are three fixes here:
1. In two places in `Array.cpp`, don't attempt to copy from before the start of the array to avoid a buffer underflow.
2. In `TDViaParser::ParseArray`, only copy at most `kArrayMaxRank` items from `dimensionLengths`, as we never put more than `kArrayMaxRank` items into it. This avoids a buffer overflow when we try to `memcpy` from it in `ArrayType::ArrayType`.
3. In `TypedScanString`, cap the length of `in` used for `AliasAssign` at `in.Length()`. This avoids a buffer overflow later when copying it in `FixedCArray<unsigned char, 255>::Append`

In addition to the above fixes, add a flag in Makefile to make enabling AddressSanitizer simple.

### DIFF
--- a/make-it/Makefile
+++ b/make-it/Makefile
@@ -96,6 +96,12 @@ else
         CC=clang++
         LDFLAGS+= -dead_strip
         CFLAGS+= -pthread $(CCOPT) $(TARGETARCH) -fno-rtti -fno-exceptions -std=c++11
+
+        ifdef ASAN
+            LDFLAGS += -fsanitize=address
+            CFLAGS += -fsanitize=address
+        endif
+
         PROGRAM_DIR=/Applications/Vireo
     endif
 endif
--- a/source/core/Array.cpp
+++ b/source/core/Array.cpp
@@ -139,7 +139,7 @@ VIREO_FUNCTION_SIGNATUREV(ArrayIndexND, ArrayIndexNDParamBlock)
             empty = true;
     }
     for (; j < rank; ++j) {
-        arraySlabLen[j] = arraySlabLen[j-1];
+        arraySlabLen[j] = j >= 1 ? arraySlabLen[j-1] : 0;
         arrayLen[j] = 1;
     }
     if (subRank > 0) {
@@ -570,8 +570,9 @@ VIREO_FUNCTION_SIGNATUREV(ArrayReplaceSubsetND, ArrayReplaceSubsetParamBlock)
         if (len == 0)
             empty = true;
     }
-    for (; j < rank; ++j)
-        arrayOutSlabLengths[j] = arrayOutSlabLengths[(j-1)];
+    for (; j < rank; ++j) {
+        arrayOutSlabLengths[j] = j >= 1 ? arrayOutSlabLengths[(j-1)] : 0;
+    }
     void* element = arguments[i]._pData;
     IntIndex srcRank =  arguments[i]._paramType->Rank();
     if (srcRank != expectedElemRank) {
--- a/source/core/NumericString.cpp
+++ b/source/core/NumericString.cpp
@@ -1473,7 +1473,7 @@ Boolean TypedScanString(SubString* inputString, IntIndex* endToken, const Format
                 break;
             }
         }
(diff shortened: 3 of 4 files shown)


make-it/Makefile
```diff
@@ -96,6 +96,12 @@ else
         CC=clang++
         LDFLAGS+= -dead_strip
         CFLAGS+= -pthread $(CCOPT) $(TARGETARCH) -fno-rtti -fno-exceptions -std=c++11
+
+        ifdef ASAN
```
sanmut: It looks like gcc 4.8 supports address sanitizer. Did you try address sanitizer flags with gcc? If that works, I think we should try and leave the flags on for Travis CI builds and see how that goes? If there are performance (build/test run time), then we should leave it on for tag(release) builds.

sanmut: I am going to wait on approving this until we find out if we can enable it for gcc on Travis.

tchipley: I know the flags work with GCC, but I haven't tried it with this project. I should have some time today to give it a shot. It does check for leaks on Linux by default, so I may need to add fixes for those if I add the flags there, FYI.

ASan has significant performance impact: `make testnative` takes about 4x as long (from ~25 seconds to ~100 seconds on my machine). I leave it to you guys to decide if the performance impact is worth it to run with it constantly (outside of released bits).

rajsite: Could probably test it pretty quick by changing:

https://github.com/ni/VireoSDK/blob/da7c8c23213ff5b7590f0a206e9f7a754b4621db/.travis.yml#L47

To `make native ASAN=1`

We probably wouldn't turn it on all the time locally but on the CI (ubuntu trusty gcc4.9) seems fine.

You can try it in a brach, just make sure you have Travis CI enabled in your fork: https://github.com/ni/VireoSDK/blob/master/CONTRIBUTING.md#connect-fork-to-ci-servers

tchipley: Thanks for the suggestion @rajsite! I was able to get a passing Travis CI with ASan enabled: https://travis-ci.org/tchipley/VireoSDK/builds/452022689, so if y'all want it enabled there I can turn it on. I had to change `make unittest` to `make unittest ASAN=1` as well as changing `make native`.

> We probably wouldn't turn it on all the time locally but on Travis CI (ubuntu trusty gcc4.9) seems fine.

> Then, it is not worth it turning it on in Travis CI builds by default. For every commit someone pushes on their PR, we will end up incurring the penalty. Turning it on for tagged(Release) builds is good enough, as bugs of this nature should be rare enough.

@rajsite and @sanmut: I'm getting mixed messages about enabling ASan in Travis CI. I want to make sure everybody knows that ASan is not intended to be enabled on released binaries, which it looks like the Travis CI generates.

rajsite: We can enable it if appropriate in a separate PR. No need to do so in this PR.

You are right that the linux build we make is released. However, there are not any known consumers of that build. It is true that if we publish a linux build we should not have that build be published with ASAN enabled. Maybe we don't really need to publish the Linux build. We can have those discussions in a separate PR.

Could you add a bit of doc to the README for Linux that describes using the ASAN flag https://github.com/ni/VireoSDK/blob/master/README.md#vireo-linux-native-target

After that looks good to me!

rajsite:
@tnelligan oh! one more thing, can you add yourself to Authors :)
https://github.com/ni/VireoSDK/blob/master/AUTHORS

make-it/Makefile
tchipley: Cool. I will submit a PR to enable ASan in Travis CI once this is in for discussion.

I added a note to the readme about the ASAN flag. I also noticed the macOS instructions didn't mention that make works there as well, so I added a note that the Linux instructions will work on that platform as well. I also added myself to the author list per your other comment.

tchipley:
@cglzaguilar: I don't really have time to do additional work on Vireo, tbh.

PR #538, opened 2018-11-08
Fix ArrayDelete bug

sanmut:
Fix ArrayDelete bug due to which more than the requested number of elements were getting deleted. Add more tests for 1D and 2D arrays for different values of index and length.


source/core/Array.cpp
```diff
@@ -1243,7 +1243,7 @@ VIREO_FUNCTION_SIGNATURE7(ArrayDeleteND, TypedArrayCoreRef, StaticType, void,
 
     IntIndex startIndex = offset > 0 ? offset : 0;
     IntIndex endIndex = (offset +
```
mehagar: Could you break apart this statement into variables so it's easier to read?

test-it/testList.json
```diff
@@ -326,6 +326,8 @@
                 "SubVISimple.via",
                 "TemplateBasics.via",
                 "TestVI_ArrayDelete.via",
+                "TestVI_ArrayDelete_2.via",
```
cglzaguilar: Please rename the test to provide more context as to what is being tested. 2 and 3 doesn't ring a bell.

test-it/ExpectedResults/TestVI_ArrayDelete_2.vtr
```diff
@@ -0,0 +1,6 @@
+TestResult_6 true
```
cglzaguilar: Please provide more context related to the case being exercised. Numbers 1 to 6 doesn't ring a bell.

source/core/Array.cpp
sanmut: Done.

test-it/ExpectedResults/TestVI_ArrayDelete_2.vtr
sanmut: Done

PR #541, opened 2018-11-29
Fixing async JSLI issue by printing to console.error

segaljared:
If a user calls a JSLI function like this:

```
window.performTask = function (jsapi) {
    var old = jsapi.getCompletionCallback();
    var cb = jsapi.getCompletionCallback();
    cb(5+5);
};
```

an Error will be thrown. This will be caught and put on the error wire, but could go unnoticed by the user. Furthermore, even if the user inspected the error wire (or it was printed to console), they would be missing the callstack that you would get by using console.error.

These changes fix that by adding console.error to the reportExecutionError function in the JSLI api.

### DIFF
--- a/source/io/module_javaScriptInvoke.js
+++ b/source/io/module_javaScriptInvoke.js
@@ -289,7 +289,7 @@ var assignJavaScriptInvoke;
                 // to abort the runtime at this point. https://github.com/ni/VireoSDK/issues/521
                 throw returnValue;
             }
-            console.error(returnValue.message);
+            console.error(returnValue);
 
             mergeNewError(errorValueRef, functionName, ERRORS.kNIUnableToInvokeAJavaScriptFunction, returnValue);
             completionCallbackStatus.retrievalState = completionCallbackRetrievalEnum.UNRETRIEVABLE;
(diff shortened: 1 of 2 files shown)


source/io/module_javaScriptInvoke.js
```diff
@@ -289,6 +289,7 @@ var assignJavaScriptInvoke;
                 // to abort the runtime at this point. https://github.com/ni/VireoSDK/issues/521
                 throw returnValue;
             }
+            console.error(returnValue.message);
```
rajsite: Can you do `console.error(returnValue)` instead? Let the browser decide how to handle the Error object (the returnValue.message does not include the stack trace in all browsers, etc)

test-it/karma/javascriptinvoke/Async.Test.js
```diff
@@ -335,6 +335,50 @@ describe('A JavaScript function invoke', function () {
...
+                expect(viPathParser('error.code')).toBe(44300);
+                expect(viPathParser('error.source')).toMatch(/retrieved more than once/);
+                expect(viPathParser('return')).toBe(0);
+            };
+        });
+        it('using the completion callback', async function () {
+            window.SingleFunction = function (input, jsapi) {
+                var completionCallback = jsapi.getCompletionCallback();
+                jsapi.getCompletionCallback();
+                // TODO mraj this allows completing after an exception is thrown, this should not be allowed.
```
rajsite: Can you remove these TODOs from here and above? The strategy has changed such that if the getCompletionCallback throws an error and a user catches it, it does make sense for the user to complete the test (we should remove the comment in the above tests).

PR #544, opened 2018-12-03
Refactor cpp files

mehagar:
This is a sampling of the kinds of refactoring that Resharper can easily find.

### DIFF
--- a/source/core/NumericString.cpp
+++ b/source/core/NumericString.cpp
@@ -208,7 +208,6 @@ void ReadPercentFormatOptions(SubString *format, FormatOptions *pOptions)
 //---------------------------------------------------------------------------------------------
 void GenerateFinalNumeric(const FormatOptions*, char*, Int32*, TempStackCString*, Boolean);
 void RefactorLVNumeric(const FormatOptions*, char* , Int32* , Int32 , Int32, Boolean);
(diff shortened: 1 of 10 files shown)


cglzaguilar:
For the next round of fixing, is it possible to focus on one type of warning? like for example unused variables, so we know the refactor is only about that.

rajsite:
Wonder if we can implement a more thorough linter like clang-tidy on the CI to keep it clean: http://clang.llvm.org/extra/clang-tidy/

Or enabling more gcc warnings. Pretty sure no unused variables is an option amongst others.

mehagar:
> For the next round of fixing, is it possible to focus on one type of warning? like for example unused variables, so we know the refactor is only about that.

Yeah, I can do that.

mehagar:
> Wonder if we can implement a more thorough linter like clang-tidy on the CI to keep it clean: [http://clang.llvm.org/extra/clang-tidy](http://clang.llvm.org/extra/clang-tidy/)
> 
> Or enabling more gcc warnings. Pretty sure no unused variables is an option amongst others.
> 
> Might be a good Prime Day project.

Resharper uses clang-tidy for some of its suggestions. Once I fix the existing warnings I would like to enable this on the CI.

PR #545, opened 2018-12-06
Variant and and variant attribute support

sanmut:
Add support for variant datatype and primitives DataToVariant and
VariantToData.
Add Support Get/Set/GetAll/Delete Variant Attributes


source/core/Variants.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+        if (variantType->IsA(destType)) {
+            void* destData = _ParamImmediate(DestData)._pData;
+            variantType->CopyData(variantType->Begin(kPARead), destData);
+        } else {
+            if (errPtr) {
+                errPtr->SetError(true, 91, "Variant To Data");
+            }
+        }
+    }
```
spathiwa: The primitive write default data to its output when an error is passed in, or when the destType doesn't match the variant's contained type.

source/core/Variants.cpp
```diff
@@ -0,0 +1,340 @@
...
+    if (_ParamPointer(Replaced)) {
+        _Param(Replaced) = replaced;
+    }
+    return _NextInstruction();
+}
+
+struct GetVariantAttributesParamBlock : public VarArgInstruction
```
spathiwa: Name should be singular, returns only one attribute.

source/core/Variants.cpp
```diff
@@ -0,0 +1,340 @@
...
+    TypeRef variantType = DefaultValueType::New(tm, inputType, true);
+    variantType->CopyData(_ParamImmediate(InputData._pData), variantType->Begin(kPAWrite));
+    _Param(OutputVariant) = variantType;
+    return _NextInstruction();
+}
+
+//------------------------------------------------------------
+struct VariantToDataParamBlock : public VarArgInstruction
+{
+    _ParamDef(TypeRef, VariantType);
```
spathiwa: Rename to just Variant?  It provides an input value, not just a type.

test-it/ViaTests/VariantCopy.via
```diff
@@ -0,0 +1,80 @@
...
+        PrintVariantWithAttributes(myVar2 "myVar2")
+
+        DeleteVariantAttribute(myVar * found error)
+        Printf("After Delete in myVar():\n")
+        PrintVariantWithAttributes(myVar "myVar")
+        PrintVariantWithAttributes(myVar2 "myVar2")
+    )
+)))
+
+enqueue(TestCopyVariantWithAttributes)
```
cglzaguilar: Add a new line to remove the "No newline at end of file" red glyph on git review window.

source/core/Variants.cpp
sanmut: Fixed. Thanks.

PR #546, opened 2018-12-06
[Refactoring] redundant boolean literal fix

mehagar:
This is the readability-simplify-boolean-expr clang-tidy check.


sanmut:
Thanks for doing this. However, this effort may not go wasteful only when we use clang-tidy in our toolchain / build process (appveyor.yml ?) to run automatically and add each setting such as this one (readability-simplify-boolean-expr) to a config file (.clang-tidy config file in the root directory to start with).

Developers can use the plugin https://marketplace.visualstudio.com/items?itemName=caphyon.ClangPowerTools with Visual Studio to both fix and catch issues using the .clang-tidy file in the root project directory.

PR #548, opened 2018-12-10
upgrade cpplint version to latest

mehagar:
I copied the latest cpplint from https://github.com/google/styleguide/blob/gh-pages/cpplint/cpplint.py into our source, and fixed the lint errors it now showed. I had to turn off -whitespace/braces warnings because it was enforcing K&R style, and our coding guidelines are based off of the linux kernel ones which do not use K&R for functions.

### DIFF
--- a/make-it/cpplint/Makefile
+++ b/make-it/cpplint/Makefile
@@ -1,4 +1,4 @@
-LINTARGS = --linelength=160 --filter=-build/header_guard,-readability/casting
+LINTARGS = --linelength=160 --filter=-build/header_guard,-readability/casting,-whitespace/braces
 
 # Ignored warnings:
 # build/header_guard -- wants include guard to include file path (_SOURCE_INCLUDE_) prefixes, which is unnecessary since all of our includes are in the same folder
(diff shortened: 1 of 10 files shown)


make-it/cpplint/Makefile
```diff
@@ -1,4 +1,4 @@
-LINTARGS = --linelength=160 --filter=-build/header_guard,-readability/casting
+LINTARGS = --linelength=160 --filter=-build/header_guard,-readability/casting,-whitespace/braces
```
mehagar: In the next change I am going to remove this -readability/casting filter and replace all C-style casts with C++ style casts.

rajsite:
I'll leave it up to the cpp folks to decide the braces stuff.

My general philosophy is that the linter should be enforcing style so if you are disable the brace linting rule then I don't know why all the other brace changes are in this PR. If I am misunderstanding and they are being forced then ignore this :)

mehagar:
> I'll leave it up to the cpp folks to decide the braces stuff.
> 
> My general philosophy is that the linter should be enforcing style so if you are disable the brace linting rule then I don't know why all the other brace changes are in this PR. If I am misunderstanding and they are being forced then ignore this :)

Yeah, I should undo the other brace changes.

PR #549, opened 2018-12-10
use c++ style cast

mehagar:
I want to eventually enable the cpplint rule to enable prevent using c-style casts. This was about 1/5th of the C-style casts here.

### DIFF
--- a/source/core/Date.cpp
+++ b/source/core/Date.cpp
@@ -277,7 +277,7 @@ namespace Vireo {
             if (*timeZoneString != nullptr) {
                 free(*timeZoneString);
             }
-            *timeZoneString = (char *)malloc(strlen(timeZoneAbbr)+1);
+            *timeZoneString = static_cast<char *>(malloc(strlen(timeZoneAbbr) + 1));
             strncpy(*timeZoneString, timeZoneAbbr, strlen(timeZoneAbbr)+1);
         }
     }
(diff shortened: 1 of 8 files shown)


source/core/Date.cpp
```diff
@@ -301,7 +301,7 @@ namespace Vireo {
             gmtime_r(&time, &tm);
         else
             localtime_r(&time, &tm);
-        _timeZoneOffset = int(isUTC ? 0 : tm.tm_gmtoff);
+        _timeZoneOffset = static_cast<int>(isUTC ? 0 : tm.tm_gmtoff);
```
spathiwa: Nothing wrong with a constructor cast here.

source/core/Date.cpp
```diff
@@ -349,7 +349,7 @@ namespace Vireo {
         struct tm tm;
         time_t timeVal = utcTime - kStdDT1970re1904;
         localtime_r(&timeVal, &tm);
-        _systemLocaleTimeZone = int(tm.tm_gmtoff);
+        _systemLocaleTimeZone = static_cast<int>(tm.tm_gmtoff);
```
spathiwa: and here.

source/core/Date.cpp
cglzaguilar: Yes, but let's use the static_cast to keep the code using only one kind of casting.

source/core/Date.cpp
cglzaguilar: Yes, but let's use the static_cast to keep the code using only one kind of casting.

PR #553, opened 2018-12-11
[Refactoring] remove redundant qualifier

mehagar:
These qualifiers were not needed.

### DIFF
--- a/source/core/TDCodecVia.cpp
+++ b/source/core/TDCodecVia.cpp
@@ -189,7 +189,7 @@ NIError TDViaParser::ParseREPL()
         RepinLineNumberBase();
     }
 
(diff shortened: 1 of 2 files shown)


source/core/TDCodecVia.cpp
```diff
@@ -2117,7 +2117,7 @@ void TDViaFormatter::FormatIEEE754(TypeRef type, void* pData)
         } else {
             _errorCode = kLVError_JSONBadNaN;
         }
-    } else if (::isinf(value)) {
+    } else if (isinf(value)) {
```
spathiwa: I think these :: qualifiers might be necessary on some platform, perhaps Mac.

PR #566, opened 2019-01-18
Fix for UnflattenFromJSON incorrect handling of empty arrays (CAR 716604)

spathiwa:


make-it/cpplint/Makefile
```diff
@@ -1,4 +1,4 @@
-LINTARGS = --linelength=160 --filter=-build/header_guard,-readability/casting,-whitespace/braces
```
mehagar: Thanks for this.

source/core/TDCodecVia.cpp
```diff
@@ -783,7 +783,7 @@ TypeRef TDViaParser::ParseDefaultValue(Boolean mutableValue)
     return cdt;
 }
 //------------------------------------------------------------
-Boolean TDViaParser::PreParseElements(Int32 rank, ArrayDimensionVector dimensionLengths)
+Boolean TDViaParser::PreParseElements(Int32 rank, ArrayDimensionVector dimensionLengths, Int32 *reachedDepth /*= NULL*/)
```
mehagar: Could you use nullptr here?

source/include/TDCodecVia.h
```diff
@@ -121,7 +121,7 @@ class TDViaParser
     Boolean EatJSONPath(SubString* path);
     NIError ParseREPL();
     TypeRef ParseEnqueue();
-    Boolean PreParseElements(Int32 rank, ArrayDimensionVector dimensionLengths);
+    Boolean PreParseElements(Int32 rank, ArrayDimensionVector dimensionLengths, Int32 *reachedDepth = NULL);
```
mehagar: Use nullptr here instead of NULL.

source/core/TDCodecVia.cpp
spathiwa: Sure, done.

source/include/TDCodecVia.h
spathiwa: 👍

source/core/TDCodecVia.cpp
```diff
@@ -804,6 +804,8 @@ Boolean TDViaParser::PreParseElements(Int32 rank, ArrayDimensionVector dimension
     // The opening array_pre like "(" has been parsed before this function has been called.
     Int32 dimIndex;
     while (depth >= 0) {
```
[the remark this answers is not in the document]
spathiwa: It doesn't depend on the caller initializing it.  The while loop already always runs at least once.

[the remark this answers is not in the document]
spathiwa: Ok, done.

PR #568, opened 2019-01-23
Refactor support for formatting and scanning time codes.

spathiwa:
Addresses CAR 720195.


spathiwa:
Yeah, I know, it's because windows mktime/mkgmtime doesn't support dates before 1970.  I'm working on implementing it from scratch.

sanmut:
@spathiwa  Thanks for massive refactoring and addressing many issues in this PR. This change being very extensive and late in the cycle, this change should be validated against G test suite before merging. I also highly recommend adding GVI version of the new test cases added here to make sure they pass on Desktop and Web.

PR #569, opened 2019-01-25
Remove error logging for unhandled JS errors

rajsite:
Previously in https://github.com/ni/VireoSDK/pull/541, we introduced error logging for uncaught JS errors from users. The assumption was that all JS exceptions caught by JavaScriptInvoke were unexpected / unhandled and worthy of logging for additional context for the user.

While implementing the WebSockets api we found it was necessary to report JS errors back to a GVI to handle from the LabVIEW NXG diagram. So we have to signal to NXG that a JavaScriptInvoke call is in an expected error state. While internal functions are capable of reporting that state through the jsapi.setLabVIEWError function, that capability is not available to general JavaScriptInvoke users yet (and we are not confident the setLabVIEWError api is necessarily a good enough api to expose to users yet).

Instead we will disable the console logging JS errors feature as some of the initial concerns are already mitigated by the automatic error handling feature. This PR also improves the quality of JS errors reported to the error wire by including the JS stack trace as well which further mitigates the need to perform the console logging.

Lastly, during implementation it was observed that it is possible for JS users to throw non-error objects and those would be treated as return values instead of as errors. This change includes a fix such that when we are on a known error path (synchronous catch or Promise reject) we coerce the thrown value to an Error object so that it is consistently reported as an error.

### DIFF
--- a/source/core/module_coreHelpers.js
+++ b/source/core/module_coreHelpers.js
@@ -310,7 +310,7 @@ var assignCoreHelpers;
                 // Some browsers do not print the message in the stack so print both
                 if (typeof exception.message === 'string' && exception.message.length !== 0) {
                     additionalInfo = ', Additional information:\nMessage: ' + exception.message;
-                    if (typeof exception.stack === 'string' && exception.message.length !== 0) {
+                    if (typeof exception.stack === 'string' && exception.stack.length !== 0) {
                         additionalInfo += '\nStack: ' + exception.stack;
                     }
                     return messageText + additionalInfo;
--- a/source/core/module_coreHelpers.js
+++ b/source/core/module_coreHelpers.js
@@ -305,8 +305,16 @@ var assignCoreHelpers;
         };
 
         Module.coreHelpers.formatMessageWithException = function (messageText, exception) {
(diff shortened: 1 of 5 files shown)


mehagar:
Is this part of a US/CAR? If so could you link that here? 😉

source/core/module_coreHelpers.js
```diff
@@ -305,8 +305,16 @@ var assignCoreHelpers;
...
 
         Module.coreHelpers.formatMessageWithException = function (messageText, exception) {
-            if (exception !== undefined && exception !== null && typeof exception.message === 'string' && exception.message.length !== 0) {
-                return messageText + ', Additional information: ' + exception.message;
+            var additionalInfo;
+            if (exception !== undefined && exception !== null) {
+                // Some browsers do not print the message in the stack so print both
+                if (typeof exception.message === 'string' && exception.message.length !== 0) {
+                    additionalInfo = ', Additional information:\nMessage: ' + exception.message;
+                    if (typeof exception.stack === 'string' && exception.message.length !== 0) {
```
cglzaguilar: I think the check for length here, should be for exception.stack, no exception.message

rajsite: good catch! thanks. updated

PR #570, opened 2019-01-25
Implement SystemLogging_WriteMessageUTF8

rajsite:
This PR implements the SystemLogging_WriteMessageUTF8 instruction as called by the SLI document of the Write to System Log VI.

Implementing this instruction allows the Write to System Log VI to be used on the Web Server target.

### DIFF
--- a/source/io/FileIO.cpp
+++ b/source/io/FileIO.cpp
@@ -401,8 +401,7 @@ VIREO_FUNCTION_SIGNATURE5(SystemLogging_WriteMessageUTF8, StringRef, StringRef, 
     if (!_Param(4).status) {
         jsSystemLogging_WriteMessageUTF8(
             _Param(2)->Type(), _ParamPointer(2),
-            typeRefInt32, _ParamPointer(3)
-        );
+            typeRefInt32, _ParamPointer(3));
     }
     return _NextInstruction();
 }
--- a/source/io/FileIO.cpp
+++ b/source/io/FileIO.cpp
@@ -45,6 +45,14 @@
 
 namespace Vireo {
 
+#if kVireoOS_emscripten
+extern "C" {
+    extern void jsSystemLogging_WriteMessageUTF8(
+        TypeRef messageType, StringRef * message,
+        TypeRef severityType, Int32 * severity);
+}
+#endif
+
 typedef Int32 FileHandle;
 
 #ifdef VIREO_FILESYSTEM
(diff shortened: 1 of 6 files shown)


source/io/FileIO.cpp
```diff
@@ -383,6 +391,22 @@ VIREO_FUNCTION_SIGNATURE2(Println, StaticType, void)
 }
 #endif
 //------------------------------------------------------------
+// An instruction to enable the Write to System Log VI which uses an SLI call to SystemLogging_WriteMessageUTF8
+// ignored(0), ignored(1), message(2), severity(3), error cluster(4)
+#if kVireoOS_emscripten
+VIREO_FUNCTION_SIGNATURE5(SystemLogging_WriteMessageUTF8, StringRef, StringRef, StringRef, Int32, ErrorCluster)
```
sanmut: It is better to use VIREO_FUNCTION_SIGNATUREV instead and use Parameter block structure that inherits from InstructionCore where each parameter can be named. See., for example, StringFormat

source/io/FileIO.cpp
```diff
@@ -396,6 +420,10 @@ DEFINE_VIREO_BEGIN(FileSystem)
     DEFINE_VIREO_FUNCTION(Printf, "p(i(VarArgCount)i(String)i(StaticTypeAndData))")
 #endif
     //--------
+#if kVireoOS_emscripten
```
sanmut: I think just ErrorCluster can be used instead of " ERROR_CLUST_TYPE_STRING "

source/io/FileIO.cpp
rajsite: So far `VIREO_FUNCTION_SIGNATUREV` has only been used for functions with a variable number of arguments. Are we making some policy change to use it for all function signatures?

rajsite: Talked to @sanmut and he provided an example not using variable arguments with the StringToByteArray function: https://github.com/ni/VireoSDK/blob/0c290475a41043f52c40b7270e333125ac396ce6/source/core/NumericString.cpp#L2135

rajsite: Much cleaner, thanks @sanmut!

I also created https://github.com/ni/VireoSDK/issues/575 as a spot we should temporarily add information about creating new Vireo instructions. If you can think of anything else that is tribal knowledge making a quick note would be great. :)

sanmut: Sure, thanks :+1:

PR #571, opened 2019-01-30
Build array fixes

sanmut:
This PR fixes two separate issues on BuildArray:
1. In the non-concatenate mode of BuildArray where output is one rank higher than the inputs, output was not padded with default value correctly when the inputs have different dimension lengths in different dimensions.
2. When output array is resized based on the input arrays and when one of the inputs is not same as the output array, elements of output array should be initialized with default value correctly. If not done, then the output array may have left over incorrect values for some elements.


source/core/GenericFunctions.cpp
```diff
@@ -1695,8 +1695,17 @@ VIREO_FUNCTION_SIGNATUREV(ArrayConcatenateInternal, ArrayConcatenateInternalPara
...
         ArrayDimensionVector tempDimensionLengths, origDimensionLengths, origSlab;
         Int32 originalOuterDimSize = pDest->DimensionLengths()[outputRank-1];
+        Boolean isDestSameAsFirstInput = false;  // true implies inplaceness case
         for (i = 0; i < numInputs; i++) {
             TypedArrayCoreRef arrayInput = *((TypedArrayCoreRef *) inputs[i]);
+            if (pDest == arrayInput) {
+                if (i == 0) {
+                    isDestSameAsFirstInput = true;
+                } else {
+                    THREAD_EXEC()->LogEvent(EventLog::kHardDataError, "Illegal ArrayConcatenate inplaceness");
```
sanmut: Moved this check early so that we bail out soon for this case where output is same as an input that is not the first input.

sanmut:
@cglzaguilar , I ran the G test suite and they all seem to pass.

PR #572, opened 2019-02-05
Refactor Events files

mehagar:

local variable may be const
Use auto
member function may be const
macro argument should be enclosed in parentheses
use =delete to prohibit calling of a special member function
the empty method should be used to check for emptiness instead of size
result of a postfix operator is discarded, consider replacing it with a prefix operator.
member function may be static
redundant else keyword
use range-based for loop instead
inline specifier is redundant on a global static function
use emplace_back instead of push_back

### DIFF
--- a/source/core/Events.cpp
+++ b/source/core/Events.cpp
@@ -60,11 +60,10 @@ UInt32 EventData::_s_eventSequenceNumber = 0;
 class EventRegQueueID {
  public:
     EventRegQueueID(EventQueueID qID, RefNum ref) : _qID(qID), _ref(ref) { }
+    EventRegQueueID() = delete;
     EventQueueID _qID;
     RefNum _ref;
     // bool _locksPanel;
- private:
-    EventRegQueueID();
 };
 typedef std::list<EventRegQueueID> EventRegQueueIDList;
 
@@ -96,9 +95,9 @@ class EventQueueObject {
         _eventQueue.push_back(eData);
         _eventQueue.back().common.eventSeqIndex = EventData::GetNextEventSequenceNumber();
         if (eData.pEventData) {  // make a unique copy of the event data for each event queue
-            Int32 topSize = eData.eventDataType->TopAQSize();
+            const Int32 topSize = eData.eventDataType->TopAQSize();
             void *pEvent = THREAD_TADM()->Malloc(topSize);
-            eData.eventDataType->InitData(pEvent, (TypeRef)nullptr);
+            eData.eventDataType->InitData(pEvent, static_cast<TypeRef>(nullptr));
             eData.eventDataType->CopyData(eData.pEventData, pEvent);
             _eventQueue.back().pEventData = pEvent;
         }
@@ -107,7 +106,7 @@ class EventQueueObject {
     }
     void ClearQueue() {
         VIREO_ASSERT(!_eventLock)
-        bool keepGoing = true;
+        bool keepGoing;
         do {
             keepGoing = DiscardTopEvent();
         } while (keepGoing);
@@ -116,13 +115,15 @@ class EventQueueObject {
         ClearQueue();
         SetStatus(kQIDFree);
     }
-    const bool GetPendingEventSequenceNumber(UInt32 *eventSeq) {
+
+    bool GetPendingEventSequenceNumber(UInt32 *eventSeq) {
         if (size() > 0) {
             *eventSeq = _eventQueue.front().common.eventSeqIndex;
             return true;
         }
         return false;
     }
+
     const EventData &GetEventData() {
         if (size() > 0) {
             _eventLock = true;
@@ -130,6 +131,7 @@ class EventQueueObject {
         }
         return _timeOutEvent.Init(kEventSourceLVUserInt, kEventTypeTimeout, RefNumVal());
     }
+
     void SetObserver(OccurrenceCore *occ) {
         if (_wakeUpOccur != occ && occ) {
             size_t i = _eventQueue.size();
@@ -153,7 +155,7 @@ class EventQueueObject {
 
  private:
     bool DiscardTopEvent() {
-        if (_eventQueue.size() > 0) {
+        if (!_eventQueue.empty()) {
             EventData &eventData = _eventQueue.front();
             eventData.Destroy();
             _eventQueue.pop_front();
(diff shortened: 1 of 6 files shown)


source/core/Events.cpp
```diff
@@ -466,9 +476,9 @@ class DynamicEventRegInfo {
     }
 
     ~DynamicEventRegInfo() {
-         for (size_t i = 0; i < _entry.size(); i++) {
+         for (auto& entry : _entry)
```
sanmut: Prefer const auto &

source/core/Events.cpp
```diff
@@ -775,9 +782,10 @@ static bool UnregisterForEventsAux(RefNum refnum) {
     DynamicEventRegInfo *regInfo = nullptr;
     if (EventRegistrationRefNumManager::RefNumStorage().DisposeRefNum(refnum, &regInfo) != kNIError_Success || !regInfo)
         return false;
-    EventQueueID qID = regInfo->_qID;
+    const EventQueueID qID = regInfo->_qID;
     EventOracle::TheEventOracle().DeleteEventQueue(qID);
-    std::vector<DynamicEventRegEntry>::iterator regInfoEntryIter = regInfo->_entry.begin(), regInfoEntryIterEnd = regInfo->_entry.end();
+    auto regInfoEntryIter = regInfo->_entry.begin();
+    const std::vector<DynamicEventRegEntry>::iterator regInfoEntryIterEnd = regInfo->_entry.end();
```
sanmut: Use const auto instead

source/include/Events.h
```diff
@@ -77,21 +77,23 @@ struct EventCommonData {
     static UInt32 GetNumberOfCommonElements() { return 5; }  // change if fields are added above
 
     void InitEventTime() { eventTime = UInt32(gPlatform.Timer.TickCountToMilliseconds(gPlatform.Timer.TickCount())); }
-    EventCommonData(UInt32 source, UInt32 type) : eventSource(source), eventType(type), eventSeqIndex(0) {
+
+    EventCommonData(const UInt32 source, const UInt32 type) : eventSource(source), eventType(type), eventSeqIndex(0) {
```
sanmut: > UInt32 source ==> const UInt32 source

I can't seem to find a clang-tidy rule that enforces this. Did you find one?

source/include/Events.h
```diff
@@ -77,21 +77,23 @@ struct EventCommonData {
...
         InitEventTime();
     }
-    EventCommonData(UInt32 source, UInt32 type, const RefNumVal &ref) : eventSource(source), eventType(type), eventSeqIndex(0), eventRef(ref) {
+
+    EventCommonData(const UInt32 source, UInt32 type, const RefNumVal &ref) : eventSource(source), eventType(type), eventSeqIndex(0), eventRef(ref) {
```
cglzaguilar: Please elaborate why const is not applied to type as in the previous function?

source/include/Events.h
```diff
@@ -104,11 +106,11 @@ struct EventData {
     void *pEventData;
 
     EventData() : common(0, 0, RefNumVal()), controlUID(kNotAnEventControlUID), eventDataType(nullptr), pEventData(nullptr) { }
-    EventData(EventSource source, EventType type, const RefNumVal &ref, TypeRef edtype = nullptr, void *pData = nullptr) :
+    EventData(const EventSource source, const EventType type, const RefNumVal &ref, TypeRef edtype = nullptr, void *pData = nullptr) :
```
cglzaguilar: In these three constructors, it seems we are applying const to some parameters and not to others, that is more confusing. We need to be consistent.

cglzaguilar:
@spathiwa, please look at these changes.

source/include/Events.h
spathiwa: const serves no purpose in either declaration and should be removed.

source/include/Events.h
spathiwa: remove const modifiers.

source/include/Events.h
spathiwa: I think he should not be adding const to any scalar non-ref parameters or locals unless there is a good, conceptual reason, such as a constant used as a static limit or computed size based on sizeof() calculations.
Just because the current usage happens to not modify the value after assignment does not mean it is beneficial to mark it const.

source/include/Thread.h
```diff
@@ -8,7 +8,7 @@ SDG
 */
 
 /*! \file
-    \brief Tools for working in a multithreaded process.
+    \brief Tools for working in a multi threaded process.
```
spathiwa: Use hyphen.

source/include/TypeDefiner.h
```diff
@@ -132,15 +132,15 @@ class TypeDefiner
...
         kPTInstructionFunction, #_name_));
 
     #define DEFINE_VIREO_FUNCTION_CUSTOM(_name_, _cfunction_, _typeTypeString_) \
-      (TypeDefiner::DefineCustomPointerTypeWithValue(tm, #_name_, (void*)_cfunction_, _typeTypeString_, \
+      (TypeDefiner::DefineCustomPointerTypeWithValue(tm, #_name_, (void*)(_cfunction_), _typeTypeString_, \
         kPTInstructionFunction, #_cfunction_));
 
     #define DEFINE_VIREO_GENERIC(_name_, _typeTypeString_, _genericEmitProc_) \
-      (TypeDefiner::DefineCustomPointerTypeWithValue(tm, #_name_, (void*)_genericEmitProc_, _typeTypeString_, \
+      (TypeDefiner::DefineCustomPointerTypeWithValue(tm, #_name_, (void*)(_genericEmitProc_), _typeTypeString_, \
```
spathiwa: _typeTypeStirng should also be parenthesized.

source/core/Events.cpp
```diff
@@ -414,13 +423,15 @@ class DynamicEventRegInfo {
...
-    Int32 DynamicEventMatchCore(TypeRef regRefType, void *pData, RefNum refnum, Int32 *dynIndexBase) {
+    static Int32 DynamicEventMatchCore(TypeRef regRefType, void *pData, RefNum refnum, Int32 *dynIndexBase) {
         Int32 dynIndex = 0;
         if (regRefType->IsRefnum() && pData && (static_cast<RefNumVal*>(pData)->GetRefNum() == refnum)) {
             dynIndex = *dynIndexBase;
             return dynIndex;
-        } else if (regRefType->IsArray() && regRefType->Rank() == 1 && regRefType->GetSubElement(0)->IsRefnum()) {
-            TypedArray1D<RefNumVal> *aRef = *(TypedArray1D<RefNumVal>**)pData;
+        }
+
```
spathiwa: I don't agree with the 'extraneous else' rule.
It doesn't hurt anything and helps readability, because you don't have to look inside the previous clauses for early returns to understand control-flow.

source/core/Events.cpp
```diff
@@ -540,7 +550,7 @@ struct EventSpec {  // Specifier for Event structure configuration data
 typedef EventSpec *EventSpecRef;
 
 // GetEventSourceForEventType -- return the EventSource associated with an event type
-static inline EventSource GetEventSourceForEventType(EventType eType) {
+static EventSource GetEventSourceForEventType(EventType eType) {
```
spathiwa: Leave inline.

mehagar: Even though the inline is unnecessary because it is a global function?

source/core/Events.cpp
mehagar: You still need to look inside the if statement to see if a return statement is being used because there may be code after the if-elseif chain, and that code will not be executed if the first if (with the early return) gets executed. That's not the case in this function, but generally it is possible.

source/include/Events.h
mehagar: Yeah, I will revert these changes.

source/core/Events.cpp
spathiwa: It's static because it doesn't need to be extern, it's not used outside this file, and it's inline because I want to instruct the compiler to inline it if possible.
Many compilers will already try to inline static functions if possible, but it's still useful to provide the hint.
We could use an anonymous namespace instead of static for local scope, but we haven't been using that technique elsewhere.

source/core/Events.cpp
spathiwa: Regardless, the else if helps readability.  Leave it.

spathiwa:
Events.cpp can no longer be diffed for some reason.
It shows the entire contents of the file as changed; one big red block followed by one big green one.
Do you know what happened?  This makes it impossible to review the changes.

mehagar:
> Events.cpp can no longer be diffed for some reason.
> It shows the entire contents of the file as changed; one big red block followed by one big green one.
> Do you know what happened? This makes it impossible to review the changes by file.
> I can tell you removed consts, etc. from the commit diff, but I like to look at the file diff to see the remaining changes from the original.
> It looks like the CIs failed because of 'null' bytes in the source, so it looks like the file is corrupt somehow. Please check it.

I'm not sure what happened. I tried to push again but nothing happened. I might delete this PR and start again.

mehagar:
This is starting to be not worth my time.

PR #573, opened 2019-02-05
Refactor Arrays files

mehagar:

local variable may be const
local variable 'x' is reassigned in all paths before being read.
inline specifier is redundant on a function declared entirely inside a struct definition.
Use auto

### DIFF
--- a/source/core/Array.cpp
+++ b/source/core/Array.cpp
@@ -34,7 +34,7 @@ DECLARE_VIREO_PRIMITIVE2(ArrayElementType, TypedArrayCoreRef, TypeRef, (_Param(1
  * */
 VIREO_FUNCTION_SIGNATURE2(ArrayLengthN, TypedArrayCoreRef, TypedArray1D<IntIndex>*)
 {
-    IntIndex rank = _Param(0)->Rank();
+    const IntIndex rank = _Param(0)->Rank();
     IntIndex* pLengths = _Param(0)->DimensionLengths();
     TypeRef elementType = _Param(1)->ElementType();
     _Param(1)->Resize1D(rank);
@@ -53,7 +53,7 @@ VIREO_FUNCTION_SIGNATURE2(ArrayCapacity, TypedArrayCoreRef, IntIndex)
 //------------------------------------------------------------
 VIREO_FUNCTION_SIGNATURE2(ArrayDimensions, TypedArrayCoreRef, TypedArray1D<IntIndex>*)
 {
-    IntIndex rank = _Param(0)->Rank();
+    const IntIndex rank = _Param(0)->Rank();
     IntIndex* pLengths = _Param(0)->DimensionLengths();
     _Param(1)->Replace1D(0, rank, pLengths, true);
     return _NextInstruction();
@@ -62,7 +62,7 @@ VIREO_FUNCTION_SIGNATURE2(ArrayDimensions, TypedArrayCoreRef, TypedArray1D<IntIn
 VIREO_FUNCTION_SIGNATURE2(ArrayResizeDimensions, TypedArrayCoreRef, TypedArray1D<IntIndex>*)
 {
     // Details on how arrays are redimensioned are in ResizeDimensions().
-    IntIndex rankProvided = _Param(1)->Length();
+    const IntIndex rankProvided = _Param(1)->Length();
     IntIndex* pLengths = _Param(1)->Begin();
     _Param(0)->ResizeDimensions(rankProvided, pLengths, false);
     return _NextInstruction();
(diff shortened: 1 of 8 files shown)


sanmut:
If you are using clang-tidy rules to clean up, could you mention the rule(s) you are using to refactor in every PR.

source/core/Array.cpp
```diff
@@ -1569,15 +1561,15 @@ VIREO_FUNCTION_SIGNATURE4(ArrayThreshold, Double, TypedArrayCoreRef, Double, Int
         }
     } else {
         for (IntIndex i = startIndex; i < arrayIn->Length(); i++) {
-            Double y = ReadDoubleFromMemory(arrayIn->ElementType(), arrayIn->BeginAt(i));
-            if (i == startIndex) { left = right = (Double)startIndex;}
+            const Double y = ReadDoubleFromMemory(arrayIn->ElementType(), arrayIn->BeginAt(i));
+            if (i == startIndex) { right = static_cast<Double>(startIndex);}
             if (y < thresholdY) {
```
sanmut: I tend to think initialing left and right in one line is cleaner/clearer to read. Is there a clang-tidy rule that enforces this?

rajsite: You can see from his PR comment that Resharper is being used locally, not clang-tidy. We do not have clang-tidy configured locally or on the CI.

sanmut: I am not sure what his configuration is. If clangtools is installed, Resharper can be configured to check for clang-tidy issues.  I would like for us to enable first only rules that we can enforce with automated build. This particular one found by Resharper does not seem to be a quite safe one to apply. If left is used below before assignment in code maintenance, it is not a warning of any sort. Here, the programmer's intent of initializing both left and right seems more maintainable. So, if it is just a Resharper suggestion, we should not simply apply it. After all, it is not enforceable anyway in automated builds.

sanmut: As mentioned in my comment in https://github.com/ni/VireoSDK/pull/572, most of his changes actually seem to correspond to clang-tidy rules.

mehagar: Yeah, resharper uses clang-tidy to supply some of its rules internally. And when you install resharper it comes with the clangtools I believe, because I didn't do any additional setup to get those clan-tidy rules.

And you're right, some of these were just suggestions while some were warnings.

source/core/Array.cpp
```diff
@@ -1587,9 +1579,9 @@ VIREO_FUNCTION_SIGNATURE4(ArrayThreshold, Double, TypedArrayCoreRef, Double, Int
         _Param(0) = left;
     } else if (leftValue == rightValue) {
         _Param(0) = (left + right) / 2;
-    } else if (::isinf(rightValue)) {
+    } else if (isinf(rightValue)) {
```
cglzaguilar: @spathiwa, I think you already mentioned in another PR that we need to keep the ::, do you remember why?

source/io/JavaScriptInvoke.cpp
```diff
@@ -31,13 +31,13 @@ extern "C" {
 
 //------------------------------------------------------------
 //! Return dataRef of the parameter that is at the given index in the parameters array
-VIREO_EXPORT void* JavaScriptInvoke_GetParameterDataRef(StaticTypeAndData *parameters, Int32 index)
+VIREO_EXPORT void* JavaScriptInvoke_GetParameterDataRef(StaticTypeAndData *parameters, const Int32 index)
```
spathiwa: Why?

source/core/Array.cpp
spathiwa: I think there is a Mac build issue without it, don't remember details.

source/core/Array.cpp
```diff
@@ -987,30 +982,30 @@ VIREO_FUNCTION_SIGNATURE4(ArraySplit, TypedArrayCoreRef, TypedArrayCoreRef, Type
...
+    TypedArrayCoreRef arrayArg = *static_cast<TypedArrayCoreRef*>(pInstructionBuilder->_argPointers[0]);
     TypeRef elementType = arrayArg->ElementType();
     SubString LTName("IsLTSort");
     // Add param slot to hold the snippet
-    Int32 snippetArgId = pInstructionBuilder->AddSubSnippet();
-    Sort1DArrayInstruction* sortOp = (Sort1DArrayInstruction*) pInstructionBuilder->EmitInstruction();
-    pInstruction = sortOp;
+    const Int32 snippetArgId = pInstructionBuilder->AddSubSnippet();
+    auto sortOp = static_cast<Sort1DArrayInstruction*>(pInstructionBuilder->EmitInstruction());
+    InstructionCore* p_instruction = sortOp;
```
spathiwa: Why??  pInstruction is fine, and consistent with pInstructionBuilder, which you did not change.

source/core/Array.cpp
mehagar: As to the original concern over whether it is cleaner, the specific resharper hint was because the variable was always being reassigned and so initializing it here didn't matter.

mehagar:
Closing because adding const to local variables was the bulk of the change, and that was not received well.

PR #574, opened 2019-02-06
Refactor TypeAndDataManager.h

mehagar:

local variable may be const
Use auto
member function may be const
macro argument should be enclosed in parentheses
redundant else keyword
use emplace_back instead of push_back
use =default to define a trivial constructor
avoid repeating the return type from the declaration, use a braced initializer instead

### DIFF
--- a/source/include/TypeAndDataManager.h
+++ b/source/include/TypeAndDataManager.h
@@ -28,7 +28,7 @@ SDG
 
 #define STL_MAP
 
-#include <stdlib.h>  // abs()
+#include <cstdlib>  // abs()
 #include <cmath>
 #include <new>       // for new placement
 
@@ -263,7 +263,7 @@ class TypeManager
     explicit TypeManager(TypeManagerRef parentTm);
     NamedTypeRef NewNamedType(const SubString* typeName, TypeRef type, NamedTypeRef existingOverload);
  public:
-    ExecutionContextRef TheExecutionContext() { return _executionContext; }
+    ExecutionContextRef TheExecutionContext() const { return _executionContext; }
     void    SetExecutionContext(ExecutionContextRef exec) { _executionContext = exec; }
     void    DeleteTypes(Boolean finalTime);
     void    TrackType(TypeCommon* type);
@@ -271,10 +271,10 @@ class TypeManager
 
     void    UntrackLastType(TypeCommon* type);
     void    GetTypes(TypedArray1D<TypeRef>*);
-    TypeRef TypeList() { return _typeList; }
+    TypeRef TypeList() const { return _typeList; }
     void    PrintMemoryStat(ConstCStr, Boolean bLast);
 
-    TypeManagerRef BaseTypeManager() { return _baseTypeManager; }
+    TypeManagerRef BaseTypeManager() const { return _baseTypeManager; }
     TypeRef Define(const SubString* typeName, TypeRef type);
 
     TypeRef FindType(ConstCStr name);
@@ -285,7 +285,7 @@ class TypeManager
     Int32   AQAlignment(Int32 size);
     static Int32   AlignAQOffset(Int32 offset, Int32 size) {
         if (size != 0) {
-            Int32 remainder  = offset % size;
+            const Int32 remainder  = offset % size;
             if (remainder)
                 offset += size - remainder;
         }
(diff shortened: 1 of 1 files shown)


source/include/TypeAndDataManager.h
```diff
@@ -1011,7 +1010,6 @@ class DefaultPointerType : public PointerType
 {
  private:
     DefaultPointerType(TypeManagerRef typeManager, TypeRef type, void* pointer, PointerTypeEnum pointerType);
-    DefaultPointerType();
```
sanmut: Why not "DefaultPointerType() = delete;"?

source/include/TypeAndDataManager.h
```diff
@@ -1111,18 +1110,16 @@ class TypedArrayCore
...
-    AQBlock1* RawBegin()            { return _pRawBufferBegin; }
+    void* RawObj() { VIREO_ASSERT(Rank() == 0); return RawBegin(); }  // some extra asserts fo  ZDAs
+    AQBlock1* RawBegin() { return _pRawBufferBegin; }
     template<typename CT> CT BeginAtAQ(IntIndex index) { return reinterpret_cast<CT>(RawBegin() + index); }
     BlockItr RawItr()               { return BlockItr(RawBegin(), ElementType()->TopAQSize(), Length()); }
 
- public:
     //! Array's type.
-    TypeRef Type()                  { return _typeRef; }
+    TypeRef Type() { return _typeRef; }
```
sanmut: It's ok here and no need to address this.. In the future, ,it is a good idea to do format/style changes (such as whitespace, newline) separately in a PR.

source/include/TypeAndDataManager.h
```diff
@@ -1245,7 +1242,7 @@ class String : public TypedArray1D< Utf8Char >
     void AppendEscapeEncoded(const Utf8Char* source, IntIndex len);
 
     void InsertSubString(IntIndex position, SubString* str) {
-        Insert(position, (IntIndex)str->Length(), (Utf8Char*)str->Begin());
+        Insert(position, static_cast<IntIndex>(str->Length()), const_cast<Utf8Char*>(str->Begin()));
```
sanmut: I think const_cast<Utf8Char*>is not required here. Just remove const_cast and even C style cast and try "str->Begin()"?

source/include/TypeAndDataManager.h
```diff
@@ -285,18 +285,17 @@ class TypeManager
     Int32   AQAlignment(Int32 size);
     static Int32   AlignAQOffset(Int32 offset, Int32 size) {
         if (size != 0) {
-            Int32 remainder  = offset % size;
+            const Int32 remainder  = offset % size;
```
spathiwa: Adding class const modifiers to member functions is great, but I'm not sure what the point of adding const modifiers to scalar arguments or local variables which are not used in a way that is _conceptually_ constant such as this.  It would be hard to be consistent with this, and I think it adds no value.

source/include/TypeAndDataManager.h
```diff
@@ -1326,16 +1325,20 @@ class StackVar
 
 struct StringRefCmp {
     bool operator()(const StringRef& a, const StringRef &b) const {
-        Int32 cmp = memcmp(a->Begin(), b->Begin(), Min(a->Length(), b->Length()));
+        const Int32 cmp = memcmp(a->Begin(), b->Begin(), Min(a->Length(), b->Length()));
```
spathiwa: Again, not sure why this is at all helpful.

source/include/TypeAndDataManager.h
mehagar: I agree. I'll remove uses of const for local variables.

source/include/TypeAndDataManager.h
mehagar: Because having a non default constructor declared already means that a default constructor won't be generated.

source/include/TypeAndDataManager.h
```diff
@@ -1329,13 +1328,17 @@ struct StringRefCmp {
         Int32 cmp = memcmp(a->Begin(), b->Begin(), Min(a->Length(), b->Length()));
         if (cmp < 0) {
             return true;
-        } else if (cmp > 0) {
+        }
```
spathiwa: Please undo the changes to this function.  Leave the 'else' statements.
This is a good example of a situation where the code reads much better with them, regardless of whether they are 'redundant'.

PR #580, opened 2019-03-04
Opaque JS Reference type

ggumen:
Allowing JSRefnum types to not just be internal-only (reverting part of commit b19704a)
Adding Static vs Dynamic reference distinction to JSRefnum
- Static reference (ie, Static control reference) shares cookie for the same jsValue.
- Dynamic reference (ie, JS opaque reference from JSLI) always creates a new cookie even for the same jsValue. This is so that when we close one reference, we don't invalidate another reference even if they are referring to the same JS object underneath it.


cglzaguilar:
As discussed, add additional info to the JavaScriptReference to know if it is a static or dynamic one.

source/io/module_javaScriptInvoke.js
```diff
@@ -111,19 +105,32 @@ var assignJavaScriptInvoke;
...
+         * Static reference (Static control reference) shares cookie for the same jsValue.
+         * Dynamic reference (JS opaque reference from JSLI) always creates a new cookie even for the same jsValue.
          */
-        Module.javaScriptInvoke.writeJavaScriptRefNum = function (javaScriptValueRef, jsValue) {
-            var cachedCookie = getCachedRefNumCookie(jsValue);
-            if (cachedCookie !== undefined) {
-                Module.eggShell.writeDouble(javaScriptValueRef, cachedCookie); // set the VIA local to be this cookie value
-                return;
+        var writeJavaScriptStaticAndDynamicRefNum = function (javaScriptValueRef, jsValue, isStaticReference) {
+            if (isStaticReference) { // static reference (ie, control reference) shares cookie for the same jsValue
```
cglzaguilar: We already discussed this. Please create a new Vireo API to check if the TypeRef in javaScriptValueRef is static or dynamic. That way you don't need to pass isStaticReference and you don't need two writeJavaScriptRefNum functions. I think clients don't need to know the difference.

ggumen: Done.

sanmut:
Deferring to @cglzaguilar, @rajsite, and @spathiwa as I may not find time to review soon. I will take a look at it later.

source/core/CEntryPoints.cpp
```diff
@@ -526,11 +526,16 @@ VIREO_EXPORT Boolean TypeRef_IsAnalogWaveform(TypeRef typeRef)
     return typeRef->IsAnalogWaveform();
 }
 //------------------------------------------------------------
-VIREO_EXPORT Boolean TypeRef_IsJavascriptRefnum(TypeRef typeRef)
+VIREO_EXPORT Boolean TypeRef_IsJavaScriptRefNum(TypeRef typeRef)
```
rajsite: Can we make a change so that we have both TypeRef_IsJavaScriptDynamicRefnum and TypeRef_IsJavaScriptStaticRefNum?

It maps better to how the types are actually implemented in Vireo as JavaScriptStaticRefnum and JavaScriptDynamicRefnum are distinct different types (there is not a JavaScriptRefnum type).

We could also have a helper TypeRef_IsJavaScriptRefNum if it's valuable but there are so few usages I don't think it is necessary. 

The following call site in the typeVisitor would have to be updated to check both cases: https://github.com/ni/VireoSDK/blob/master/source/core/module_typeHelpers.js#L260

Talking to Carlos there may be benefit in the future of making the typeVisitor being aware of static vs dynamic types as well. One example is that the JSLI's type visitor (that is user facing) could enforce that only dynamic references are allowed while the Editor side type visitor (for internal functions) would take both static and dynamic references. We won't make that change now as that would be a public api change.

ggumen: Done! Thanks!

cglzaguilar:
@ggumen already ran the changes on AzDo with a local build of Vireo and they are passing.

PR #582, opened 2019-03-07
Added extensive tests for Log10

rkudukoli:


test-it/ViaTests/Log10.via
```diff
@@ -0,0 +1,1320 @@
+// Autogenerated Vireo assembly file
+// Generated on 03/07/2019 15:38:14
+//---------------------------------------------------
+//VI Definition: ::Web Server::Interactive::WebApp::index.gviweb
```
sanmut: Thanks for adding extensive test cases. This is great. If the VIA files are generated from GVIs (and you are planning to check in the VIs in G test suite on ASW side), I would recommend keeping the names of the VIA files same as .GVI files (TestVI*.via).

test-it/ExpectedResults/Log10.vtr
```diff
@@ -0,0 +1,92 @@
...
+Log10(NaN) => NaN
+------ INT --------
+Log10(0) => -Inf
+Log10(1) => 0
+Log10(10) => 1
+Log10(1012) => 3.00518
+Log10(3) => 0.477121
+Log10(-1) => NaN
+Log10(2.14748E+09) => 9.33193
+------ CDB --------
```
sanmut: It is a good idea to do epsilon comparison for floating point than using expected values (to make sure they continue to pass in different platforms, compilers, so forth).

test-it/ViaTests/Log10.via
rkudukoli: I was waiting for the outcome of test location meeting with GregR before responding. Since TestVI prefix was not considered as a recommendation, I feel we should not adopt it. Also naming conventions are brittle.

Carlos mentioned creating nuget packages from NextGen containing via files for consumption by Vireo. I feel we should look more into it as an option.

test-it/ExpectedResults/Log10.vtr
rkudukoli: I agree. In this case, Printf seems to output the right number of digits under both native and JS and hence it works.  I will address it when I submit the NXG acceptance tests.

PR #583, opened 2019-03-14
Fixes to Variant Set,Get and Delete Attributes instructions

cglzaguilar:
The variant attribute map was holding pointers to the original data. Now they hold pointers to copies of the data.

Added more tests.


sanmut:
Let's create a folder called "generated" under ViaTests and move them under that folder, if we don't want to prefix the tests with "TestVI" anymore. It helps to distinguish handwritten tests from generated tests during development and debugging.

cglzaguilar:
Renamed test asset with TestVI_ prefix. We also need to update the test runner to also work with folders, which I would like to do in another PR in the future.

PR #584, opened 2019-03-18
Make functions const or static

sanmut:
Make C++ class methods const or static as appropriate.


rkudukoli:
Looks good to me. Thanks for doing it!

PR #585, opened 2019-03-19
Update emsdk version to sdk-1.38.30-64bit

rajsite:
This change updates the emscripten sdk version to sdk-1.38.30-64bit

Some benefits that have gone in are improved dead code elimination which takes our final output size from 882 KB to 825 KB.

The newer sdk version required a newer python version so used this as a chance to switch to the latest support Travis xenial image. This let us remove a bunch of workarounds needed for the older Precise image:
- No longer need to bring along a libc copy that the emsdk toolchain is built with
- No longer need to bring in a newer GCC, the xenial version (GCC 5.5) supports C++11 features

I did have to update the usages of ::isinf and isinf to be std::isinf because [cmath in newer GCC undefines the globals](https://stackoverflow.com/questions/39130040/cmath-hides-isnan-in-math-h-in-c14-c11).


rkudukoli:
Looks good. Thanks!

sanmut:
@cglzaguilar , @rajsite , let's hold on merging this until a private CI passes on ASW side and until some of the variant work for this iteration goes in just to reduce the churn.

rajsite:
@cglzaguilar had a passing Pre build: https://ni.visualstudio.com/DevCentral/_build/results?buildId=116656

And the lvjs build had one existing failing test and one unrelated failure (during webvi open when runtime isnt loaded): https://ni.visualstudio.com/DevCentral/_build/results?buildId=116698

Examples load and run fine, GitHub tests all good.

PR #587, opened 2019-03-19
Allow any data type to be wired into a VariantToData node, and enable coercion support for data types to be converted to Variants

siddhukrs:
This change enables two use cases:

1. `VariantToData` node must handle any data type wired to it, and not just Variants:
To enable this, I changed the instruction to accept `StaticTypeAndData`. If the type is not a Variant, I do a straight write through to the output type. However, if it is a Variant, I extract the inner type (e.g. Int32 if the input is a Variant of Int32). This inner data type is then used in IsA and Write operations that follow. I added tests to make sure this works for most scalar data types, 1D/2D/3D arrays of scalars and clusters.

2. It should be possible for any data type to be wired into a sub VI that accepts a Variant.
These are coercion nodes in NXG, and are translated to Convert instructions in VIA. Since the implementation for this must do exactly what the DataToVariant node does, I directly called `DataToVariant` to convert the input to a Variant. I added tests to verify that this works for a variety of scalar data types, 1D/2D/3D arrays of scalars and clusters.


source/core/Variants.cpp
```diff
@@ -55,14 +55,23 @@ VIREO_FUNCTION_SIGNATURET(VariantToData, VariantToDataParamBlock)
...
+        TypeRef inputType = _ParamImmediate(InputData._paramType);
+        void* inputData = _ParamImmediate(InputData)._pData;
+
         TypeRef destType = _ParamImmediate(DestData._paramType);
+        void* destData = _ParamImmediate(DestData)._pData;
 
-        if (variant->IsA(destType)) {
-            void* destData = _ParamImmediate(DestData)._pData;
-            variant->CopyData(variant->Begin(kPARead), destData);
+        if (inputType->Name().Compare(&TypeCommon::TypeVariant)) {
```
spathiwa: This is definitely an improvement but it doesn't handle compatible type conversion, which Variant to Data also does.
For example, if the input type is an Int32 and the output type is another numeric, it should do the conversion.
Also, if the output type is a variant and the input type isn't, that should work too, converting the input data to a variant (bizarrely, this makes the node act like Data to Variant, its own inverse).  This happens in the source model when the 'type' input is unwired, as it defaults to variant.

sanmut:
@siddhukrs , did you look into the CI failures?

test-it/testList.json
```diff
@@ -331,14 +331,18 @@
...
                 "TemplateBasics.via",
+                "TestVI_ArrayDataTypeCoerecedToVariant.via",
                 "TestVI_ArrayDelete.via",
                 "TestVI_ArrayDelete_1D_DifferentIndices.via",
                 "TestVI_ArrayDelete_2D_DifferentIndices.via",
                 "TestVI_ArrayInsertPartial.via",
                 "TestVI_ArrayNonFlatInflate.via",
                 "TestVI_ArrayOfRefs.via",
                 "TestVIArrayMaxAndMin.via",
+                "TestVI_ArraysWiredToVariantToData",
```
rajsite: Missing the via extension

test-it/testList.json
```diff
@@ -331,14 +331,18 @@
...
                 "TestVIConversion.via",
+                "TestVI_ClusterDataTypeCoercedToVariant.via",
+                "TestVI_ClustersWiredToVariantToData",
```
rajsite: Missing the via extension

test-it/testList.json
```diff
@@ -349,6 +353,8 @@
                 "TestVI_LargeArray.via",
                 "TestVI_LossyEnqueueElement.via",
                 "TestVI_MultipleDelimiters.via",
+                "TestVI_ScalarTypeCoercedToVariant.via",
+                "TestVI_ScalarsWiredToVariantToData",
```
rajsite: Missing the via extension

test-it/testList.json
siddhukrs: Thanks for catching this, and pushing a fix!

source/core/Variants.cpp
siddhukrs: We have a work item tracking the first case you mention. For the second one, I brought it up during standup today, and it sounded like something we did not want to carry forward on Web VIs -- the team leaned towards reporting an error if the DataType terminal is unwired. I will track the outcome of that discussion as a separate work item.

source/core/Variants.cpp
```diff
@@ -55,14 +55,23 @@ VIREO_FUNCTION_SIGNATURET(VariantToData, VariantToDataParamBlock)
...
+            TypeRef variantInnerType = *(TypeRef *)_ParamImmediate(InputData._pData);
```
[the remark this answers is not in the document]
cglzaguilar: @siddhukrs is out, I'll do this in my PR in the next set of changes.

source/core/Variants.cpp
```diff
@@ -55,14 +55,23 @@ VIREO_FUNCTION_SIGNATURET(VariantToData, VariantToDataParamBlock)
...
+            if (variantInnerType->IsA(destType)) {
+                variantInnerType->CopyData(variantInnerType->Begin(kPARead), destData);
+            } else if (errPtr) {
+                errPtr->SetErrorAndAppendCallChain(true, 91, "Variant To Data");
```
rkudukoli: Can we use error code consts instead of hardcoding the number directly?

rkudukoli: Same for "Variant To Data" string.

cglzaguilar: Definitely yes. Since Sid is out, I'll take care of this on my PR.

PR #590, opened 2019-03-25
Add support for keeping track of when dataitem elements are updated in Vireo and need refresh

spathiwa:
Note:  I changed the functions Milan had added to rename the concept of 'dirty' to be called 'update needed' instead.
In LabVIEW, the concept of a dirty flag already exists but has the opposite meaning, indicating that a control has been updated by the user and needs to be read by the execution system.  LabVIEW's has an updateNeeded flag which has the same meaning as this (execution has updated a control/indicator, which needs to be redisplayed), so I renamed Vireo's concept to match.
(We don't yet need the concept of a dirty flag in Vireo, but will when we are multi-threaded and we have to make sure value writes by the JS UI are atomic with respect to running Vireo exec threads.)


source/io/module_eggShell.js
```diff
@@ -250,6 +250,14 @@ var assignEggShell;
             return dispatchFunction(typeVisitor, valueRef, data);
         };
 
+        Module.eggShell.checkAndResetValueRefNeedsUpdateFlag = publicAPI.eggShell.checkAndResetValueRefNeedsUpdateFlag = function (valueRef) {
```
rajsite: I think we can simplify the name a bit. Maybe just `needsUpdate`.
in the context of usage:
```
var valueRef = vireo.eggShell.findValueRef('MyVI', 'dataItem_Boolean');
if(vireo.eggShell.needsUpdate(valueRef)) {
   ...
}
```

And that can be the only function on the public api. We can avoid assigning `checkValueRefNeedsUpdateFlag_SkipReset_` to the  `publicAPI.eggShell` object to make sure it's not used except for debugging. We could probably also simplify the name, maybe `needsUpdateSkipReset`?

We will still assign the debug function to the `Module.eggShell` object and the name helps makes sure it won't get used except for debugging and [maybe some tests](https://github.com/ni/VireoSDK/blob/0e94fa522f9b0564bf9713eb8386a229ec8c7ce1/test-it/karma/publicapi/ExecuteSlicesUntilClumpsFinshed.Test.js#L97) :)

spathiwa: I tried to leave checkValueRefNeedsUpdateFlag_SkipReset_ out of publicAPI.eggShell, but then I wasn't sure how to call it from the karma test. Can you provide guidance on that?

spathiwa: Ah, never mind.  I see the 'maybe some tests' link you just gave already provides an example.

spathiwa: I'm not a fan of long names either but I don't like what looks like a simple predicate function actually having a side effect of resetting the value without any indication that it does so.
How about something like:
`if(vireo.eggShell.needsUpdate_testAndReset(valueRef)) {
}
`

rajsite: Yea I agree about not hiding side-effects.

During the api refactor we tried to cleanup the api to be consistently camelCase and not mix formats.

What about `needsUpdateAndReset` or `needsUpdateAndResetFlag` or `needsUpdateAndResetDataItem`?

sanmut: TestAndSet, TestAndReset denote the atomic operation, so removing Test from the name will lose that clarity.

sanmut:
If dirty and update are treated separately and optimized differently (this aims at optimizing update), do we need to know whether an item is a control or indicator as opposed to just a data item? Data Item in NXG DFIR is a generic term for any terminal that is control or indicator. If some operation has to be done only for control or only for indicator, should that be distinguished in the flags/specifier for the local variable such as "d".

To make matters more complicated, NXG does not have local variables. A terminal can be duplicated in the diagram and direction of the duplicated terminal can be changed (so that a control can be written to and an indicator can be read from using the duplicate terminal). We use the same variable in VIA to support duplicate terminals as we don't need to anything special. I would like to understand what "d" means in those cases as well. Could one of you explain the scope of "d" and scope of handling "update" only and not "dirty" and handling "dirty" in the future?

source/io/module_eggShell.js
rajsite: My suggestion was before the latest updates to testAndResetNeedsUpdateFlag and testNeedsUpdateFlagWithoutReset. The newest updates saying they test and reset are good to me as well.

spathiwa:
I think 'd' is probably sufficient.   Since JS actively pokes values into Vireo's memory when a dataItem changes, we will not need the concept of a dirty flag per se, but when we are multithreaded we will need a mutex to denote when an input dataitem is being modified by JS, so Vireo does not try to read half-written data. Vireo could either block if it tries to read at the same time, or potentially enqueue something else if there are other clumps which could run.
Regarding "d" scope:  In the future, reads of locals declared with de() or inputs (parms declared i() or io()) will check the mutex before trying to read.  In this PR, writes of locals declared with de() or outputs (parma declared with o() or io()) set the needsUpdate flag so JS knows to read the value.

source/include/TypeAndDataManager.h
```diff
@@ -485,21 +485,23 @@ class TypeCommon
...
+    UInt16  _elementUsageType:3;  // (12-14) ElementType::UsageType
+    UInt16  _isDataItem:1;        // (15) Element keeps track of updatedNeeded state for read/writes
+    UInt16  _updateNeeded:1;      // (0) Value has been written, needs display (tested by JS)
     //  properties unique to DefaultPointerType objects
-    UInt16  _pointerType:3;       // (10-12)
-    UInt16  _ownsDefDefData:1;    // (13) Owns DefaultDefault data (clusters and arrays)
-    UInt16  _opaqueReference:1;   // (14) Data is not an instance of the type it wraps or is templated from (e.g. refnum(Queue))
+    UInt16  _pointerType:3;       // (1-3)
+    UInt16  _ownsDefDefData:1;    // (4) Owns DefaultDefault data (clusters and arrays)
+    UInt16  _opaqueReference:1;   // (5) Data is not an instance of the type it wraps or is templated from (e.g. refnum(Queue))
```
sanmut: Should we make it a UInt32 for the bitfields?

source/core/TypeAndDataManager.cpp
```diff
@@ -704,13 +708,17 @@ NIError TypeCommon::CopyData(const void* pSource, void* pDest, IntIndex count)
...
+    if (_isDataItem)
+        SetValueRefNeedsUpdate(this, pDest);
     return err;
 }
 //------------------------------------------------------------
 NIError TypeCommon::MultiCopyData(const void* pSource, void* pDest, IntIndex count)
 {
     if (IsFlat() && TopAQSize() == 1) {
         memset(pDest, (int)*(AQBlock1*)pSource, count);
+        if (_isDataItem)
```
sanmut: Should we do this in the IsFlat() case? Let's discuss.

source/core/TypeAndDataManager.cpp
```diff
@@ -1776,6 +1785,8 @@ NIError ArrayType::CopyData(const void* pData, void* pDataCopy)
             pDestElt += stride;
         }
     }
+    if (_isDataItem)
```
sanmut: Should we also do this for "ClusterType::CopyData"?

source/core/TDCodecVia.cpp
```diff
@@ -500,23 +500,30 @@ void TDViaParser::ParseAggregateElementList(std::vector<TypeRef> *elementTypesVe
...
+        } else if (token.CompareCStr(tsDataitemElementToken)) {
+            usageType = kUsageTypeSimple;
+            isDataItem = true;
         } else if (token.CompareCStr(tsConstElementToken)) {
             usageType = kUsageTypeConst;
         } else if (token.CompareCStr(tsInputParamToken)) {
             usageType = kUsageTypeInput;
         } else if (token.CompareCStr(tsOutputParamToken)) {
             usageType = kUsageTypeOutput;
+            isDataItem = true;
```
sanmut: Let's check for "d" in Params block also to mark isDataItem true for completeness (as we do for data Items in local variables). It is a good idea to make DFIR generate VIA to indicate which variables are data items regardless of which block it belongs to (Locals or Params). In the future, there are potentially other attributes/meta-data DFIR may want to put in VIA for variables or instructions that Vireo can make use of. It is better to parse that directly in Vireo. I made the same comment in @ggumen 's DFIR side PR too. We all can discuss in person if we want to.

spathiwa: That doesn't make sense, because Param blocks can only contain inputs and outputs.  Otherwise they are not 'parameters'.

source/core/TypeAndDataManager.cpp
```diff
@@ -1180,20 +1188,21 @@ TypeRef WrappedType::GetSubElementAddressFromPath(SubString* name, void *start,
 //------------------------------------------------------------
 // ElementType
 //------------------------------------------------------------
-ElementType* ElementType::New(TypeManagerRef typeManager, SubString* name, TypeRef wrappedType, UsageTypeEnum usageType, Int32 offset)
+ElementType* ElementType::New(TypeManagerRef typeManager, SubString* name, TypeRef wrappedType, UsageTypeEnum usageType, Int32 offset, bool isDataItem)
 {
-    ElementType* type = TADM_NEW_PLACEMENT_DYNAMIC(ElementType, name)(typeManager, name, wrappedType, usageType, offset);
-
+    ElementType* type = TADM_NEW_PLACEMENT_DYNAMIC(ElementType, name)(typeManager, name, wrappedType, usageType, offset, isDataItem);
+    if (isDataItem)
```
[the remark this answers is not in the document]
spathiwa: > Why do we early return from here if it is data item?
Because we don't want the types unique'd for dataItems, since we need to track the updateNeeded flag independently.

> In VIA generated from DFIR , for aggregate data types such as clusters, only the variable itself will be marked "d" (data item) and not every field in the aggregate data type.
This is fine, because when you write to a terminal in G, you always update the entire thing, which will generate a Copy instruction to the entire cluster.

source/include/TypeAndDataManager.h
spathiwa: I wouldn't be opposed, but it doesn't really matter, I think the compiler will generate the same code.

source/core/TypeAndDataManager.cpp
spathiwa: I'm actually removing these checks, because they are not needed and only work if we are called through the destination type.
The actual code path which implements the needed check is in the custom emit proc for the Copy instruction.

source/core/TypeAndDataManager.cpp
spathiwa: See previous comment.

source/core/TDCodecVia.cpp
spathiwa: Also, the reason inputs are not auto-marked dataItems is because you cannot write to an input in the first place; a Copy to it would generate an error.
My hunch is that if you duplicate an input terminal and write to it, it because 'io' in DFIR for Vireo. in which case it would work.  Ask Gandi about this.

source/core/GenericFunctions.cpp
```diff
@@ -117,6 +117,13 @@ InstructionCore* EmitGenericCopyInstruction(ClumpParseState* pInstructionBuilder
         SubString copyOpToken(copyOpName);
         pInstructionBuilder->ReresolveInstruction(&copyOpToken);
         pInstruction = pInstructionBuilder->EmitInstruction();
+        if (destType->IsDataItem()) {
```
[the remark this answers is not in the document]
spathiwa: This was explained in the PR's description.
We always generate a Copy instruction to write to dataItems in DFIR, so we don't need to handle all possible instructions (yet).
In the future, if we implement better inplaceness/optimization, we may need to address this.

source/core/TDCodecVia.cpp
```diff
@@ -500,23 +500,30 @@ void TDViaParser::ParseAggregateElementList(std::vector<TypeRef> *elementTypesVe
...
+        isDataItem = false;
         if (token.CompareCStr(tsElementToken)) {
             usageType = kUsageTypeSimple;
+        } else if (token.CompareCStr(tsDataitemElementToken)) {
+            usageType = kUsageTypeSimple;
+            isDataItem = true;
         } else if (token.CompareCStr(tsConstElementToken)) {
             usageType = kUsageTypeConst;
         } else if (token.CompareCStr(tsInputParamToken)) {
             usageType = kUsageTypeInput;
```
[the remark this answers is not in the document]
spathiwa: I answered this in another comment, but I believe such terminals would be marked 'io', not 'i'.
If they are 'i', they cannot be written to by Vireo; it's an error, detected at parse time, not run-time, to do so.

source/core/ExecutionContext.cpp
```diff
@@ -445,6 +492,8 @@ void ExecutionContext::IsrEnqueue(QueueElt* elt)
 DEFINE_VIREO_BEGIN(Execution)
     DEFINE_VIREO_REQUIRE(VirtualInstrument)
     DEFINE_VIREO_FUNCTION(FPSync, "p(i(String))")
+    DEFINE_VIREO_FUNCTION(SetValueNeedsUpdate, "p(i(StaticTypeAndData dirtyValue))")
+    DEFINE_VIREO_FUNCTION(CheckValueNeedsUpdate, "p(i(StaticTypeAndData dirtyValue) o(Boolean))")
```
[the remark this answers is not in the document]
spathiwa: Done.

source/core/ExecutionContext.cpp
```diff
@@ -140,6 +140,53 @@ VIREO_FUNCTION_SIGNATURE1(FPSync, StringRef)
...
+{
+    _ParamDef(StaticType, valueType);
+    _ParamDef(void, valueData);
+    _ParamDef(Boolean, needsUpdate);
+    NEXT_INSTRUCTION_METHOD()
+};
+VIREO_FUNCTION_SIGNATURET(CheckValueNeedsUpdate, CheckValueNeedsUpdateParamBlock)
+{
+    VirtualInstrument* vi = THREAD_EXEC()->_runningQueueElt->OwningVI();
+    if (!vi->Clumps()->Begin()->_caller) {
```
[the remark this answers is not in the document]
spathiwa: Done

source/core/ExecutionContext.cpp
```diff
@@ -140,6 +140,53 @@ VIREO_FUNCTION_SIGNATURE1(FPSync, StringRef)
...
+Boolean TestValueRefNeedsUpdate(TypeRef typeRef, const void *, Boolean reset) {
+    Boolean isDirty = typeRef->IsUpdateNeeded();
+    if (isDirty && reset)
+        typeRef->SetUpdateNeeded(false);
+    return isDirty;
+}
+
+struct CheckValueNeedsUpdateParamBlock : InstructionCore
+{
+    _ParamDef(StaticType, valueType);
```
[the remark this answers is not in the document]
spathiwa: Done.

source/core/ExecutionContext.cpp
```diff
@@ -140,6 +140,53 @@ VIREO_FUNCTION_SIGNATURE1(FPSync, StringRef)
...
+VIREO_FUNCTION_SIGNATURET(SetValueNeedsUpdate, SetValueNeedsUpdateParamBlock)
+{
+    VirtualInstrument* vi = THREAD_EXEC()->_runningQueueElt->OwningVI();
+    if (!vi->Clumps()->Begin()->_caller)
+        SetValueRefNeedsUpdate(_ParamPointer(valueType), _ParamPointer(valueData));
+    return _NextInstruction();
+}
+
+Boolean TestValueRefNeedsUpdate(TypeRef typeRef, const void *, Boolean reset) {
+    Boolean isDirty = typeRef->IsUpdateNeeded();
```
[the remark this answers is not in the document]
spathiwa: Done.

source/core/ExecutionContext.cpp
```diff
@@ -140,6 +140,53 @@ VIREO_FUNCTION_SIGNATURE1(FPSync, StringRef)
...
+{
+    _ParamDef(StaticType, valueType);
+    _ParamDef(void, valueData);
+    NEXT_INSTRUCTION_METHOD()
+};
+
+VIREO_FUNCTION_SIGNATURET(SetValueNeedsUpdate, SetValueNeedsUpdateParamBlock)
+{
+    VirtualInstrument* vi = THREAD_EXEC()->_runningQueueElt->OwningVI();
+    if (!vi->Clumps()->Begin()->_caller)
```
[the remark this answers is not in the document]
spathiwa: Done.

source/core/ExecutionContext.cpp
```diff
@@ -140,6 +140,53 @@ VIREO_FUNCTION_SIGNATURE1(FPSync, StringRef)
...
 }
+//------------------------------------------------------------
+//
+void SetValueRefNeedsUpdate(TypeRef typeRef, void *dataPtr) {
+    typeRef->SetUpdateNeeded(true);
+}
+
+struct SetValueNeedsUpdateParamBlock : InstructionCore
+{
+    _ParamDef(StaticType, valueType);
```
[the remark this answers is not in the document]
spathiwa: Done.

source/include/TDCodecVia.h
```diff
@@ -239,10 +239,14 @@ void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], Strin
 #define tsEnqueueTypeToken      "enqueue"
 #define tsElementToken          "e"   // used for Cluster, BitCluster, and array aggregate types for simple elements
 #define tsConstElementToken     "ce"  // used for Cluster elements in Locals: section to indicate immutable value
-#define tsAliasToken            "al"  // alias to another element.
+#define tsDataitemElementToken  "de"  // used for Cluster elements in Locals: section to indicate dataItem value
+                                      // (value will be marked dirty when written from Vireo,
```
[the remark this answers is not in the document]
spathiwa: Done.

spathiwa:
Addressed feedback, answered questions about Copy/input dataItem usage.

source/core/TDCodecVia.cpp
ggumen: Currently, we do not emit 'io' in VIA from DFIR yet. Sankara tried to put that change in last release, but had to revert it before we shipped. 

To me the naming of this variable `isDataItem` is confusing. DataItems in LabVIEW can be either Control or Indicator, and whether it's a Control or an Indicator, it can always be read and written to through its multiple DataAccessors. 
The concept of vireo `io` doesn't match LabVIEW's multiple DataAccessors. We should change Vireo to accept copying to `i` DataItems from its DataAccessor local. I think that's why Sankara had to revert his change.

Maybe we want to call this variable `isIndicatorDataItem`?

spathiwa: So I infer from this that having a write DataAccessor for a dataItem that is an input on the conpane will currently NOT work in Vireo (already, regardless of any changes made in this PR)?

spathiwa: It feels fundamentally wrong to me for Vireo to allow writing to inputs.  If it's written to, it should be 'io', and I don't understand why this doesn't match multiple data accessors.  The use-case seems the same to me.

sanmut: We discussed this in person. "io" should be used only in the cases where a callee's modification should reflect in the caller. We need to support in NXG (for Web target) allowing writing to duplicate terminal of control in the connector pane in the future and probably not urgent. I think the notion of "i", "o", "io" with respect to the contract between callee/caller should be distinguished from the notion of mutability of the same in the SubVI. 

DFIR marking only items in local as data item and not the items on the param block and then Vireo not marking only input data items in Param block as not a data item is inconsistent and not clear.

spathiwa: I just updated so inputs are also considered dataItems by Vireo, even though they don't need to check for updateNeeded, for consistency.

sanmut:
@cglzaguilar , have you had a chance to review this yet?

cglzaguilar:
> @cglzaguilar , have you had a chance to review this yet?

I'm about to do it.

source/core/CEntryPoints.cpp
```diff
@@ -184,6 +184,18 @@ VIREO_EXPORT EggShellResult EggShell_FindSubValue(TypeManagerRef tm,
     return kEggShellResult_Success;
 }
 //------------------------------------------------------------
+//! Check if a value has been written by Vireo (and reset the state)
+VIREO_EXPORT Boolean EggShell_TestAndResetNeedsUpdateFlag(TypeManagerRef tm, const TypeRef typeRef, const void* pData)
```
cglzaguilar: Please change the name of the function to something like:
EggShell_IsUpdateNeededAndReset and the next one to
EggShell_IsUpdateNeeded

source/core/ExecutionContext.cpp
```diff
@@ -140,6 +140,53 @@ VIREO_FUNCTION_SIGNATURE1(FPSync, StringRef)
...
+}
+
+struct CheckValueNeedsUpdateParamBlock : InstructionCore
+{
+    _ParamDef(StaticType, ValueType);
+    _ParamDef(void, ValueData);
+    _ParamDef(Boolean, NeedsUpdate);
+    NEXT_INSTRUCTION_METHOD()
+};
+VIREO_FUNCTION_SIGNATURET(CheckValueNeedsUpdate, CheckValueNeedsUpdateParamBlock)
```
cglzaguilar: Update the name of the function to CheckValueNeedsUpdateAndReset. This function has the side effect of resetting the NeedsUpdate flag.

source/include/TDCodecVia.h
```diff
@@ -239,10 +239,14 @@ void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], Strin
 #define tsEnqueueTypeToken      "enqueue"
 #define tsElementToken          "e"   // used for Cluster, BitCluster, and array aggregate types for simple elements
 #define tsConstElementToken     "ce"  // used for Cluster elements in Locals: section to indicate immutable value
-#define tsAliasToken            "al"  // alias to another element.
+#define tsDataitemElementToken  "de"  // used for Cluster elements in Locals: section to indicate dataItem value
```
cglzaguilar: Please rename this constant to tsDataItemElementToken

source/core/CEntryPoints.cpp
spathiwa: We already had a discussion about this, and landed on this name because 'TestAndReset' is a well-defined term of art describing this behavior.  Your proposed name is awkward, combining a question predicate with a verb.
Do you have a suggestion for an improved name that retains the 'TestAndReset' phrase?

cglzaguilar: Ok, let's keep this name.

rajsite:
Unfortunately this work has been deprioritized. Closing until can be revisited.

PR #591, opened 2019-03-26
Fix errors in type comparison in variant nodes involving arrays and clusters

siddhukrs:
1. All type comparisons in Variants code was using the IsA(destinationType) check. This check relies on names, and breaks down when the types are clusters. In this change, I am changing them to use IsA(destinationType, compatibleStructure), which additionally compares inner types and ranks in clusters and arrays. It is generally a more stricter check.

2. The rank checking code in case of arrays had a bug where it was allowing arrays of different ranks to pass through, eventually comparing them by name, eventually claiming arrays of different ranks were equal. I am fixing this issue, too.

3. A special case that we do not support currently is when an cluster is wired in to the VariantToData node expecting an array. Though unsupported today, this is a valid use-case and should not report back the usual error 91. In this change, I am changing the error code we return in this case to 1558, which is the closest existing error code we could find. In the future, we will remove this code and actually implement this functionality. This is just so we do not confuse users with varying behavior between desktop and web vis in the meantime.


test-it/ViaTests/TestVI_ClusterWiredToVariantToDataExpectingAnArray_Reports1558Error.via
```diff
@@ -0,0 +1,47 @@
+// Autogenerated Vireo assembly file
+// Generated on 03/26/2019 16:59:09
```
siddhukrs: TestVI_ClusterWiredToVariantToDataExpectingAnArray_Reports1558Error

test-it/ViaTests/TestVI_ClusterWiredToVariantToDataExpectingAnArray_Reports1558Error.via
```diff
@@ -0,0 +1,47 @@
...
+//---------------------------------------------------
+//VI Definition: ::Web Server::Interactive::Application::TestVI_ClusterWiredToVariantToDataExpectingAnArray_Reports1558Error.gviweb
```
siddhukrs: TestVI_ClusterWiredToVariantToDataExpectingAnArray_Reports1558Error

test-it/ViaTests/TestVI_GetVariantAttribute_IncompatibleClusterDataTypeReportsError.via
```diff
@@ -0,0 +1,85 @@
+// Autogenerated Vireo assembly file
```
siddhukrs: This VI verifies that an error with the right error code is reported on VIs when incompatible data types are wired to the VariantToData node.
The following cases are tested:

1. Variant of Int32 vs String
2. Variant of Int32 vs Cluster of Int32
3. Variant of cluster of Int32 vs Int32
4. Variant of cluster of scalars vs cluster with same scalar data types but in different order
5. Variant of string vs Path
6. Variant of 3D array of Int32 vs 3D array of strings
7. Variant constant vs int32
8. Variant of 2D array of Int32 vs 1D array of Int32
9. Existing errors on the error wire are retained in case of no new errors
10. Unwired error input does not report error
11. Unwired data output does not report error

test-it/ViaTests/TestVI_IncompatibleScalarDataTypes_VariantToData.via
```diff
@@ -0,0 +1,323 @@
+// Autogenerated Vireo assembly file
+// Generated on 03/26/2019 16:47:20
+//---------------------------------------------------
```
siddhukrs: This VI verifies that an error with the right error code is reported on VIs when incompatible data types are wired to the VariantToData node.
The following cases are tested:

1. Variant of Int32 vs String
2. Variant of Int32 vs Cluster of Int32
3. Variant of cluster of Int32 vs Int32
4. Variant of cluster of scalars vs cluster with same scalar data types but in different order
5. Variant of string vs Path
6. Variant of 3D array of Int32 vs 3D array of strings
7. Variant constant vs int32
8. Variant of 2D array of Int32 vs 1D array of Int32
9. Existing errors on the error wire are retained in case of no new errors
10. Unwired error input does not report error
11. Unwired data output does not report error

source/core/Variants.cpp
```diff
@@ -162,14 +171,13 @@ VIREO_FUNCTION_SIGNATURET(GetVariantAttribute, GetVariantAttributeParamBlock)
             auto attributeMapIter = attributeMap->find(name);
             if (attributeMapIter != attributeMap->end()) {
                 TypeRef foundValue = attributeMapIter->second;
+                found = true;
```
siddhukrs: Moved this out here because the found terminal should return true when it finds an attribute, regardless of whether it matches the default data's type.

test-it/ViaTests/TestVI_GetVariantAttribute_IncompatibleClusterDataTypeReportsError.via
```diff
@@ -0,0 +1,85 @@
...
+// Generated on 03/26/2019 16:39:18
+//---------------------------------------------------
+//VI Definition: ::Web Server::Interactive::Application::TestVI_GetVariantAttribute_IncompatibleClusterDataTypeReportsError.gviweb
```
siddhukrs: TestVI_GetVariantAttribute_IncompatibleClusterDataTypeReportsError

source/core/Variants.cpp
siddhukrs: Undoing this change. I will look into it in a separate PR.

source/core/TypeAndDataManager.cpp
```diff
@@ -739,8 +739,11 @@ Boolean TypeCommon::CompareType(TypeRef otherType)
     if (this == otherType) {
         return true;
     } else if (thisEncoding == kEncoding_Array && otherEncoding == kEncoding_Array) {
-        if (this->Rank() == otherType->Rank())
+        if (this->Rank() != otherType->Rank()) {
```
spathiwa: Wow, good catch.  This was a pretty bad bug.

test-it/ViaTests/TestVI_VariantToData_VariantWiredToDataTypeTerminal.via
```diff
@@ -0,0 +1,65 @@
+// Autogenerated Vireo assembly file
+// Generated on 03/26/2019 19:20:54
+//---------------------------------------------------
+//VI Definition: ::Web Server::Interactive::Application::TestVI_VariantToData_VariantWiredToDataTypeTerminal.gviweb
+define (Application%3A%3ATestVI_VariantToData_VariantWiredToDataTypeTerminal%2Egviweb dv(.VirtualInstrument (
```
siddhukrs: TestVI_VariantToData_VariantWiredToDataTypeTerminal

source/include/Variants.h
```diff
@@ -60,7 +60,7 @@ class VariantAttributeManager {
     }
 };
 
-enum { kVariantArgErr = 1, kVariantIncompatibleType = 91};
+enum { kVariantArgErr = 1, kVariantIncompatibleType = 91, kVariantTypeMismatch = 1558 };
```
spathiwa: What situations is error 1558 normally used for (since this is a stopgap case)?

cglzaguilar: @spathiwa, good question. This is the closest thing we were able to find to indicate unsupported behavior. What we basically need is an error code indicating that some behavior is not supported in the target. Are you aware of an error code we can use instead?

spathiwa: The error usually used for functionality not supported on a particular target (for example, desktop features not supported in RT, or compiler or scripting property nodes in a built app), is mgNotSupported, error code 53.
I think this would be a good choice here.

siddhukrs: Options:
1. 1558 - There is a type mismatch between the wire type and the internal data type of the reference. This problem might occur due to a type cast error.
2. 1057 - Type mismatch: Object cannot be cast to the specified type.
3. 2304 - The command is not supported on the target.
4. 1429 - Command not supported by Target (sigh, why do we need this AND the previous one!)
5. 53 - Manager call not supported.

@spathiwa The unsupported on current target would be the best fit, but is 53 the right code for the error for that? Explain error doesn't seem to match that description. 

Looking at the descriptions above, 1057 seems generic enough for this case. The others have too much detail that doesn't match what is actually going on.

siddhukrs: Talked to @cglzaguilar and we agreed that 2304 suits this case the best. Let me know if you disagree. I am going to update my tests and code with this error code otherwise.

spathiwa: Actually, I think that one is currently only used for RT targets, so it's a little weird to use it here.

spathiwa: Maybe we should consider a 'not implemented' error code, rather than 'not supported', since this is a temporary failing that will be addressed in the future.
There is Error 1072, 'This property or method is not implemented', but that's unfortunately a little too specific, since this isn't for a property or method.
Other existing errors expressing the concept of not implemented are either not in the LabVIEW core range (they are negative or large hex values), or they are specific to an add-on product (16502, for SMTP part of internet toolkit).

Maybe you should consider adding a new error code?
https://niweb.natinst.com/confluence/display/RDITAppsHelp/Using+the+NI+Error+Codes+db

siddhukrs: I would've added a new code if not for this PR blocking some committed work that is due this week. I am also hesitant because we will likely implement the actual feature and remove the error before ever releasing.

Although 2304 is currently only used by RT targets, there's nothing about its messaging that prevents us from using it, right? It is the target that is preventing this case from working (it works on Desktop and in CG).

spathiwa: Ok. If it’s short/term, I’m fine with 2304.

source/core/TypeAndDataManager.cpp
siddhukrs: For better history tracking, I am moving this change to a separate PR. I will rebase this PR on top of master once the other one is submitted.

test-it/ViaTests/TestVI_GetVariantAttribute_IncompatibleClusterDataTypeReportsError.via
siddhukrs: Updated this second cluster (Cluster(Bool+Double)) to instead be a Cluster(Bool). That way, the two clusters are not coercible and will fail on desktop too.

cglzaguilar:
@sanmut, It is my understanding that after looking at the code changes with @siddhukrs, there is not more feedback. I'm merging this as soon as tests pass to generate a new Vireo release. If you have further comments we can attend them in another PR.

PR #593, opened 2019-03-27
Fix bug in the IsA checking code

siddhukrs:
Bug in the IsA check allowed Arrays of different ranks to fall through to a name-based comparison, incorrectly reporting that they are the same. I'm guessing this was likely not too big of an issue pre-variants, because compile time checks would've protected us from hitting it.

This change was reviewed in [PR](https://github.com/ni/VireoSDK/pull/591/files) initially, moving it to a separate PR for history tracking purposes on @sanmut's request.


cglzaguilar:
This has already been reviewed by @sanmut and @spathiwa in another PR.

PR #594, opened 2019-03-27
Fix IsA type comparison for arrays

siddhukrs:
This is a change that I missed in my previous submission.


cglzaguilar:
Trivial fix to the previous PR.

PR #595, opened 2019-04-03
Consume type and data terminals separately in via for VariantToData

siddhukrs:
The VariantToData node was previously passing the target type terminal as a passthrough into the data out terminal. To enable additional work coming up and to better reflect the node's representation in DFIR, I am splitting the two terminals in emitted via.

In this change, I am separating out the type input terminal from the data output terminal. I decide if the input types are compatible using the type terminal, and copy data into the data terminal.

To update via tests, I regenerated the via wherever the original source was available. In case of handwritten via files, I manually updated them.


test-it/ExpectedResults/TestVI_VariantToData_VariantWiredToDataTypeTerminal.vtr
```diff
@@ -1,2 +1,3 @@
 TestResult_Int32InAndUnwiredDataTypeInDoesNotReportError true
 TestResult_VariantOfInt32InAndUnwiredDataTypeInDoesNotReportError true
+TestResult_VariantOfInt32InWithUnwiredDataTypeInAndDataOutDoesNotReportsError true
```
siddhukrs: Added a new test case here to verify that if the type _and_ data terminals are unwired, we do not crash or report error.

spathiwa:
I noticed this descrepancy earlier when reviewing another variant PR, and I assumed that we just emitted an obligatory Copy() of the type input in DFIR to protect it from being overwritten.
Did you consider addressing the issue that way? It would be a lot simpler fix, but it might be better to have the Vireo API more directly parallel the G prim anyway.

source/core/Variants.cpp
```diff
@@ -75,23 +76,32 @@ VIREO_FUNCTION_SIGNATURET(VariantToData, VariantToDataParamBlock)
         TypeRef inputType = _ParamImmediate(InputData._paramType);
         void* inputData = _ParamImmediate(InputData)._pData;
 
-        TypeRef destType = _ParamImmediate(DestData._paramType);
+        TypeRef destType = _ParamImmediate(DestType._paramType);
+
```
spathiwa: You now need to verify that destData's actual type is compatible with the wired destType.
The checks below compare variantInnerType and destType, which no longer is guaranteed to be the same as destData's actual type. A mismatch would probably crash.

source/core/Variants.cpp
```diff
@@ -75,23 +76,32 @@ VIREO_FUNCTION_SIGNATURET(VariantToData, VariantToDataParamBlock)
...
+                if (destData) {
+                    variantInnerType->CopyData(variantInnerType->Begin(kPARead), destData);
+                }
             } else if (variantInnerType && destType->IsVariant()) {
-                *static_cast<TypeRef*>(destData) = CopyToVariant(variantInnerType);
+                if (destData) {
+                    *static_cast<TypeRef*>(destData) = CopyToVariant(variantInnerType);
+                }
             } else if (errPtr) {
                 SetVariantToDataTypeError(variantInnerType, destType, errPtr);
```
sanmut: We should set the destination data to default value for the type and add tests.

sanmut:
> I noticed this discrepancy earlier when reviewing another variant PR, and I assumed that we just emitted an obligatory Copy() of the type input in DFIR to protect it from being overwritten.
> Did you consider addressing the issue that way? It would be a lot simpler fix, but it might be better to have the Vireo API more directly parallel the G prim anyway.

It is actually not a functional issue. Emitted VIA correctly introduces a copy. There was a glitch in the Vireo HTML panel where it was updating the value incorrectly for any control (not just the one wired to this node). @siddhukrs , I would update the PR description to remove this as it is misleading.

source/core/Variants.cpp
siddhukrs: Since this isn't directly related to what this PR intentions were, I will address this in a subsequent PR. I also created a work item to track this.

source/core/Variants.cpp
siddhukrs: I went ahead and made these changes, though we're protected by semantic analysis when coming from LabVIEW. I am yet to add tests, so hold off on reviewing my recent push. I will resolve this comment when I am done.

source/core/Variants.cpp
```diff
@@ -75,25 +87,42 @@ VIREO_FUNCTION_SIGNATURET(VariantToData, VariantToDataParamBlock)
...
+        void* outputData = _ParamImmediate(OutputData)._pData;
 
         if (inputType->IsVariant()) {
             TypeRef variantInnerType = *reinterpret_cast<TypeRef *>_ParamImmediate(InputData._pData);
-            if (variantInnerType && variantInnerType->IsA(destType, true)) {
-                variantInnerType->CopyData(variantInnerType->Begin(kPARead), destData);
-            } else if (variantInnerType && destType->IsVariant()) {
-                *static_cast<TypeRef*>(destData) = CopyToVariant(variantInnerType);
+            if (variantInnerType
+                && IsTypeCompatibleWithInput(variantInnerType, targetType, targetData)
```
sanmut: 1. targetData should not be used but only targetType should be used, because this corresponds to a type only input
2. It is safe to assume TargetType cannot be null and it is a required argument to this API. In NXG, this is going to be fixed to be a required terminal to match the CG behavior. So, in DFIR, if it is unwired, we can generate and pass a variant constant for now and assume here this will never be null.
3. targetType and outputType should be compared separately that they are identical. If they are not identical, we should bail out early. This is just an extra sanity check that DFIR already guarantees and we just want to ensure the same guarantee.
4. We should keep comparing variantInnerType against targetType separate, as we will be making it more liberal for coercible types. If we mix up these, then it seems to cause confusion we may end up with more restrictive or less restrictive than needed with 3-way type checking

source/core/Variants.cpp
```diff
@@ -67,6 +68,17 @@ TypeRef CopyToVariant(TypeRef sourceType)
     return variant;
 }
 
+bool IsTypeCompatibleWithInput(TypeRef inputType, TypeRef destType, void* destData)
+{
+    // If type terminal is unwired, or it is the same as the input, return true.
+    return inputType && (!destData || inputType->IsA(destType, true));
+}
+
+bool IsTerminalUnwiredOrVariant(TypeRef destType, void* destData)
```
sanmut: It is better to name functions and explain comments in VIA parlance first whenever we can (and use G terminology in addition to that in comments or instead of when necessary). In stead of "terminal unwired", it might be better to say "if optional argument is null or not specified or something to that effect"

source/core/Variants.cpp
sanmut: It is also probably fine to make the additional check (targetType and outputType) under a VIREO_ASSERT. As the test suite uses debug builds, it will catch any violations. This is a good safety net in Vireo for a very good guarantee we already receive from DFIR emitted VIA that ensure these two types are identical.

siddhukrs: 1, 2. I am only passing in targetData to make it consistent with passing outputType and outputData. Having two varied APIs to compare input type with target vs output makes it harder to maintain. Additionally, it is possible for the targetType to be unavailable when it is unwired, and this can only be detected through targetData being null. (In this case, it means the targetType is a variant). While I can see how this can be avoided with better via emission, (which I in fact am) I think Vireo should be protected from bad inputs regardless. If we use the argument that "NXG doesn't allow this case", none of the changes in these new commits about outputType comparisons are necessary because SA protects us against them.
3. This is achieved by comparing target to input, and then output to input. (in the line below). I can add the additional VIREO_ASSERT that your next comment mentions.
4. Since what I have is correct for now, I would wait and make that change when we get to it. (I also think that the coercible type check will eventually apply for both the type comparisons.)


I vote to keep the comparison APIs simple. I say this because I did have multiple functions doing this at some point and it was hard for me to keep track even though I wrote them myself. If the complexity is necessary, we should do it whenever there's a need for it.

source/core/Variants.cpp
siddhukrs: Yeah, I was conflicted on my use of "terminal". I will change this.

test-it/ViaTests/TestVI_VariantSetGetAttributes.via
```diff
@@ -344,153 +425,203 @@ define (Application%3A%3AVariant_SetGetAttributes_ExpectedResults%2Egviweb dv(.V
...
-        Copy(local186 dataItem_TestResult_GetAttributeWorksCorrectly)
-		Printf("dataItem_TestResult_AttributeDoesNotExist = %z\n" dataItem_TestResult_AttributeDoesNotExist)
+        IsEQ(local253 c121 local254)
+        IsEQ(c122 local252 local255)
+        IsEQ(local251 c123 local256)
+        And(local255 local256 local257)
+        And(local254 local257 local257)
+        Copy(local257 dataItem_TestResult_GetAttributeWorksCorrectly)
+        Printf("dataItem_TestResult_AttributeDoesNotExist = %z\n" dataItem_TestResult_AttributeDoesNotExist)
 		Printf("dataItem_TestResult_EmptyAttributeNameReturnsAnErrorOn = %z\n" dataItem_TestResult_EmptyAttributeNameReturnsAnErrorOn)
```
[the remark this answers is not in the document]
siddhukrs: Done.

source/core/Variants.cpp
siddhukrs: Talked to Sankara IRL. Updated the code to:
1. check that the inputType is compatible with targetType (in the future, this function will verify that they are coercible)
2. check that the inputType is the same as the outputType (aka, they satisfy isA, or that the outputType is unwired, or both are unwired, or both are variants)

source/core/Variants.cpp
```diff
@@ -75,25 +94,42 @@ VIREO_FUNCTION_SIGNATURET(VariantToData, VariantToDataParamBlock)
...
+        TypeRef outputType = _ParamImmediate(OutputData._paramType);
+        void* outputData = _ParamImmediate(OutputData)._pData;
 
         if (inputType->IsVariant()) {
             TypeRef variantInnerType = *reinterpret_cast<TypeRef *>_ParamImmediate(InputData._pData);
-            if (variantInnerType && variantInnerType->IsA(destType, true)) {
-                variantInnerType->CopyData(variantInnerType->Begin(kPARead), destData);
-            } else if (variantInnerType && destType->IsVariant()) {
-                *static_cast<TypeRef*>(destData) = CopyToVariant(variantInnerType);
+            if (variantInnerType
```
cglzaguilar: Excellent using functions here. It facilitates reading the code.

PR #596, opened 2019-04-05
Write default value of type terminal to output in erroneous cases in VariantToData

siddhukrs:
When an error occurs in `VariantToData` due to incompatible data types, the default value of the type wired to the type terminal must be written to the output. 

To facilitate this, I am updating via emission to strip out the value wired to the type terminal in `VariantToData` and replace it with an uninitialized local. That way, whatever value is found in the type terminal's data can safely assumed to be the default value for that type and written to the output parameter, in case of an error.

Updates to via emission can be seen [here](https://ni.visualstudio.com/DevCentral/_git/ASW/pullrequest/23056?_a=overview).


test-it/ViaTests/TestVI_VariantToData_DefaultValueOfTypeTerminalIsWrittenIntoOutputInCaseOfError.via
```diff
@@ -0,0 +1,244 @@
+// Autogenerated Vireo assembly file
+// Generated on 04/05/2019 12:40:43
+//---------------------------------------------------
+//VI Definition: ::Web Server::Interactive::Application::TestVI_VariantToData_DefaultValueOfTypeTerminalIsWrittenIntoOutputInCaseOfError.gviweb
```
siddhukrs: case1
case2

siddhukrs: The last string tunnel on the right is a keep-last-value.

spathiwa:
Why don't you just use InitData instead of CopyData?

source/core/Variants.cpp
```diff
@@ -86,6 +77,21 @@ bool IsTargetTypeCompatibleWithInput(TypeRef inputType, TypeRef targetType, void
...
+    if (inputType && inputType->IsCluster() && targetType->IsArray()) {
+        errPtr->SetErrorAndAppendCallChain(true, kUnsupportedOnTarget, "Variant To Data");
+    } else {
+        errPtr->SetErrorAndAppendCallChain(true, kVariantIncompatibleType, "Variant To Data");
+    }
+    if (!targetData
+        && !outputData
+        && IsTargetTypeSameAsOutput(targetType, targetData, outputType, outputData)) {
+        // Emitted via guarantees that  targetData is always the default value of targetType.
+        outputType->CopyData(targetData, outputData);
```
cglzaguilar: Doing InitData here as Craig suggested, will make the DFIR change unnecessary. Please try this suggestion.

siddhukrs: @spathiwa and @cglzaguilar 

Can I do something like this? If yes, I will likely make this a helper method in TypeAndDataManager.cpp:

`
        TypeRef t = outputType;
        while (t->HasCustomDefault()) {
            t = t->BaseType();
        }
        t->InitData(outputData);
`

siddhukrs: Went back to using `initData`. We should be protected in this case, because the output terminal should always be a local with the default-default value in via. That said, I am filing an issue to come up with a `InitDefaultData` that does what I described above. The reason I am not doing it now is that I would like this API to have dedicated tests, and I don't have the time for it at the moment.

source/core/Variants.cpp
```diff
@@ -86,6 +77,18 @@ bool IsTargetTypeCompatibleWithInput(TypeRef inputType, TypeRef targetType, void
     return inputType && (!targetData || inputType->IsA(targetType, true));
 }
 
+void SetVariantToDataTypeError(TypeRef inputType, TypeRef targetType, void* targetData, TypeRef outputType, void* outputData, ErrorCluster* errPtr)
```
sanmut: targetData not used anymore. Delete.

sanmut: Make _ParamImmediateDef(StaticTypeAndData, TargetType) as StaticType instead as we don't care about data in that anymore.

siddhukrs: Like I mentioned the other day, we do require access to the data to check whether the terminal is connected to anything. If not (if the data is null), we should treat the target type as a variant. We cannot handle this case if we make it just StaticType. Look at how targetData is used below.

sanmut: I thought we were going to (or already doing with your change) to emit a variant constant (from DFIR) if the "type" terminal is unwired in G. It is better to do that now, as that will become a required terminal in G at some point in the future. Then, we can make this terminal as a "required" input and not an "optional" input for this primitive in VIA. I think that will be a lot cleaner.

siddhukrs: I am doing that, but it seemed like a good idea to have an extra check to guard against hand-written via that doesn't adhere to the insert-typed-local-if-null requirement. 

Are you worried that we would accidentally use the data from the type terminal? It doesn't seem like a breach of contract to me, since we are wiring in a type+data instance.

siddhukrs: Done.

PR #597, opened 2019-04-05
Introduce a new Vireo type for Variant

sanmut:
Introduce a new Vireo data type that inherits from WrappedType for supporting Variant data type.

This PR aims at achieving parity with the current functionality and making all the existing tests pass.


sanmut:
@spathiwa , @cglzaguilar , As I am out with sick kid, I did not find time to clean this up more before raising this PR. I did not want to delay this as I would like to get feedback on the approach to make sure this is a good step in the right direction. 

There is some work left to make it "more first class type" to support other committed and non-committed features (for instance, nesting variants). 

@spathiwa , I attempted a couple of other ways including making the VariantType a DefaultValueType but those did not turn out to work smoothly. I apologize I could not meet with you y'day afternoon with other meetings and work items and today being out with sick kid. We can catch up Monday to discuss more details on this in person.

sanmut:
@spathiwa , @cglzaguilar , This is on top of the commit "14.3.10 Bug fixes for VariantToData and GetVariantAttribute instructions". I would like to wait for other pending PRs to be submitted before merging this. On the other hand, the main things to review are just the approach itself and probably best to review the variants.h and variants.cpp directly as is.

source/core/Variants.cpp
```diff
@@ -19,25 +19,117 @@ SDG
...
+    _hasCustomDefault = true;
+    _topAQSize = sizeof(void*);
+    _aqAlignment = sizeof(void*);
+    TypeRef copyInputType = DefaultValueType::New(typeManager, type, true);
+    _underlyingTypeRef = copyInputType;
+    _attributeMap = nullptr;
+    _wrapped = copyInputType;
+}
+
+NIError VariantType::ClearData(void*)
```
cglzaguilar: I think we might need to implement InitData and CopyData and the overloads that a count parameter. This might be required for Array support to work.

PR #598, opened 2019-04-11
Two type visitor

cglzaguilar:
Initial implementation to add support for Variant equality.

Variant equality checks are done a run-time. We are adding a visitor that traverses two types at the same type to see if the types and the data are the same.

The visitor will be helpful later to deal with conversions in VariantToData and getting attributes from a Variant.


cglzaguilar:
This is ongoing work done on a feature branch and not ready for review.

PR #605, opened 2019-04-15
Add support for enums, add more tests

siddhukrs:


siddhukrs:
Reamed classes to use "dual"
Added support for Enums.
Added more test cases
In data comparison, I made intrinsic clusters behave same as user-defined clusters. No more custom handling intrinsic clusters -- merged them into a single method.

PR #609, opened 2019-04-17
Add support for variant equality comparison

siddhukrs:
# DETAILS
Unlike current equality checks in Vireo, Variant equality cannot be decomposed into equality checks on underlying types during parse time. The decision should instead be made at run time.

# IMPLEMENTATION
This change introduces a type visitor that traverses through nested types and performs an operation on them to return a resultant value. In this case, the operation is an equality operation on the types. To make this infrastructure useful for type conversion work that is coming up, the visitor works in two passes. In the first pass, the types are traversed to make sure that they're compatible types. In the second pass, the data in these types are compared. (for conversion, the data values will be converted and copied over in the second pass instead)

# TESTING
I also added elaborate tests to verify that equality check works across scalar and container data types. For details on what the tests are doing, looking at the `TestResult_` log line in the vtr files will explain it. There are too many tests for me to screenshot here. If you would like to look at the gvi/gviweb files used to generate them, I can send them over.


test-it/ExpectedResults/TestVI_VariantsAreEqual.vtr
```diff
@@ -0,0 +1,2 @@
+uInt32VariantsAreEqual = false
```
cglzaguilar: Remove this test result and the associated VIA file. They were used as part of developing the feature. You already have better tests.

sanmut:
@cglzaguilar , @rajsite , Please remember to squash merge (and not rebase merge) this in master. We discussed about this last week. We will do "squash merge" by default for merging PRs into master hereafter. A PR author will request explicitly if they want to preserve their commits and commits to be rebase-merged -- in which case they will order their commits in a meaningful way (and rebase on top of latest master before PR can be merged).

source/core/DualTypeEqual.cpp
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    //------------------------------------------------------------
+    bool DualTypeEqual::ApplyIEEE754Binaries(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY)
+    {
+        bool areEqual = false;
+        if (typeRefX->TopAQSize() == sizeof(Single)) {
+            Single singleValueX = *static_cast<Single*>(pDataX);
+            Single singleValueY = *static_cast<Single*>(pDataY);
+            areEqual = (singleValueX == singleValueY);
```
spathiwa: Use something like this:
`areEqual = (singleValueX == singleValueY || (singleValueX != singleValueX && singleValueY != singleValueY))`
to easily check for both values being NaN (which should be considered equal in a variant comparison).

siddhukrs: Using `::isnan` instead for better readability. It is also being used all over our codebase and should be easier to maintain.

test-it/ExpectedResults/TestVI_VariantsAreEqual.vtr
siddhukrs: Done.

source/core/DualTypeVisitor.cpp
```diff
@@ -0,0 +1,323 @@
...
+        }
+        if (success) {
+            // compare attributes
+            if (!variantTypeX->_attributeMap && !variantTypeY->_attributeMap) {
+                return true;
+            } else if (variantTypeX->_attributeMap && variantTypeY->_attributeMap) {
+                if (variantTypeX->_attributeMap->size() != variantTypeY->_attributeMap->size()) {
+                    return false;
+                }
+                for (const auto attributePairInX : *variantTypeX->_attributeMap) {
```
sanmut: Since traversal order of std::map is ordered, iterating both attribute maps together in one loop will be more efficient. However, it is ok not to do this in the PR. Not a bad idea to leave a comment though.

sanmut:
> But variants do a 'bitwise' comparison for equivalence, not a 'math' test for equality.

Please add a test with ToVariant on NaN, Inf, -Inf and comparing two variants as a GVI test.

test-it/testList.json
```diff
@@ -361,7 +361,13 @@
                 "TestVI_SimpleArrayOfRefs.via",
                 "TestVI_SpreadSheetStringToArray.via",
                 "TestVI_UnflattenFromJSONInvalidCluster.via",
```
rajsite: TestVI_CompatibleDataTypes_VariantToData_DoesNotReportError.via
is not in testlist.json

rajsite: We should make a disabled section and place _TestVI_VariantEquality_VariantsWithVariantsNestedInsideThem.via in there if it is currently disabled

siddhukrs:
> > But variants do a 'bitwise' comparison for equivalence, not a 'math' test for equality.
> 
> Please add a test with ToVariant on NaN, Inf, -Inf and comparing two variants as a GVI test.

Added these tests as via in my previous commit. Also checking them in as gvis.

test-it/testList.json
siddhukrs: Accidentally added it. Removing the files.

source/core/DualTypeVisitor.cpp
```diff
@@ -0,0 +1,323 @@
...
+                    success = operation->IEEE754BinaryCompatible(typeRefX, typeRefY);
+                    break;
+                case kEncoding_Cluster:
+                    success = ClusterCompatible(typeRefX, pDataX, typeRefY, pDataY, operation);
+                    break;
+                case kEncoding_Enum:
+                    success = EnumCompatible(typeRefX, typeRefY, operation);
+                    break;
+                case kEncoding_Array: {
+                    if (typeRefX->Rank() == 1 && typeRefX->GetSubElement(0)->BitEncoding() == kEncoding_Unicode) {
```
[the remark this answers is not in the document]
siddhukrs: I am not entirely sure how to handle this. I will check back with you on this once you're back and fix it in a follow-up commit.

PR #610, opened 2019-04-25
Support conversion to compatible data types in VariantToData and GetVariantAttribute instructions

siddhukrs:
- `VariantToData` instructions consumes an input type along with a target type. As long as this target type is "compatible" with the input type, the input value must be coerced (converted) to the target type before being written to the output terminal. In case of successful conversion, the error wire should not report an error. 
- The `GetVariantAttribute` node behaves similarly in that the `default value` terminal dictates the target type for the attribute value. If this type is compatible with the actual attribute's type, then we try to convert the attribute value to this target type before writing to the output terminal. If unsuccessful, error is reported through the error wire.

This change uses the type visitor created as part of the equality checking PR along with a new `DataTypeConversion` operation that copies values over to the destination with the right data type.

For description on tests, see comments inline. You can find the source VIs here - https://ni.visualstudio.com/DevCentral/_git/ASW/pullrequest/24734


test-it/ViaTests/TestVI_CompatibleDataTypes_VariantToData_DoesNotReportError.via
```diff
@@ -0,0 +1,916 @@
+// Autogenerated Vireo assembly file
+// Generated on 04/24/2019 12:34:24
```
siddhukrs: See associated VTR files for a list of test cases.

test-it/ViaTests/TestVI_GetVariantAttribute_DefaultAttributeValueOfCompatibleDataTypeDoesNotError.via
```diff
@@ -0,0 +1,315 @@
+// Autogenerated Vireo assembly file
+// Generated on 04/25/2019 10:43:21
```
siddhukrs: See associated VTR files for a list of test cases.

source/core/DualTypeVisitor.cpp
```diff
@@ -188,19 +178,23 @@ namespace Vireo
...
     {
         // Verify number of dimensions are the same
         bool success = typeRefX->Rank() == typeRefY->Rank();
         // Verify each dimension has the same size
         if (success) {
-            TypedArrayCoreRef arrayX = *(static_cast<const TypedArrayCoreRef*>(typeRefX->Begin(kPARead)));
-            TypedArrayCoreRef arrayY = *(static_cast<const TypedArrayCoreRef*>(typeRefY->Begin(kPARead)));
+            TypedArrayCoreRef arrayX = *(static_cast<const TypedArrayCoreRef*>(pDataX));
+            TypedArrayCoreRef arrayY = *(static_cast<const TypedArrayCoreRef*>(pDataY));
+            if (operation->ShouldInflateDestination()) {
```
siddhukrs: This is an important change. I had to allow the operation to decide if the target needs to be inflated before they're considered compatible. Since there is not real data associated with the output terminal (yet), we do not want them to be considered incompatible.

source/core/DualTypeConversion.cpp
```diff
@@ -0,0 +1,339 @@
...
+            return true;
+        }
+        SubString typeXName, typeYName;
+        Boolean isTypeXIntrinsicClusterType = typeRefX->IsIntrinsicClusterDataType(&typeXName);
+        Boolean isTypeYIntrinsicClusterType = typeRefY->IsIntrinsicClusterDataType(&typeYName);
+        return typeXName.Compare(&typeYName);
+    }
+
+    //------------------------------------------------------------
+    bool DualTypeConversion::DoTypesHaveSameEncodingAndSize(TypeRef typeRefX, TypeRef typeRefY)
```
cglzaguilar: Move this method to the base class, so both the conversion and equal operations, share the same code.

source/core/DualTypeVisitor.cpp
```diff
@@ -129,23 +129,13 @@ namespace Vireo
         Boolean isTypeXIntrinsicClusterType = typeRefX->IsIntrinsicClusterDataType(&typeXName);
         Boolean isTypeYIntrinsicClusterType = typeRefY->IsIntrinsicClusterDataType(&typeYName);
         if (isTypeXIntrinsicClusterType && isTypeYIntrinsicClusterType) {
-            success = IntrinsicClustersCompatible(typeRefX, pDataX, typeRefY, pDataY, operation);
+            success = operation->AreIntrinsicClustersCompatible(typeRefX, typeRefY);
```
siddhukrs: Comparison of intrinsic clusters is now delegated to the operation. Equality treats ComplexSingle and ComplexDouble as  incompatible, but for conversion purposes, they are compatible.

source/core/DualTypeConversion.cpp
```diff
@@ -0,0 +1,339 @@
...
+                    break;
+                }
+                break;
+            case kEncoding_Enum:
+                switch (typeRefY->TopAQSize()) {
+                    case 0:
+                        break;
+                    case 1:
+                    {
+                        Int8 integerValueX = static_cast<Int8>(valueX);
```
cglzaguilar: Please create another template function for this code section that is kind of "duplicated" four times.

source/core/DualTypeConversion.cpp
siddhukrs: Done.

source/core/DualTypeConversion.cpp
siddhukrs: Done.

test-it/testList.json
```diff
@@ -380,7 +382,6 @@
                 "TestVI_VectorScalarBinaryOp.via",
                 "TestVI_VectorUnaryOp.via",
                 "TestVI_VectorUnary2OutputOp.via",
-                "TestVI_VariantToData_VariantWiredToDataTypeTerminal.via",
```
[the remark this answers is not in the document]
rajsite: I think the test is worth keeping around just to ensure unwired inputs in hand written via error instead of crash the runtime (understanding that via generated by the editor will not have this case).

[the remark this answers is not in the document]
siddhukrs: Done.

source/core/DualTypeVisitor.cpp
sanmut: As operation cannot be null (it is always one of the derivative objects of DualTypeOperation, it will be cleaner to mark the parameter as "const DualTypeOperation &" throughout in all the APIs. That will in turn make the methods of DualTypeOperation const -- which is true today but just not specified explicitly. We should just fix it.

source/core/DualTypeConversion.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+    //------------------------------------------------------------
+    bool DualTypeConversion::AreIntrinsicClustersCompatible(TypeRef typeRefX, TypeRef typeRefY)
+    {
+        if (typeRefX->IsComplex() && typeRefY->IsComplex()) {
+            return true;
+        }
+        SubString typeXName, typeYName;
+        Boolean isTypeXIntrinsicClusterType = typeRefX->IsIntrinsicClusterDataType(&typeXName);
+        Boolean isTypeYIntrinsicClusterType = typeRefY->IsIntrinsicClusterDataType(&typeYName);
```
sanmut: Don't we need to use isTypeXIntrinsicClusterType and isTypeYIntrinsicClusterType?

source/core/DualTypeConversion.cpp
```diff
@@ -0,0 +1,316 @@
...
+    bool DualTypeConversion::AreIEEE754BinaryCompatible(TypeRef typeRefX, TypeRef typeRefY)
+    {
+        bool typesAreCompatible = TypesAreCompatible(typeRefX, typeRefY);
+        return typesAreCompatible;
+    }
+
+    //------------------------------------------------------------
+    bool DualTypeConversion::AreIntrinsicClustersCompatible(TypeRef typeRefX, TypeRef typeRefY)
+    {
+        if (typeRefX->IsComplex() && typeRefY->IsComplex()) {
```
sanmut: We need to check for only Complex here? How about Timestamp type? Could you explain/add a comment why this is necessary?

source/core/DualTypeConversion.cpp
```diff
@@ -0,0 +1,316 @@
...
+                    *reinterpret_cast<UInt64*>(pDataY) = static_cast<UInt64>(valueX);
+                    break;
+                }
+                break;
+            case kEncoding_Enum:
+                switch (typeRefY->TopAQSize()) {
+                    case 0:
+                        break;
+                    case 1:
+                    {
```
sanmut: <nit> get rid of braces in every case as done in the other encoding types.

source/include/DualTypeVisitor.h
```diff
@@ -33,7 +33,7 @@ class DualTypeVisitor {
     bool IntrinsicClustersCompatible(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, DualTypeOperation* operation);
```
sanmut: Unused. Delete?

source/include/DualTypeVisitor.h
```diff
@@ -33,7 +33,7 @@ class DualTypeVisitor {
...
     bool UserDefinedClustersCompatible(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, DualTypeOperation* operation);
     bool StringCompatible(TypeRef typeRefX, TypeRef typeRefY);
-    bool ArrayCompatible(TypeRef typeRefX, TypeRef typeRefY, DualTypeOperation* operation);
+    bool ArrayCompatible(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, DualTypeOperation* operation);
 
     bool Apply(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, DualTypeOperation* operation);
     bool ApplyVariant(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, DualTypeOperation* operation);
```
sanmut: Can all the methods in DualTypeVisitor() be made static?

source/core/Variants.cpp
```diff
@@ -172,25 +173,43 @@ VIREO_FUNCTION_SIGNATURET(VariantToData, VariantToDataParamBlock)
...
+                DualTypeConversion dualTypeConversion;
+                bool typesCompatible = false;
                 TypeRef underlyingType = variant->_underlyingTypeRef;
                 if (targetType->IsVariant()) {
-                    if (outputData)
+                    if (outputData) {
                         *static_cast<VariantTypeRef *>(outputData) = VariantType::CreateNewVariantFromType(underlyingType);
-                } else if (underlyingType->IsA(targetType, true)) {
-                    if (outputData)
-                        targetType->CopyData(underlyingType->Begin(kPARead), outputData);
```
sanmut: I tend to think we should delegate to type's CopyData if the types are exact, because the type's CopyData can do an efficient Copy when it can. Either make the Visitor to find out whether they are exact or use IsA first to see whether the types are exact and use type's CopyData. Use visitor to copy only when the types are not exactly same but convertible and thus efficient copies cannot be made. We could make the decision in the Visitor's Apply methods also.

source/core/DualTypeConversion.cpp
siddhukrs: Not in this case. The call to `IsIntrinsicClusterDataType` is necessary to get `typeXName` and `typeYName` out.

source/core/DualTypeConversion.cpp
siddhukrs: Because `IsComplex` covers both `ComplexSingle` and `ComplexDouble`, which would fail the name check below although they're actually compatible for conversion purposes. It is OK for timestamp to go through the name check below.

Added comment.

source/include/DualTypeVisitor.h
siddhukrs: `TypesAreCompatible` is non-static, and aggregate types call into `TypesAreCompatible` with their subtypes. I changed the others to be static.

source/core/Variants.cpp
siddhukrs: I was wrong about `IsA` when I mentioned that it is a strict check. It does treat numeric types as compatible. See [here](https://github.com/ni/VireoSDK/blob/defac12ea6b1634cba2b8cd16d77aac4b4c1bb44/source/core/TypeAndDataManager.cpp#L795). So it isn't really an exact type check. The version of `IsA` without the boolean is even less strict because it doesn't dig through nested types.

Is there a different method that can tell us if it is an exact match? Sounds to me like we should implement a third `operation` for the visitor (like `equals` and `conversion`) that does strict type checking..

source/core/Variants.cpp
```diff
@@ -172,25 +173,43 @@ VIREO_FUNCTION_SIGNATURET(VariantToData, VariantToDataParamBlock)
...
-                } else if (errPtr) {
+                    }
+                    typesCompatible = true;
+                } else {
+                    if (outputData) {
+                        typesCompatible = visitor.Visit(underlyingType, underlyingType->Begin(kPARead), outputType, outputData, &dualTypeConversion);
+                    } else {
+                        typesCompatible = underlyingType->IsA(targetType, true);
```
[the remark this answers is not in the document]
siddhukrs: Yes, `IsA` in this case is wrong and does the same thing it was doing before. I have now updated it to use just the type compatibility checking part of the `Visit` method.

source/core/DualTypeVisitor.cpp
siddhukrs: Done.

source/core/Variants.cpp
siddhukrs: Created this tech debt item to track this performance optimization - https://ni.visualstudio.com/DevCentral/_workitems/edit/142216

sanmut: > It does treat numeric types as compatible. See here. So it isn't really an exact type check. 

I think it is a strict enough check. It checks for different numeric encoding but it does look for the same size (TopAQSize) for the two types. So, for conversion operations, doing bitwise memory copy for integers of same size whether signed or unsigned should still be correct. I think we can fix the code to do what it used to do when the types are exact as determined by IsA with compatibleStructure as true. 

Basically, my point is that G code trying to use exact types of arrays and clusters of numerics (likely a common use case, especially for code written to use GTypes) should be allowed to do memcpy of the whole data (as the existing code is already doing before this change). Almost exact compatible types but  (as determined by IsA with compatibleStructure as true) seems to identify this already correctly.

The scripting tests we need to write for all permutations of different sizes of signed/unsigned numbers and arrays/clusters of numbers can be used to validate this. So, I am not too worried about making this change.

source/core/DualTypeVisitor.cpp
```diff
@@ -188,19 +176,23 @@ namespace Vireo
...
         // Verify each dimension has the same size
         if (success) {
-            TypedArrayCoreRef arrayX = *(static_cast<const TypedArrayCoreRef*>(typeRefX->Begin(kPARead)));
-            TypedArrayCoreRef arrayY = *(static_cast<const TypedArrayCoreRef*>(typeRefY->Begin(kPARead)));
+            TypedArrayCoreRef arrayX = *(static_cast<const TypedArrayCoreRef*>(pDataX));
+            TypedArrayCoreRef arrayY = *(static_cast<const TypedArrayCoreRef*>(pDataY));
+            if (operation.ShouldInflateDestination()) {
+                arrayY->ResizeToMatchOrEmpty(arrayX);
+            }
             IntIndex* dimensionLenghtsX = arrayX->DimensionLengths();
```
sanmut: Looks like I missed during the first pass. Apologies.
The while loop to check for dimensions may be done only if ShouldInflateDestination() returns false (as an else case), as the inflation case anyhow resizes to match.

source/include/DualTypeVisitor.h
```diff
@@ -23,23 +23,22 @@ namespace Vireo {
...
-    bool Apply(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, DualTypeOperation* operation);
-    bool ApplyVariant(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, DualTypeOperation* operation);
-    bool ApplyCluster(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, DualTypeOperation* operation);
-    bool ApplyString(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, DualTypeOperation* operation);
-    bool ApplyArray(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, DualTypeOperation* operation);
+    bool VariantCompatible(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, const DualTypeOperation &operation);
+    bool ClusterCompatible(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, const DualTypeOperation &operation);
+    bool EnumCompatible(TypeRef typeRefX, TypeRef typeRefY, const DualTypeOperation &operation);
+    bool UserDefinedClustersCompatible(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, const DualTypeOperation &operation);
+    static bool StringCompatible(TypeRef typeRefX, TypeRef typeRefY);
```
sanmut: Not in this change. However, StringCompatible() does not seem to look at both arguments but looks at only one. That should be fixed?

source/core/Variants.cpp
siddhukrs: That does sounds correct. I will make this change. 

I am working on the scripting tests for edge case numeric data types now and having this vireo version integrated will make things easier. If I do discover that something is broken, I will come back and fix it here.

source/include/DualTypeVisitor.h
siddhukrs: The visitor switch case is on `typeRefX`, and then calls `StringCompatible` which checks `typeRefY`. Am I missing something?

source/core/DualTypeVisitor.cpp
siddhukrs: Fair enough. Will move that to the else case.

source/include/DualTypeVisitor.h
sanmut: ah, I see. I just happened to look at the function alone without looking at the callsite. It was confusing to read.

PR #611, opened 2019-04-26
Fix equality of empty variants (with and without attributes)

siddhukrs:
Empty variants, with and without attributes, were failing to pass an equality check because our code wasn't handling nullptrs as data values. 

Fixing those cases and adding tests.


source/core/DualTypeVisitor.cpp
```diff
@@ -61,6 +61,9 @@ namespace Vireo
                         success = ArrayCompatible(typeRefX, typeRefY, operation);
                     }
                     break;
+                case kEncoding_None:
```
siddhukrs: The `underlyingType` on empty variant constants is of `kEncoding_None`. Treating them as equals in the `TypesCompatible` check and `Apply` cases.

source/core/Variants.cpp
```diff
@@ -439,7 +439,8 @@ bool VariantsAreEqual(VariantTypeRef variantX, VariantTypeRef variantY)
 {
     DualTypeVisitor visitor;
     DualTypeEqual dualTypeEqual;
-    return visitor.Visit(variantX, variantX->Begin(kPARead), variantY, variantY->Begin(kPARead), &dualTypeEqual);
+    return (!variantX && !variantY)  // Empty variant constant with no attributes are equal
```
siddhukrs: Empty variant constants with no attributes are plain nullptrs. Handling that special case here. 

However, if they do go through a `SetVariantAttribute` instruction they are initialized and have attribute maps and go through the visitor.

test-it/ExpectedResults/TestVI_VariantEquality_AttributeComparison.vtr
```diff
@@ -1,3 +1,6 @@
+TestResult_EmptyVariantConstantWithEqualDataAndUnequalAttributeNameAndValue true
```
siddhukrs: emptyvariants

siddhukrs: I also added a test that compares empty variant constant vs variant created using ToVariant.

cglzaguilar: Do we have tests that cover empty variant constant vs variant with data?

source/core/Variants.cpp
```diff
@@ -439,6 +439,11 @@ bool VariantsAreEqual(VariantTypeRef variantX, VariantTypeRef variantY)
 {
     DualTypeVisitor visitor;
     DualTypeEqual dualTypeEqual;
+    if (!variantX && !variantY) {
```
cglzaguilar: I think we need to move this check inside the visitor or the operation. For inner variants this case may present again.

source/core/DualTypeVisitor.cpp
cglzaguilar: I think we need to verify the encoding of typeRefY is also kEncoding_None and return true in that case.

siddhukrs:
Will re-open this PR after this other one is in: https://github.com/ni/VireoSDK/pull/610

PR #613, opened 2019-05-06
Fix variant equality comparison to allow empty variant constants

siddhukrs:
Empty variant constants are currently consumed as nullptrs since their data is not initialized. Changing up the equality visitors to handle this case.

I will include tests in a follow-up push. I wanted a review on the implementation in the meantime.

NOTE that this implementation doesn't handle the case where an empty variant constant is nested inside of an empty variant constant. That case is not supported in G today (on webVIs) and I don't see the point in handling that without implementing support for nested cases first. I updated the work item for nested variants to [track](https://ni.visualstudio.com/DevCentral/_workitems/edit/143896) this.


source/core/Variants.cpp
```diff
@@ -487,6 +487,11 @@ bool VariantsAreEqual(VariantTypeRef variantX, VariantTypeRef variantY)
 {
     DualTypeVisitor visitor;
     DualTypeEqual dualTypeEqual;
+    if (!variantX && !variantY) {
```
siddhukrs: Ideally, this check should happen inside the Visitor's Visit method when we handle Variant compatibility. However, since the TypeRefs are `nullptr` in case of empty variant constants, calling Begin(read) on it is not possible. 

Additionally, we will have to revisit this once nested variants are supported anyway.

source/core/DualTypeVisitor.cpp
```diff
@@ -60,6 +60,9 @@ namespace Vireo
                 }
                 break;
             }
+            case kEncoding_None:
+                success = typeRefY->BitEncoding() == kEncoding_None;
```
sanmut: I will add separate encoding for variant as part of a fix for a CAR. However, until then, I think we should check for IsVariant() on both types to find out if both of them are variant. 

Otherwise, just leave a TODO that kEncoding_None check exists only for checking variants (if it is) and that this should be fixed with kEncoding_Variant when we add a new encoding type. When I add the new encoding type, I will likely fix IsVariant() to use the encoding than the Name() as it will be more efficient.

siddhukrs: The `kEncoding_None` case is only encountered when we traverse the underlying type of an empty variant constant wired through a `setattribute`. I think we will still encounter this when variants have a separate encoding. I will add a note in the code mentioning that this is for the empty variant case.

test-it/ViaTests/EmptyVariantConstant_ComputeEqualsAndNotEquals_ReturnsExpected.via
```diff
@@ -0,0 +1,248 @@
+define (Application%3A%3AEmptyVariantConstant_ComputeEqualsAndNotEquals_ReturnsExpected%2Egviweb dv(.VirtualInstrument (
+    Locals: c(   // Data Space
```
siddhukrs: emptyvariants

Additional tests involving clusters and arrays containing empty variants and wired to equals and not equals are written as scripted tests. See here - https://ni.visualstudio.com/DevCentral/_git/ASW/pullrequest/25582?_a=overview

PR #614, opened 2019-05-09
Always include the JSAPI reference for internal functions

rajsite:
This change makes it so that the jsapi is always included for internal functions. This enables ASW to reflect on the `arguments` property of a function to count the number of arguments that were passed in. This way a single function can be called with a variable number of arguments and always get access to the jsapi reference.

This is for the following user story: https://ni.visualstudio.com/DevCentral/_workitems/edit/120177/

### DIFF
--- a/source/io/module_javaScriptInvoke.js
+++ b/source/io/module_javaScriptInvoke.js
@@ -428,7 +428,7 @@ var assignJavaScriptInvoke;
             };
 
             var jsapi;
-            if (functionToCall.length === parameters.length + 1) {
+            if (isInternalFunction || functionToCall.length === parameters.length + 1) {
                 jsapi = generateAPI(occurrencePointer, functionName, returnValueRef, errorValueRef, completionCallbackStatus, isInternalFunction);
                 parameters.push(jsapi);
             }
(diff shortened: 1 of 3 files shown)


source/io/module_javaScriptInvoke.js
```diff
@@ -428,7 +428,7 @@ var assignJavaScriptInvoke;
             };
 
             var jsapi;
-            if (functionToCall.length === parameters.length + 1) {
+            if (isInternalFunction || functionToCall.length === parameters.length + 1) {
```
ggumen: Should there be some error if: (isInternalFunction && functionToCall.length != parameters.length + 1) ?

rajsite: Nope, the use case is that an internal function can use the `arguments` variable in a function so the same function can be used with a variable number of parameters.

ie:
```javascript
// myFunc is called by multiple JavaScriptInvoke calls. One calls with zero parameters, one calls with 1, one calls with 5, etc

window.myFunc = function () {
    for (var i = 0; i < arguments.length; i++) {
        const currArg = arguments[i];
        // do something with currArg
    }
}

```

PR #615, opened 2019-05-10
Fix DeleteVariantAttribute bug in cases where there are no attributes to delete

siddhukrs:
When the name parameter is empty/null, `DeleteVariantAttribute` deletes all attributes.

There are two bugs associated with this:
1. There is a bug in Vireo where the `DeleteVariantAttribute` node returns `found?` as true when the name is empty/null, even when there were no attributes to delete. This is incorrect. In this PR, I update it to return `false` if no actual attributes were deleted.
2. There is a bug in NXG desktop and CG LabVIEW where the `found?` value is different based on whether the input is an empty variant constant vs. variant with data.
i. Empty variant constant (no attributes) and name empty returns `found = false`.
ii. Variant of numeric (no attributes) and name empty returns `found = true`. This is incorrect, since no attributes were "found" or deleted.

(@spathiwa let me know if you disagree that case 2 is a bug, and let me know if you need a CAR for CG)

I filed a bug to fix the second behavior in NXG desktop, and I am fixing it here in Vireo.


test-it/ViaTests/VariantWithNoAttributes_DeleteVariantAttributeWithNameUnspecified_ReturnsFoundAsFalse.via
```diff
@@ -0,0 +1,89 @@
+// Autogenerated Vireo assembly file
+// Generated on 05/10/2019 11:55:20
+//---------------------------------------------------
+//VI Definition: ::Web Server::Interactive::Application::DeleteVariantAttributeWithNoName_DeletesAllAttributesAndReturnsFalse.gviweb
```
siddhukrs: DeleteNoAttributes

test-it/ViaTests/TestVI_VariantDeleteAttributes.via
```diff
@@ -153,7 +153,7 @@ define (Application%3A%3ATestVI_VariantDeleteAttributes%2Egviweb dv(.VirtualInst
         ce(dv(.Boolean true)c146)
         ce(dv(.ErrorCluster (false 0 '' ))c147)
         e(.Boolean local148)
-        ce(dv(.Boolean true)c149)
+        ce(dv(.Boolean false)c149)
```
siddhukrs: This is related to the Case 2 I described above. This comes from a cross-target test that worked on both desktop and vireo previously, but incorrectly. It was returning `found = false` when no attributes were deleted (From a variant all of whose attributes were already deleted. This was trying to delete them a second time). The test passed in desktop because of bug 2, and it passed in Vireo because of the bug 1, as described in the overview of this PR.

sanmut:
> I filed a bug to fix the second behavior in NXG desktop, and I am fixing it here in Vireo.

Good catch. I am assuming you are creating a cross target test in ASW and disabling it on Desktop.

siddhukrs:
> > I filed a bug to fix the second behavior in NXG desktop, and I am fixing it here in Vireo.
> 
> Good catch. I am assuming you are creating a cross target test in ASW and disabling it on Desktop.

The test I am modifying here is a cross-platform test in ASW that was incorrectly passing on both Desktop and Web due to two separate bugs (this and the Desktop/CG bug). I am fixing the test asset to test the correct behavior and disabling it on Desktop with a link to the CAR.

PR #616, opened 2019-05-10
Treat variants with no data or attributes as an empty variant constant in equality operations

siddhukrs:
Since an empty variant constant is a `nullptr` under the hood, it is not "equal" to an empty variant constant goes through a `SetVariantAttribute` and then a `DeleteVariantAttribute` to have that attribute deleted. This is because after `SetVariantAttribute`, the variant is no longer a `nullptr` because we initialize an attribute map.

In this PR, rather than just check for `nullptr` to identify an empty variant constant, I am extending the check to verify that the data is empty and that there are no attributes if not null.


test-it/ViaTests/EmptyVariantConstant_SetAndDeleteAttribute_EqualsEmptyVariant.via
```diff
@@ -0,0 +1,110 @@
+// Autogenerated Vireo assembly file
+// Generated on 05/10/2019 15:48:44
```
siddhukrs: compareemptywithnonempty

PR #617, opened 2019-05-15
Support arbitrary paths for VTR test suite runner

sanmut:
Allow either relative or absolute path of via files as command-line
argument for the test runner.
Exampleis
node test.js --once -j /c/temp/MyTests/ViaTests/*.via
node test.js --once -j ../../../../temp/MyTests/ViaTests/*.via

### DIFF
--- a/test-it/test.js
+++ b/test-it/test.js
@@ -181,8 +181,8 @@
 
     // Process the via files for the provided tester function and compare with
     // the results file of the same name but with a '.vtr' extension
-    var runTestCore = function (testName, tester, testFinished, execOnly) {
-        var resultsFileName = 'ExpectedResults/' + path.basename(testName, '.via') + '.vtr';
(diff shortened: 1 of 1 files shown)


sanmut:
This is to allow running the test suite locally on test files at a different location for debugging purposes.

test-it/test.js
```diff
@@ -510,7 +516,7 @@
                         process.exit(errorCode);
                     });
                 }
-            });
+            }, execOnly, testsBaseDir);
```
rajsite: Here runNextTest is being invoked with four parameters (testFiles, a function, execOnly, testsBaseDir) but the runNextTest function only has two arguments (testFiles, chain) and does not seem to do anything fancy for retrieving these additional arguments. Are we sure this is behaving as expected?

sanmut:
I think with recent Ram's changes to generate .via files, this may not be sufficient. I will try this later with his changes and reopen this PR later.

PR #618, opened 2019-05-19
Redesign and Refactor Variant Type and Data

sanmut:
- Crate a new encoding for Variant type
- Emit instruction for Conversion
- Emit instructions for Comparison
- Make InitData/CopyData of underlyingTypeRef consistent
- Make DualTypeVisitor to accept pointer to VariantDataRef to
  be consistent with other Vireo types
- Add comment on unnecessary copies in ArrayType::CopyData
- Attempt to treat underlyingType as innerVariant. Without
  much effort, nested variants seem to work
- Add tests for setting "VariantWithAttributes" as a variant attribute
- Add test for nesting variants and setting attributes at different
  nesting levels
- Improve complexity of DualTypeVisitor variant attribute walkking
  algorithm
- Remove the workaround in CopyVariant where a new variant
  type had to be allocated to allow empty variant constants and
  terminals to work with attributes. This is mainly what prompted
  the refactor. Empty variant constant or local variable (terminal)
  is handled a lot cleaner in this new design.


sanmut:
This fixes CAR 736230: Using variant with attributes as attribute for variant does not work correctly

sanmut:
@siddhukrs , could you please take a quick look at the DualType*.* files?

sanmut:
TestVI_NestedVariantWithAttributesAtNestedLevels
TestVI_SetVariantWithAttributeAsVariantAttribute_AndGetAttributeOneByOne
TestVI_SetVariantWithAttributeAsVariantAttribute_AndGetAttributesAll

sanmut:
TestVI_NestedVariantWithAttributesAtNestedLevels
 
VI Image corresponding to TestVI_NestedVariantWithAttributesAtNestedLevels.via

sanmut:
TestVI_SetVariantWithAttributeAsVariantAttribute_AndGetAttributeOneByOne

VI Image corresponding to TestVI_SetVariantWithAttributeAsVariantAttribute_AndGetAttributeOneByOne.via

sanmut:
TestVI_SetVariantWithAttributeAsVariantAttribute_AndGetAttributesAll

VI Image corresponding to TestVI_SetVariantWithAttributeAsVariantAttribute_AndGetAttributesAll.via

source/core/DualTypeVisitor.cpp
```diff
@@ -29,97 +29,99 @@ namespace Vireo
...
         }
         return success;
     }
 
     //------------------------------------------------------------
     bool DualTypeVisitor::VariantCompatible(TypeRef typeRefX, void* pDataX, TypeRef typeRefY, void* pDataY, const DualTypeOperation &operation)
     {
-        VariantTypeRef variantTypeX = reinterpret_cast<VariantTypeRef>(typeRefX);
-        VariantTypeRef variantTypeY = reinterpret_cast<VariantTypeRef>(typeRefY);
+        VariantDataRef variantX = *reinterpret_cast<VariantDataRef *>(pDataX);
```
sanmut: @siddhukrs , This is one notable functional change to keep this interface consistent in handling pointer to data. The caller should pass pointer to VariantDataRef.

test-it/testList.json
```diff
@@ -1,10 +1,5 @@
 {
     "tests": {
-        "disabled": {
-            "tests": [
-                "TestVI_VariantEquality_VariantsWithVariantsNestedInsideThem.via"
```
sanmut: I tried to just enable this and see what happens. There was only one bug that I hat to fix to make this work. So, enabled it and wrote a couple of more complex tests and added them to the PR.

source/core/DualTypeVisitor.cpp
```diff
@@ -29,97 +29,99 @@ namespace Vireo
...
-                break;
-            case kEncoding_Array: {
-                if (typeRefX->Rank() == 1 && typeRefX->GetSubElement(0)->BitEncoding() == kEncoding_Unicode) {
-                    success = StringCompatible(typeRefX, typeRefY);
-                } else {
-                    success = typeRefY->BitEncoding() == kEncoding_Array && ArrayCompatible(typeRefX, pDataX, typeRefY, pDataY, operation);
-                }
-                break;
-            }
-            case kEncoding_None:
```
sanmut: Adding a new encoding type per @spathiwa 's advice cleaned this up a lot in using the encoding type for variants. Thanks a lot Craig.

source/core/Variants.cpp
```diff
@@ -26,53 +26,76 @@ SDG
...
-        || (variant->_underlyingTypeRef->BitEncoding() == kEncoding_None  // variant's data is empty
-            && IsAttributeMapNullOrEmpty(variant));  // variant's attributes are not set or empty.
-}
+    NIError err = kNIError_Success;
 
-bool VariantType::IsAttributeMapNullOrEmpty(VariantTypeRef variant)
-{
-    return variant->_attributeMap == nullptr || variant->_attributeMap->size() == 0;
+    auto variantValue = static_cast<VariantDataRef *>(pData);
+    if (*variantValue != nullptr) {
```
cglzaguilar: We might want to call InitData on a pointer already initialized to set it to the default value. In that scenario consider instead deleting the current content of pData and then continue with the initialization code.

source/core/Variants.cpp
```diff
@@ -26,53 +26,76 @@ SDG
...
+        VIREO_ASSERT((*variantValue)->IsUninitializedVariant());
+    } else {
+        if (pattern == nullptr) {
+            pattern = this;
+        }
+        VariantDataRef newVariant = VariantData::New(pattern);
+        *static_cast<VariantDataRef*>(variantValue) = newVariant;
```
cglzaguilar: I think we shouldn't need the static_cast here. *variantValue = newVariant, should work. Is that not the case?

source/core/DualTypeVisitor.cpp
```diff
@@ -29,97 +29,99 @@ namespace Vireo
...
+                variantX->GetInnerData(),
+                underlyingTypeY,
+                variantY->GetInnerData(),
+                operation)) {
                 return false;
             }
         }
-        return success;
+
+        if (variantX->HasMap() && variantY->HasMap()) {
```
siddhukrs: The previous implementation was comparing attribute-names by querying the name from one map against the other, we don't seem to do that anymore. Is that correct?

source/core/DualTypeVisitor.cpp
siddhukrs: Yup, I think this is how I wanted it to be, but for some reason I was forced to use the TypeRef to access data.  This makes me more comfortable. Thanks for this change!

source/core/DualTypeVisitor.cpp
siddhukrs: Well,  just noticed that we compare names as part of the Apply operation. However, by comparing attribute values in the order in which they appear, we could likely be comparing TypeAreCompatible on attribute values that are not associated with the same attribute name. But I think that is fine because it would help us filter out a subset of incompatible maps, and then verify them again during Apply. So I think it is fine this way.

sanmut: Yeah, tried to follow: "check for types in *Compatible method and look at values in Apply* method" as type checks are supposed to be faster than value checks.

source/core/GenericFunctions.cpp
```diff
@@ -457,14 +457,27 @@ InstructionCore* EmitGenericBinOpInstruction(ClumpParseState* pInstructionBuilde
...
-                snippetBuilder.InternalAddArg(nullptr, clusterOp == kFakedInstruction ? nullptr : clusterOp->_piSnippet);
+                snippetBuilder.InternalAddArgBack(nullptr, clusterOp == kFakedInstruction ? nullptr : clusterOp->_piSnippet);
                 snippetBuilder.EmitInstruction();
                 pInstructionBuilder->EndEmitSubSnippet(&snippetBuilder);
             }
 
             pInstructionBuilder->RecordNextHere(&clusterOp->_piNext);
             break;
         }
+        case kEncoding_Variant:
```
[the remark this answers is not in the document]
sanmut: With the new design, types other than Variant do not get resolved correctly with DEFINE_VIREO_FUNCTION_CUSTOM. As you can see we were emitting CopyVariant from EmitGenericCopyInstruction before. Now, this is consistent to emit Convert and IsEQ variations also using the same Emit mechanism.

source/core/GenericFunctions.cpp
```diff
@@ -531,6 +544,15 @@ InstructionCore* EmitGenericUnOpInstruction(ClumpParseState* pInstructionBuilder
 
     InstructionCore* pInstruction = nullptr;
     switch (destType->BitEncoding()) {
+        case kEncoding_Variant:
+        {
+            ConstCStr unaryOpName = "DataToVariant";
+            SubString unaryOpToken(unaryOpName);
+            pInstructionBuilder->InternalAddArgFront(nullptr, pInstructionBuilder->_argTypes[0]);
```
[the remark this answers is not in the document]
sanmut: ah, good catch. Missed to check if it is "Convert" before emitting "DataToVariant". I think we got lucky because other unary operations throw SA error and we may never hit this except "Convert". Fixing it now to add the check. Thanks.

source/core/GenericFunctions.cpp
```diff
@@ -531,6 +544,15 @@ InstructionCore* EmitGenericUnOpInstruction(ClumpParseState* pInstructionBuilder
 
     InstructionCore* pInstruction = nullptr;
     switch (destType->BitEncoding()) {
+        case kEncoding_Variant:
```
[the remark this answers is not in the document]
sanmut: ah, good catch. Missed to check if it is "Convert" before emitting "DataToVariant". I think we got lucky because other unary operations throw SA error and we may never hit this except "Convert". Fixing it now to add the check. Thanks.

source/core/GenericFunctions.cpp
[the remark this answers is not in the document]
sanmut: You build the arguments the way the instruction expects. For conversion, DataToVariant expects arguments in the order and we have to add in the front. The existing instructions did not run into this situation and they always required additional arguments at the end.

source/core/Variants.cpp
sanmut: I think InitData should not delete data. It looks like InitData for other types are not designed that way. I have added an assert to find out if we will ever get to the scenario. We can address this if there is a valid test case. This seems to work correctly for variant inside array and cluster correctly. We should refrain from changing this without a test case. Also, please see my detailed comment in ArrayType::InitData. We should address that issue for all the types correctly to avoid the performance issue.

source/core/Variants.cpp
```diff
@@ -83,65 +106,256 @@ void VariantType::SetVariantToDataTypeError(TypeRef inputType, TypeRef targetTyp
         errPtr->SetErrorAndAppendCallChain(true, kVariantIncompatibleType, "Variant To Data");
     }
     if (outputType && outputData) {
-            outputType->InitData(outputData);
+        outputType->ClearData(outputData);
```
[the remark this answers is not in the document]
sanmut: The reason InitData and ClearData are separated this way is to save freeing and allocating pointer to VariantData itself for named variables. Otherwise it is unnecessary deallocation for which we are need to allocate again. We can discuss this in person tomorrow in detail if you still have questions.

PR #619, opened 2019-05-21
Variant cluster to array

sanmut:
Add support for converting a variant of cluster to either "Array of Variants" or "Array of inner type of the cluster".


sanmut:
@cglzaguilar, @spathiwa, @siddhukrs , Please review **ONLY** the changes in the commit **"VariantToData:Cluster To Array conversion"** and any **newer that may get added**. The older commit "Redesign and Refactor Variant Type and Data" is exactly same as  https://github.com/ni/VireoSDK/pull/618 . This may be submitted after https://github.com/ni/VireoSDK/pull/618 goes in.

test-it/ExpectedResults/TestVI_VariantOfClusterToArrayOfInnerType.vtr
```diff
@@ -0,0 +1,3 @@
+TestResult_ClusterOfDifferentSignednessNumericTypes_ConvertsCorrectly true
```
siddhukrs: I don't see tests that verify incompatible types report errors. That is an important test to have -- incompatible due to different array dimension and array data type. Also, one test that exercises the inflation codepath. I am guessing this is when the output terminal is unwired.

sanmut:
TestVI_VariantOfClusterToArrayOfInnerType
TestVI_VariantOfClusterToArrayExpectedErrors
TestVI_VariantOfClusterToArrayOfVariants

test-it/ExpectedResults/TestVI_VariantOfClusterToArrayOfInnerType.vtr
sanmut: Good point. Added a test for expected errors. Added one more test with cluster inside a cluster for non-error case. They are all passing.

sanmut:
> Changes look good to me. Please rebase your changes, so that they can be squashed.

Done. Thanks Carlos.

PR #620, opened 2019-05-29
Add support for waveform attributes and remove AnalogWaveformBuild in lieu of cluster field writes

siddhukrs:
- Add a variant field to the cluster that represents an analog waveform to store waveform attributes.
- Get rid of the `AnalogWaveformBuild` (and a struct in Waveform.h that was merely used to hold on to the waveform's fields). Instead, VIA would use Copy instructions to copy into the waveform's fields. On LabVIEW, the `BuildWaveform` node would decompose into property node reads and writes.
- Update existing tests that were using `AnalogWaveformBuild` to use copy instructions to copy into the waveform cluster's fields instead.
- Implement a new formatter for Variant data types to be serialized to String. This is also used by the Read/Write JSON API. Unfortunately, implementing WriteJSON for variant data is complex, and not worth investing in at this point, considering the fact that it isn't actively used. However, there is still value in the string representation for the VTR test suite that readJSON can benefit from.
- Updated tests to use the new Variant string representation in their VTR files.


test-it/ExpectedResults/Waveform.vtr
```diff
@@ -1,35 +1,35 @@
 Analog Waveform Testing
 ***********************
-Empty analog waveform a_wave = ((0 0) 0 ())
+Empty analog waveform a_wave = ((0 0) 0 () ^Variant)
 Analog parts (0 0) 0 ()
-Analog waveform a_wave = ((0 0) 0 (1.2 1.3 1 -0.5))
-Analog waveform a_wave = ((3566073600 123) 5.8 (1.2 1.3 1 -0.5))
+Analog waveform a_wave = ((0 0) 0 (1.2 1.3 1 -0.5) ^Variant)
```
siddhukrs: This works now, but the `^Variant` doesn't read well. I think it is because variants do not have a custom print function?

siddhukrs: Fixed this to allow Variants be printed in the following format:
`{"value": <expand-underlying-type>, "attributes":["foo": <expand-attribute-value>, "bar": <expand attribute value>]}`

source/core/TDCodecVia.cpp
```diff
@@ -2329,6 +2329,50 @@ void TDViaFormatter::FormatClusterData(TypeRef type, void *pData)
     _string->Append(Fmt()._clusterPost);
 }
 //------------------------------------------------------------
+void TDViaFormatter::FormatVariant(TypeRef type, void *pData)
+{
+    _string->Append(Fmt()._clusterPre);
```
siddhukrs: Variants are represented as follows:
`{"value": <expand-underlying-type>, "attributes":["foo": <expand-attribute-value>, "bar": <expand attribute value>]}`

When either of these fields are null, you will see something like this:
`{"value": null, "attributes": null}`

source/core/Variants.cpp
```diff
@@ -37,6 +37,7 @@ VariantType::VariantType(TypeManagerRef typeManager)
     _topAQSize = TypeManager::HostPointerToAQSize();
     _aqAlignment = TypeManager::AQAlignment(sizeof(void*));
     _encoding = kEncoding_Variant;
+    _isValid = 1;
```
siddhukrs: Lack of this field was causing a waveform test to fail. The readJSON API verifies that the typeref being read is "valid" before reading it.

test-it/karma/publicapi/TypeWaveform.Test.js
```diff
@@ -40,9 +41,12 @@ describe('Peek/Poke different datatypes', function () {
                 fraction: '0'
             },
             dt: 0,
-            Y: [] // eslint-disable-line id-length
+            Y: [], // eslint-disable-line id-length
+            attributes: { value: null, attributes: null }
         });
 
+        /*
```
siddhukrs: Supporting readJSON is complex and beyond what this PR is trying to get to. Since we do not really use the JSON API much, and there being more value in implementing the string formatter for Variants (they are used by the VTR test suite), I discussed with Milan and decided to disable these tests. The rest of the tests here give sufficient coverage.

test-it/karma/publicapi/WriteJson.Test.js
```diff
@@ -1044,38 +1044,43 @@ describe('The Vireo EggShell writeJSON api can write', function () {
             writeTest('dataItem_ClusterOfArrays', original, lessVals);
         });
 
-        it('analog waveform of double', function () {
+        // Vireo does not yet support writeJSON on Variant data types.
+        xit('analog waveform of double', function () {
```
siddhukrs: See comment above in `AllocateTypes.Test.js` as to why this test is disabled.

source/core/TDCodecVia.cpp
```diff
@@ -2329,6 +2329,56 @@ void TDViaFormatter::FormatClusterData(TypeRef type, void *pData)
...
+    _string->Append(*tsNameSuffix);
+    if (variantData->HasMap()) {
+        _string->Append(Fmt()._clusterPre);
+        auto iterator = variantData->GetMap_cbegin(), iteratorBegin = variantData->GetMap_cbegin(), iteratorEnd = variantData->GetMap_cend();
+        for (; iterator != iteratorEnd; ++iterator) {
+            if (iterator != iteratorBegin) {
+                _string->Append(Fmt()._itemSeparator);
+            }
+            StringRef attributeName = iterator->first;
+            if (useQuotes)
```
rajsite: Can you do the pattern shown here?: https://github.com/ni/VireoSDK/blob/669a35df79af13ccac2c3f3039bc54841b5fbfca/source/core/TDCodecVia.cpp#L2235

Then we can avoid hardcoding the quote character and make sure to escape the attribute name string

test-it/karma/publicapi/TypeWaveform.Test.js
rajsite: Was this intended to say writeJSON?

test-it/karma/publicapi/TypeWaveform.Test.js
```diff
@@ -40,9 +41,12 @@ describe('Peek/Poke different datatypes', function () {
                 fraction: '0'
             },
             dt: 0,
-            Y: [] // eslint-disable-line id-length
+            Y: [], // eslint-disable-line id-length
+            attributes: {_data: null, _attributes: null}
         });
 
+        /*
```
rajsite: Instead of commenting out the source could we either move it to its own it statement that is marked `xit` or just delete it and leave a comment saying that writeJSON for waveforms is not supported near the top of the file.

source/core/TDCodecVia.cpp
```diff
@@ -2329,6 +2329,56 @@ void TDViaFormatter::FormatClusterData(TypeRef type, void *pData)
...
 }
 //------------------------------------------------------------
+void TDViaFormatter::FormatVariant(TypeRef type, void *pData)
+{
+    _string->Append(Fmt()._clusterPre);
+    Boolean useQuotes = Fmt().QuoteFieldNames();
+    VariantDataRef variantData = *reinterpret_cast<VariantDataRef *>(pData);
+
+    if (useQuotes)
+        _string->Append('\"');
```
rajsite: instead use `Fmt()._quote`

source/core/TDCodecVia.cpp
```diff
@@ -2329,6 +2329,56 @@ void TDViaFormatter::FormatClusterData(TypeRef type, void *pData)
...
+    _string->AppendCStr(variantInnerData);
+    if (useQuotes)
+        _string->Append('\"');
+    _string->Append(*tsNameSuffix);
+
+    TypeRef underlyingType = variantData->GetInnerType();
```
sanmut: I would rename "underlyingType" to "innerType" to match the API being called.

test-it/ExpectedResults/VariantAttributes.vtr
```diff
@@ -1,30 +1,30 @@
...
-Set: myVar=^Variant, replaced=false error=(false 0)
-Delete: myVar=^Variant, found=false error=(false 0)
-Get: myVar=^Variant, found=true error=(false 0) values=(xxx)
-Delete: myVar=^Variant, found=true error=(false 0)
-Get: myVar=^Variant, found=false error=(false 0) values=(z)
-Set: myVar=^Variant, replaced=false error=(false 0)
-Set: myVar=^Variant, replaced=false error=(false 0)
-Delete: myVar=^Variant, found=true error=(false 0)
-Get: myVar=^Variant, found=(false false) error=(false 0) values=(z 0)
+Set: myVar=(_data:255 _attributes:(aaa:(_data:xxx _attributes:null))), replaced=false error=(false 0)
```
sanmut: Values of data types used in variant (as data or attribute name or attribute value) should be formatted same as if they were used directly for the data type. For example, strings should be single quoted. Here "aaa" and "xxx" are printed without quotes.  It is likely because we are formatting in the new function directly as opposed to calling "Format" to format the string.

source/core/TDCodecVia.cpp
```diff
@@ -2329,6 +2329,56 @@ void TDViaFormatter::FormatClusterData(TypeRef type, void *pData)
     _string->Append(Fmt()._clusterPost);
 }
 //------------------------------------------------------------
+void TDViaFormatter::FormatVariant(TypeRef type, void *pData)
```
sanmut: I would add a test to print for nested variant with attributes. You may just add print to one of the recently added nested variant test via files.

source/include/Waveform.h
```diff
@@ -1,31 +0,0 @@
-/**
```
sanmut: I think we might need this in the future. If there are operations on waveform that are not same as operations on simple cluster, then that should be encapsulated to a class like this (for example, Timestamp). When we need to deal with variants of waveforms, the DFIR decomposition would not help and we have to dispatch operations at runtime to do the right thing. Then, abstracting those operations to a Waveform class would be the right way to do. @spathiwa  thoughts?

test-it/karma/publicapi/TypeWaveform.Test.js
siddhukrs: Yup! Fixed it now.

source/include/Waveform.h
spathiwa: I agree. And there definitely are operations on Waveforms that are treated differently than a cluster; arithmetic operations use the waveform array and ignore the the other components, for example, as does inequality comparison with a scalar constant.

siddhukrs: OK. But there's hardly any code in this class, we can always easily add it back when necessary. Did you want me to leave it as it is but with no references? I am OK with either.

sanmut: It is good to leave it and feel free to leave a comment that it is not used now but will soon be used. I think we will get to use this in this cycle itself. I just don't want all operations to be spread out in future PRs. There are many nodes that don't have an error cluster and we will have to support them anyway.

siddhukrs: Added it back.

source/core/TDCodecVia.cpp
siddhukrs: Done. Added new test in `VariantWithVariantAsAttribute`

test-it/ExpectedResults/VariantAttributes.vtr
siddhukrs: Updated this. I wasn't setting a flag correctly.

source/core/TDCodecVia.cpp
siddhukrs: Done.

source/core/TDCodecVia.cpp
siddhukrs: Done.

test-it/karma/publicapi/TypeWaveform.Test.js
siddhukrs: Done.

source/core/TDCodecVia.cpp
siddhukrs: Done.

source/core/TDCodecVia.cpp
siddhukrs: Done.

PR #621, opened 2019-05-31
Implement getVariantAttribute and setVariantAttributeAsString

rajsite:
These changes expose two public api functions:
- getVariantAttribute
- setVariantAttributeAsString

It is possible to make a generic getVariantAttribute function as the function just returns a valueRef to the attribute value in memory.

However for setVariantAttribute to be generic the eggShell user would have to be able to allocate an arbitrary type and pass the valueRef to setVariantAttribute. Vireo's public api does not currently have a way to express the format of arbitrary types. This was an issue for events as well but was worked around for events by emitting a variable with a known name format that could be copied.

So instead of exposing a generic setVariantAttribute api we expose a setVariantAttributeAsString api to allow us to allocate a string to use as a variant attribute value from the JavaScript side. This is specifically to support the Waveform use case.

This PR is just for handling these variant functions. How it applies to waveforms, which contain a variant, will be handled in a separate PR.


source/core/CEntryPoints.cpp
```diff
@@ -159,6 +159,9 @@ VIREO_EXPORT EggShellResult EggShell_DeallocateData(TypeManagerRef tm, const Typ
 VIREO_EXPORT EggShellResult EggShell_FindValue(TypeManagerRef tm, const char* viName, const char* eltName, TypeRef* typeRefLocation, void** dataRefLocation)
 {
     TypeManagerScope scope(tm);
+    *typeRefLocation = nullptr;
```
rajsite: just making this function follow the same pattern of nulling out the pointer storage location at the start in-case we return early

source/core/CEntryPoints.cpp
```diff
@@ -303,6 +309,59 @@ VIREO_EXPORT EggShellResult EggShell_ResizeArray(TypeManagerRef tm, const TypeRe
...
+
+    if (typeRef == nullptr || !typeRef->IsValid() || !typeRef->IsVariant()) {
+        return kEggShellResult_InvalidTypeRef;
+    }
+
+    STACK_VAR(String, attributeSV);
+    StringRef attributeName = attributeSV.Value;
+    attributeName->AppendCStr(attributeNameCStr);
+
+    VariantDataRef variant = *(static_cast<const VariantDataRef*>(pData));
```
cglzaguilar: Please check that pData is not null.

source/core/CEntryPoints.cpp
```diff
@@ -303,6 +309,59 @@ VIREO_EXPORT EggShellResult EggShell_ResizeArray(TypeManagerRef tm, const TypeRe
...
+
+    if (attributeTypeRef == nullptr || !attributeTypeRef->IsValid()) {
+        return kEggShellResult_InvalidTypeRef;
+    }
+
+    STACK_VAR(String, attributeSV);
+    StringRef attributeName = attributeSV.Value;
+    attributeName->AppendCStr(attributeNameCStr);
+
+    VariantDataRef variant = *(static_cast<const VariantDataRef*>(pData));
```
cglzaguilar: Please check that pData is not null.

source/core/Variants.cpp
```diff
@@ -37,6 +37,7 @@ VariantType::VariantType(TypeManagerRef typeManager)
     _topAQSize = TypeManager::HostPointerToAQSize();
     _aqAlignment = TypeManager::AQAlignment(sizeof(void*));
     _encoding = kEncoding_Variant;
+    _isValid = 1;
```
cglzaguilar: I think this change was just done by Sid.

source/io/module_eggShell.js
```diff
@@ -507,6 +507,56 @@ var assignEggShell;
...
+            var eggShellResult = Module._EggShell_GetVariantAttribute(Module.eggShell.v_userShell, valueRef.typeRef, valueRef.dataRef, attributeNameStackPointer, typeStackPointer, dataStackPointer);
+            if (eggShellResult !== EGGSHELL_RESULT.SUCCESS && eggShellResult !== EGGSHELL_RESULT.OBJECT_NOT_FOUND_AT_PATH) {
+                throw new Error('Could not get variant attribute for the following reason: ' + eggShellResultEnum[eggShellResult] +
+                    ' (error code: ' + eggShellResult + ')' +
+                    ' (type name: ' + Module.typeHelpers.typeName(valueRef.typeRef) + ')' +
+                    ' (subpath: ' + attributeName + ')');
+            }
+
+            var typeRef, dataRef, subValueRef;
+            if (eggShellResult !== EGGSHELL_RESULT.OBJECT_NOT_FOUND_AT_PATH) {
```
siddhukrs: This check for `OBJECT_NOT_FOUND_AT_PATH` is the same as above and not necessary, right?

siddhukrs: Nevermind, it is what the above check is not doing.

rajsite: Nope, there are two cases that make it past the check, success and objectnotfound.

Guess it would be clearer to make the case check for success instead of not objectnotfound

rajsite: Updated the check to be clearer

source/core/Variants.cpp
rajsite: Rebased on top of Sid's changes in master

PR #622, opened 2019-06-03
Fix readJSON for several types and prevent memory corruption on reboot

### LINKED ISSUE #163: Aborts caused when using the same Vireo instance after an http 404
At the moment we are using a new Vireo instance for every test in a karma spec.

Ideally we could use the same vireo instance and just perform a reboot and loadVia between tests.

When the http tests are modified to use a single instance like in the following example branch: https://github.com/rajsite/VireoSDK/commit/6cc892c615f1d88ea02fb69d5074b46ae2f793a0#diff-3f40dfffc8c2df601f713d108805601f

We get the output error from the commit, reproduced as follows (using debug level -g4)
[out.txt](https://github.com/ni/VireoSDK/files/709716/out.txt)

With the following stack trace
```
Firefox 49.0.0 (Windows 7 0.0.0) ERROR
  uncaught exception: abort(5) at jsStackTrace@dist/vireo.js:1148:13
  stackTrace@dist/vireo.js:1165:22
  abort@dist/vireo.js:127869:44
  nullFunc_ii@dist/vireo.js:6181:464
  b5@dist/vireo.js:126969:13
  __ZN5Vireo14TypedArrayCore16ResizeDimensionsEiPib [Vireo::TypedArrayCore::ResizeDimensions(int, int*, bool)]@dist/vireo.js:94684:9
  __ZN5Vireo14TypedArrayCore8Resize1DEi [Vireo::TypedArrayCore::Resize1D(int)]@dist/vireo.js:6570:8
  _EggShell_ReadValueString@dist/vireo.js:17078:4
  applyVireoEmscriptenModule/asm._EggShell_ReadValueString@dist/vireo.js:127542:8
  ccallFunc@dist/vireo.js:458:15
  cwrap/<@dist/vireo.js:482:14
  assignEggShell/publicAPI.eggShell.readJSON@source/io/module_eggShell.js:114:20
  assignEggShell/</publicAPI.eggShell[methodName]@source/io/module_eggShell.js:178:27
  createVIPathParser/<@test-it/karma/utilities/testHelpers.vireoRunner.js:83:31
  @test-it/karma/http/httpget.js:141:20
  runExecuteSlicesAsync@test-it/karma/utilities/testHelpers.vireoRunner.js:73:21
```

Talking to @sanmut it seems like a possible culprit is the static variable in CEntryPoint.cpp in the EggShell_ReadValueString function: https://github.com/rajsite/VireoSDK/blob/6cc892c615f1d88ea02fb69d5074b46ae2f793a0/source/core/CEntryPoints.cpp#L176

But we have not investigated it too much further.

While creating a new Vireo instance between tests is not a big deal while the HTTP test suite is small, as it grows it will become a more noticeable test performance hit

### LINKED ISSUE #275: Calling readJSON between reboots causes abort
Static allocated string in EggShell_DataReadString and EggShell_DataWriteString is likely causing aborts() when Vireo is rebooted or re-run using a new user shell.

Possibly relevant source locations: 
   [Location where static string allocation occurs in scope of type manager reference?](https://github.com/ni/VireoSDK/blob/a8df6f39d23f04e55b05faf392260e3dcf17f8d2/source/core/CEntryPoints.cpp#L180)
   [Location where type manager reference is passed on call](https://github.com/ni/VireoSDK/blob/a8df6f39d23f04e55b05faf392260e3dcf17f8d2/source/io/module_eggShell.js#L140)
   [Location where type manager is deleted on reboot](https://github.com/ni/VireoSDK/blob/a8df6f39d23f04e55b05faf392260e3dcf17f8d2/source/io/module_eggShell.js#L118)

In the following example the first visit to endpoint eightball is successful and the second visit results in an abort.

This does not effect the vtr suite because it relies only on console io and not readJSON and writeJSON.
The karma test suite has had this issue before and the workaround is to create a new instance of the vireo runtime for every test.

In the express app creating a new runtime for every request slows response time by about 100ms (from 60ms sharing instances and just rebooting to 160ms creating a whole new instance)

```javascript
// server.js
// calls a web vi using the vireo runtime
// this example assumes the vi runs once per http request
// as such vireo is rebooted for every request

// init project
var express = require('express');
var app = express();
var fs = require('fs');

// load library to measure response time (optional)
var responseTime = require('response-time');
app.use(responseTime());

// load the Vireo Constructor Function
var Vireo = require('vireo');

// create a new Vireo instance
var vireo = new Vireo();

// register some logging functions (only used for debugging)
vireo.eggShell.setPrintFunction(console.log);
vireo.eggShell.setPrintErrorFunction(console.error);

// load the via source code
var via = fs.readFileSync('eightball.via.txt', "utf8");
var viName = 'Main%2Egviweb';

// http://expressjs.com/en/starter/static-files.html
app.use(express.static('public'));

// http://expressjs.com/en/starter/basic-routing.html
app.get("/", function (request, response) {
  response.sendFile(__dirname + '/views/index.html');
});

app.get("/eightball", function (request, response) {
  // reboot load the via code into the runtime
  vireo.eggShell.reboot();
  vireo.eggShell.loadVia(via);
  
  // take the input parameter of the query string and write it to the Input control
  var input = JSON.stringify(request.query.input);
  vireo.eggShell.writeJSON(viName, 'dataItem_Input', input);
  
  // run the via code and allow it to complete asynchronously
  (function runExecuteSlicesAsync() {
    var output;
    var status = vireo.eggShell.executeSlices(1000);
    
    if (status > 0) {
      setImmediate(runExecuteSlicesAsync);
    } else {
      output = JSON.parse(vireo.eggShell.readJSON(viName, 'dataItem_Output'));
      response.send(output);
    }
  }());
});
  
// listen for requests :)
var listener = app.listen(process.env.PORT, function () {
  console.log('Your app is listening on port ' + listener.address().port);
});
```

The via code
```
// Autogenerated Vireo assembly file
// Generated on 03/31/2017 11:57:28
define (g0 dv(a(.String *) ('It is certain' 'It is decidedly so' 'Without a doubt' 'Yes definitely' 'You may rely on it' 'As I see it, yes' 'Most likely' 'Outlook good' 'Yes' 'Signs point to yes' 'Reply hazy try again' 'Ask again later' 'Better not tell you now' 'Cannot predict now' 'Concentrate and ask again' 'Don\'t count on it' 'My reply is no' 'My sources say no' 'Outlook not so good' 'Very doubtful')))
//VI Definition: Main.gviweb
define (Main%2Egviweb dv(.VirtualInstrument (
    Locals: c(   // Data Space1
        e(.Int32 local1)
        e(.Int32 local2)
        e(.Double local3)
        e(.Double local4)
        e(.Double local5)
        e(.Int32 local6)
        e(.String local7)
        e(dv(.String '')dataItem_Output)
        e(dv(.String '')dataItem_Input)
        e(.Double local10)
    )
        clump(1
        ArrayLength(g0 local1 )
        Decrement(local1 local2 )
        Convert(local2 local3)
        Random(local4 )
        Mul(local4 local3 local5)
        Convert(local5 local6)
        ArrayIndex(g0 local7 local6 )
        Copy(local7 dataItem_Output)
/* Clump Ended. */    )
)))
enqueue (Main%2Egviweb)
//Finished!! :D
```

output
```
//Allocations    3, AQCount    82, ShareTypes 0 (ES Delete end)
1:05 AM
5
1:05 AM
5
1:05 AM
abort(5) at Error
1:05 AM
    at jsStackTrace (/app/node_modules/vireo/dist/vireo.js:1:16679)
1:05 AM
    at stackTrace (/app/node_modules/vireo/dist/vireo.js:1:16862)
1:05 AM
    at abort (/app/node_modules/vireo/dist/vireo.js:17:7679)
1:05 AM
    at Array.CK (/app/node_modules/vireo/dist/vireo.js:10:81651)
1:05 AM
    at OB (/app/node_modules/vireo/dist/vireo.js:6:23654)
1:05 AM
    at hd (/app/node_modules/vireo/dist/vireo.js:9:48893)
1:05 AM
    at ccallFunc (/app/node_modules/vireo/dist/vireo.js:1:9506)
1:05 AM
    at /app/node_modules/vireo/dist/vireo.js:1:9812
1:05 AM
    at Object.Module.eggShell.readJSON.publicAPI.eggShell.readJSON [as readJSON] (/app/node_modules/vireo/source/io/module_eggShell.js:125:20)
1:05 AM
Jump to
at runExecuteSlicesAsync (/app/server.js:54:42)
1:05 AM
If this abort() is unexpected, build with -s ASSERTIONS=1 which can give more information.
```

rajsite:
This PR performs several tasks:

- Fixes readJSON for several types. Using `vireo.eggShell.readJSON` is a valuable tool for quickly inspecting Vireo's memory during debug.

    However, the stringification of several types was generating invalid JSON. This change modifies those type / value serializations to generate valid JSON

- Removes usage of static variables from the eggshell. This would save pointers coupled to a Vireo EggShell session. When reboot is called and those sessions were destroyed, the next access frequently resulted in unexpected behavior. Fixes #275 Fixes #163 

- Because of the fix to remove static variable usage, tests were able to change from creating a vireo instance per spec to creating an instance per suite. This changes test time in Chrome from ~1min30sec to ~30sec (which reduces overall CI times by about 10-15%) 🎉 

- Fixes some vulnerabilities devDependencies by updating to the latest versions

- Cleans up vireo instances in test suites. Running karma would use excess of 1GB of memory after finishing test execution. This was due to the tests holding onto Vireo instances unnecessarily. Added steps to clean-up memory usages after clean-up. Now after a test run Chrome hold on to about 30MB of memory.


test-it/karma/utilities/TestHelpers.VireoRunner.js
```diff
@@ -62,11 +62,14 @@
                 });
             }
 
-            var complete = function () {
+            vireo.eggShell.executeSlicesUntilClumpsFinished().then(function () {
+                var noop = function () {
+                    // intentionally blank
```
cglzaguilar: Please add more rationale as to why this section is intentionally blank.

rajsite: Added a comment describing why the print functions were cleared

PR #623, opened 2019-06-04
 'ReinitializeToDefaultData' API

basant26:
As per discussion in [this RFC](https://ni.visualstudio.com/DevCentral/_git/ASW/pullrequest/26201?_a=files) , Implemented 'ReinitializeToDefaultData' API in Vireo. This API will reinitialize the given dataRef to the default value of typeRef backing it.

Testing: I have added a test for just numeric control, I will add more tests similar to this if the way I have written the test gets approved.  
What test I have added is similar to `JavaScriptInvoke` tests where I am calling a custom function with `JavaScriptInvoke`. The custom function will call reinitialize to default on the valueRef.

Any suggestions for improving the test? 
I have also seen tests which match the output of the Via containing print statements with the expected output defined in vtr files.


source/core/CEntryPoints.cpp
```diff
@@ -154,6 +154,36 @@ VIREO_EXPORT EggShellResult EggShell_DeallocateData(TypeManagerRef tm, const Typ
 
     return kEggShellResult_Success;
 }
+
+//------------------------------------------------------------
+//! ReinitializeToDefault
```
cglzaguilar: Please update the comment to something likes this:
Clears data and reinitializes to default in dataRef described by typeRef

cglzaguilar:
@rajsite, please discuss with @basant26 about the pattern for the test. I don't think you need to write VIA files with print statements.

Once that has been solved, please add more tests with different data types.

source/core/CEntryPoints.cpp
```diff
@@ -154,6 +154,36 @@ VIREO_EXPORT EggShellResult EggShell_DeallocateData(TypeManagerRef tm, const Typ
...
+VIREO_EXPORT EggShellResult EggShell_ReinitializeToDefaultData(TypeManagerRef tm, const TypeRef typeRef, void* dataRef)
```
rajsite: Let's keep the name consistent as ReinitializeToDefault

source/include/CEntryPoints.h
```diff
@@ -42,6 +42,7 @@ VIREO_EXPORT Int32 EggShell_PokeMemory(TypeManagerRef tm, const char* viName, co
                                        Int32 bufferSize, char* buffer);
 VIREO_EXPORT EggShellResult EggShell_AllocateData(TypeManagerRef tm, const TypeRef typeRef, void** dataRefLocation);
 VIREO_EXPORT EggShellResult EggShell_DeallocateData(TypeManagerRef tm, const TypeRef typeRef, void* dataRef);
+VIREO_EXPORT EggShellResult EggShell_ReinitializeToDefaultData(TypeManagerRef tm, const TypeRef typeRef, void *dataRef);
```
rajsite: Update name here too

source/io/module_eggShell.js
```diff
@@ -155,6 +155,15 @@ var assignEggShell;
             }
         };
 
+        Module.eggShell.reinitializeToDefaultData = publicAPI.eggShell.reinitializeToDefaultData = function (valueRef) {
```
rajsite: Rename to `reinitializeToDefault`

test-it/karma/javascriptinvoke/Reinitialize.Test.js
```diff
@@ -0,0 +1,43 @@
...
+        };
+    });
+
+    afterEach(function () {
+        // Cleanup functions
+        window.NI_ConcatenateValue = undefined;
+    });
+
+    it('with no parameters succesfully works', function (done) {
+        var runSlicesAsync = vireoRunner.rebootAndLoadVia(vireo, jsComplexFunctionsViaUrl);
```
rajsite: I think using the JavaScriptInvoke is adding unnecessary complexity to the test.

It should be structured like:
```
it {
reebootandloadvia
run the via to completion
expect preconditions before reinit is called
execute the reinitialize to default api function
expect postconditions after reinit is called
}
```

For new tests we should also be using the [async based runner](https://github.com/ni/VireoSDK/blob/master/test-it/karma/systemlogging/SystemLogging.Test.js#L33). Note that the `it` statement is an `async function`, that we call `await runSlicesAsync()`, and that we use let / const instead of var. (Vireo has a couple steps left in its es6 transition :P)

test-it/karma/javascriptinvoke/Reinitialize.Test.js
```diff
@@ -0,0 +1,43 @@
+describe('A JavaScript function invoke', function () {
+    'use strict';
+    // Reference aliases
+    var vireoHelpers = window.vireoHelpers;
+    var vireoRunner = window.testHelpers.vireoRunner;
+    var fixtures = window.testHelpers.fixtures;
+
+    var vireo;
+
+    var jsComplexFunctionsViaUrl = fixtures.convertToAbsoluteFromFixturesDir('javascriptinvoke/ReinitializeToDefault.via');
```
rajsite: This PR is not exercising the JavaScriptInvoke feature but is instead adding a new public api function.

This fixture should be placed in `test-it\karma\fixtures\publicapi\ReinitializeToDefault.via`

test-it/karma/fixtures/javascriptinvoke/ReinitializeToDefault.via
```diff
@@ -0,0 +1,22 @@
+define (MyVI dv(.VirtualInstrument (
+    Locals: c(   // Data Space
+        ce(dv(.Double 380)c1)
+        e(dv(.Double 1280)dataItem_Numeric)
```
sanmut: As mentioned in the conversation, please add more test cases for different data types (controls). Would like to take a look at it with more tests.

source/core/CEntryPoints.cpp
basant26: Done.

rajsite:
@basant26 Looks great! I'll make you a release to merge. Make sure to add yourself to AUTHORS in a separate PR :) https://github.com/ni/VireoSDK/blob/master/AUTHORS

rajsite:
Waiting on @cglzaguilar to review changes

PR #627, opened 2019-06-21
Fix Queue prims to assume refnum pointer won't be null

sanmut:
Also, fix behavior of output variables of Queue prims for invalid queue reference input and/or error input

### DIFF
--- a/source/core/Synchronization.cpp
+++ b/source/core/Synchronization.cpp
@@ -686,20 +686,22 @@ VIREO_FUNCTION_SIGNATURE7(QueueRef_Obtain, TypeCommon, RefNumVal, Int32, StringR
     RefNum refnumVal = 0;
     QueueRef queueRef = nullptr;
 
-    VIREO_ASSERT(refnumPtr != nullptr);
     if (name && name->Length() == 0)
         name = nullptr;
     ErrorCluster *errPtr = _ParamPointer(6);
     if (errPtr && errPtr->status) {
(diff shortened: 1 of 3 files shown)


sanmut:
> The error handling refactor looks good and is more readable.
> What prompted the change to not allow null refnums, though? What happens when you pass * now that refnums are not allowed to be null? Doing an assert implies that this should never happen, but I don't see any code to disallow passing * when parsing the via file to ensure that.

Due to a silly bug in DFIR emission we were emitting wild card (*) for queue operations primitives (such as enqueue, dequeue) in certain use cases. I am fixing that in DFIR. That's what prompted me to look at this and make the intent and interface keep it in sync with G prims. I am adding ASSERTS as we are running our tests against Vireo Native Debug and it should help with pinpointing failures easily. 

I think it would be nice to express the intent of "required" vs "optional" in VIA for prims and SubVI calls. If we do that, we can throw a parse error where wildcard is passed at the callsite where callee expects a required argument. It is sort of a feature work (could be someone's prime day effort :-) ) . As not wiring required arguments in G is a SA error, we will not run into this in generated VIA from G using NXG. So, an assert is a good enough stop-gap. I think adding extra error handling for this case until we implement "required" vs "optional" parameters in VIA is duplicated throw away effort we need not pay now.

BTW, I fixed and pushed a commit to still keep queue reference parameter for ObtainQueue as optional as it is after all an output parameter for that prim.

PR #628, opened 2019-06-26
VariantToData:Fix float to integral conversion

sanmut:
When VariantToData is used to convert a variant of float to integer, it was not following G's general coercion rules. Fixing it to match the coercion rules and added some tests.

### DIFF
--- a/source/core/DualTypeConversion.cpp
+++ b/source/core/DualTypeConversion.cpp
@@ -94,25 +94,24 @@ namespace Vireo
         return true;
     }
 
-    template<typename X, typename Y, typename Z> Y ConvertFromEnum(Z valueX, TypeRef typeRefY)
+    template<typename TSource, typename TDest> TDest ConvertFromEnum(TSource src, TypeRef destTypeRef)
     {
-        X integerValueX = static_cast<X>(valueX);
-        Y unsignedIntegerValueX = integerValueX >= 0 ? integerValueX : 0;
-        Y numElementsInY = static_cast<Y>(typeRefY->GetEnumItemCount());
-        if (unsignedIntegerValueX >= numElementsInY) {
-            unsignedIntegerValueX = numElementsInY - 1;
-        }
-        return unsignedIntegerValueX;
+        src = src >= 0 ? src : 0;
+        auto destEnumCount = static_cast<TDest>(destTypeRef->GetEnumItemCount());
+#pragma warning(push)
+#pragma warning(disable : 4018)  // Warning C4018 '>=': signed / unsigned mismatch
+        // It is safe to disable this warning because 'src' should be >= 0 here.
+        TDest dest = (src >= destEnumCount) ? destEnumCount - 1 : static_cast<TDest>(src);
+#pragma warning(pop)
+        return dest;
     }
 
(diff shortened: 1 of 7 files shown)


source/core/DualTypeConversion.cpp
```diff
@@ -94,25 +94,24 @@ namespace Vireo
...
-        Y numElementsInY = static_cast<Y>(typeRefY->GetEnumItemCount());
-        if (unsignedIntegerValueX >= numElementsInY) {
-            unsignedIntegerValueX = numElementsInY - 1;
-        }
-        return unsignedIntegerValueX;
+        src = src >= 0 ? src : 0;
+        auto destEnumCount = static_cast<TDest>(destTypeRef->GetEnumItemCount());
+#pragma warning(push)
+#pragma warning(disable : 4018)  // Warning C4018 '>=': signed / unsigned mismatch
+        // It is safe to disable this warning because 'src' should be >= 0 here.
```
spathiwa: Please don't put compiler-specific pragmas in the code like this unless absolutely necessary, and if you do, but #if guards to test the platform before using them.
In this case, I think it would be better to change the code to `static_cast<TDest>(src >= destEnumCount)` than to disable the warning.

source/core/DualTypeConversion.cpp
```diff
@@ -94,25 +94,24 @@ namespace Vireo
...
+        TDest dest = (src >= destEnumCount) ? destEnumCount - 1 : static_cast<TDest>(src);
+#pragma warning(pop)
+        return dest;
     }
 
-    template<typename T> void ApplyNumeric(T valueX, TypeRef typeRefY, void* pDataY)
+    template<typename T> void ApplyIntegralToNumeric(T valueX, TypeRef typeRefY, void* pDataY)
```
spathiwa: 'Integral' reads weird to me.  Why not just 'Integer'?

source/core/DualTypeConversion.cpp
sanmut: Since clang seems to ignore the pragmas, I left them in. I agree it is a good idea not to do compiler specific pragmas without guards. OTOH, casting either src or destEnumCount with other's type won't work as the C++ cast will do the normal roll-over and not stop at the target type's max value. OTOH, enum supports only UInt8,16,32, so it is safe to cast both to UInt32 to compare -- which would avoid the compiler warning. I updated the PR with this.

source/core/DualTypeConversion.cpp
sanmut: I went with "Integral" to imply Integer and enum. Enum is sort of an integral type with underlying integer. If you feel strongly about it, I can fix this to make it 'Integer'.

sanmut:
While adding more test cases to make sure the conversion works correctly, I also found IsA does not do the right thing in comparing enum type against an integer type. So, I fixed that and added tests for the same too.

source/core/DualTypeConversion.cpp
spathiwa: I don't feel that strongly, if you want to leave it, it's fine.

sanmut: I added some comments about the convention used in the file for Integer/Integral/Numeric.

sanmut:
Addressed Sid's feedback on renaming ConvertFromEnum method and leaving comment.

PR #629, opened 2019-06-26
Bug fix to StringScan to properly process inputs greater than 255 characters

cglzaguilar:
Correctly implemented the intention of the original programmer to just pass a fixed size buffer to just parse one number.

Added test to validate this behavior.


test-it/ViaTests/StringScanInputSizeBug.via
```diff
@@ -0,0 +1,92 @@
+// Autogenerated Vireo assembly file
```
cglzaguilar: Here is a picture of the source VI:
StringScanInputSizeBug

source/include/StringUtilities.h
```diff
@@ -385,6 +385,11 @@ class TempStackCString : public FixedCArray<Utf8Char, kTempCStringLength>
         return FixedCArray::Append(str->Begin(), (size_t)str->Length());
     }
 
+    //! Append a SubString up to lenght.
```
spathiwa: typo in comment

test-it/ViaTests/StringScanInputSizeBug.via
sanmut: Please assert all the output indicators against expected values in writing tests like this.

source/core/NumericString.cpp
```diff
@@ -1496,7 +1496,10 @@ Boolean TypedScanString(SubString* inputString, IntIndex* endToken, const Format
         }
         in.AliasAssign(in.Begin(), in.Begin() + std::min(in.Length(), formatOptions->MinimumFieldWidth + leadingSpace));
     }
-    truncateInput.Append(&in);
+    if (in.Length() <= truncateInput.Capacity())
```
sanmut: `if (in.Length() <= truncateInput.Capacity())`
It does not seem like a good idea to check for something that Append(&in) already does and returns false. I would do something like this:
`
    if (!truncateInput.Append(&in))
        truncateInput.Append(&in, truncateInput.Capacity());
`
Or, even something like this:
`
    truncateInput.Append(&in, std::min(in.Length(), truncateInput.Capacity()));
`

source/core/NumericString.cpp
```diff
@@ -1496,7 +1496,10 @@ Boolean TypedScanString(SubString* inputString, IntIndex* endToken, const Format
...
+        truncateInput.Append(&in);
+    else
+        truncateInput.Append(&in, truncateInput.Capacity());
```
sanmut: I think there is another issue with using the truncated input string. It looks like we are not counting for the fact that we are operating on the truncated string for adjusting "endPointer" after finishing the scan. When we get a string representing double or complex double (especially over the wire) that may be longer than 255 characters, we should calculate the endPointer correctly so that "offset past scan" and "remaining string" will be correct. I think we should use this opportunity to fix this issue.

I would write a test that uses an input string "75.123........with 275 digits of precision or so, 45.8" and try to scan using "%f" and also using "%f, %f" (we should scan 45.8 correctly for the second double output) and make sure all the output parameters have the correct expected values.

cglzaguilar:
Abandoning this PR in favor of a new one coming up soon.

PR #632, opened 2019-07-12
Fix to CAR 735528 Scan from String function displays wrong result if string exceeds 255 elements

cglzaguilar:
After some discussions, we decided on actually copying the whole buffer to guarantee numbers are correctly parsed.


test-it/ViaTests/ScanFromLongString.via
```diff
@@ -0,0 +1,284 @@
+// Autogenerated Vireo assembly file
```
cglzaguilar: WebVI is basically calling the following code:

Remaining

NotRemaining

source/include/DataTypes.h
```diff
@@ -353,5 +353,96 @@ class FixedCArray : public SubVector<T>
...
+
+    //! Return the maximum capacity of the array.
+    IntIndex Capacity() { return _maxSize; }
+
+    //! Return a reference to the indexed element in the vector (no range checking).
+    const T&  operator[] (const int i) { return _buffer[i]; }
+
+    //! Append an element to the array if there is room.
+    Boolean Append(T element) {
+        IntIndex i = this->Length();
```
sanmut: nitpick: I would rename "i" to "length" or something. "i" is typically used as index.

source/include/DataTypes.h
```diff
@@ -353,5 +353,96 @@ class FixedCArray : public SubVector<T>
...
+        if (i <= _maxSize) {
```
sanmut: I would replace _maxSize with Capacity() for clarity.

source/include/DataTypes.h
```diff
@@ -353,5 +353,96 @@ class FixedCArray : public SubVector<T>
...
+            _buffer[i] = element;
+            this->_end++;
+            _buffer[this->Length()] = (T)0;
```
sanmut: As done in other places, I would replace this line "_buffer[this->Length()] = (T)0;" with *NonConstEnd() = (T)0;"

source/include/DataTypes.h
cglzaguilar: Done

source/include/DataTypes.h
cglzaguilar: Done.

source/include/DataTypes.h
cglzaguilar: Done.

PR #633, opened 2019-07-17
Implement Name() on Variant data type

siddhukrs:
`VariantType` derives from `TypeCommon`, which does not have a default implementation for `Name()`. 

The `Name()` function is used in a number of different locations for many different reasons. One such usage is in the `Convert` instruction [here](https://github.com/ni/VireoSDK/blob/6c17b903c3dc459a893426cdd6bc96a9da95d579/source/core/GenericFunctions.cpp#L538). Here, we check that the `Name()` is non-empty before calling `IsA`. This is because `IsA` relies on the `TypeRef->Name()` to compare types.

Consider the case of a `Waveform<Int32>` being converted for `Waveform<Int64>`. Since Waveforms are treated as Clusters in VIA, we convert them element by element. When converting the attributes element, which of type Variant, we incorrectly skip past the code where we should emit a `Copy` instruction (since the source and destination types are the same). Instead, we emit a `DataToVariant` instruction based on the destination type being a Variant. `DataToVariant` cannot handle Variant to Variant conversions, so we crash.

Implementing `Name()` on the Variant data type will correctly allow us to emit a `Copy` instruction as we do with all other data types.


source/include/Variants.h
```diff
@@ -33,6 +33,7 @@ class VariantType : public TypeCommon
     NIError InitData(void* pData, TypeRef pattern = nullptr) override;
     NIError CopyData(const void* pData, void* pDataCopy) override;
     NIError ClearData(void* pData) override;
+    SubString Name() override { return SubString("Variant"); }
```
sanmut: It is probably a good idea to single source the string "Variant" into a constant/macro and use it here and this other place: 
https://github.com/ni/VireoSDK/blob/c6a28aa2f6c9c61cbd2e5bd2a59db8450927f46f/source/core/TypeDefiner.cpp#L250

sanmut: I think you may be able to use this directly: https://github.com/ni/VireoSDK/blob/c6a28aa2f6c9c61cbd2e5bd2a59db8450927f46f/source/core/TypeAndDataManager.cpp#L608

siddhukrs: Yup, that is what I did.

PR #634, opened 2019-07-17
Fix to CAR 741599: Waveform array wired to Search 1D Array primitive throws an exception in Vireo

cglzaguilar:
We forgot to implement IsEQSearchVariant. The search code looks for this function.


test-it/ViaTests/Search1DArrayVariant.via
```diff
@@ -0,0 +1,94 @@
+// Autogenerated Vireo assembly file
```
cglzaguilar: Search1DArrayVariant

PR #637, opened 2019-07-31
JSON null value field support

rajsite:
Fix for [CAR 736417](http://nippm.natinst.com/itg/web/knta/crt/RequestDetail.jsp?REQUEST_ID=736417) / [WorkItem 162985](https://ni.visualstudio.com/DevCentral/_workitems/edit/162985/) where fields of a cluster with a `null` JSON value were incorrectly reporting as error when use defaults for null was enabled.

Example error case from CAR:
 image

image

Fixed the specific case and added tests to cover more situations.


test-it/ViaTests/UnflattenJSONWithNullValues.via
```diff
@@ -0,0 +1,193 @@
+define (MyVI dv(.VirtualInstrument (
```
sanmut: Extensive testing looks good. I would also add a few tests where "path" is also passed. If added in the VI test suite, that is good enough.

rajsite: I'll add it to the VI test suite

PR #638, opened 2019-08-09
Update unicode character encoding

siddhukrs:
There seems to be a mismatch between the encoding algorithms used by NXG and Vireo. This change makes them behave identically.

Specifically, the first character of a VI's name was being encoded for all non alphanumeric characters. This isn't done on the NXG side. This mismatch causes errors in the peeker/poker code while trying to read values. (see NXG equivalent [here](https://ngsourcebrowser:4110/#NationalInstruments.LabVIEW.VireoTarget/Compiler/Emit/DataBuilder.cs,396))

Relying on existing tests on the Vireo side since to catch regressions. Adding integration test on the ASW side to assert that the issue is fixed.

### DIFF
--- a/source/core/vireo.loader.staticHelpers.js
+++ b/source/core/vireo.loader.staticHelpers.js
@@ -8,7 +8,7 @@ const encodeIdentifier = function (str) {
         ch = str.charAt(0);
 
     // First character must be encoded if is not a letter [A-Za-z]
-    if (!(codePoint >= 0x41 && codePoint <= 0x5A) && !(codePoint >= 0x61 && codePoint <= 0x7A)) {
+    if (!(codePoint >= 0x41 && codePoint <= 0x5A) && !(codePoint >= 0x61 && codePoint <= 0x7A) && !(codePoint > 0x7F)) {
         encoded += '%' + codePoint.toString(16).toUpperCase();
     } else {
         encoded += ch;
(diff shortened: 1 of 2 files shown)


source/core/vireo.loader.staticHelpers.js
```diff
@@ -7,8 +11,8 @@ const encodeIdentifier = function (str) {
         codePoint = str.charCodeAt(0),
         ch = str.charAt(0);
 
-    // First character must be encoded if is not a letter [A-Za-z]
-    if (!(codePoint >= 0x41 && codePoint <= 0x5A) && !(codePoint >= 0x61 && codePoint <= 0x7A)) {
+    // First character must be encoded if is not a letter [A-Za-z] or a non-ascii character
+    if (!isCharacterAlphabetOrNonAscii(codePoint)) {
```
siddhukrs: ESLint thinks my code's cyclomatic complexity is high. Broke it out into a function. 🙄

source/core/vireo.loader.staticHelpers.js
```diff
@@ -1,3 +1,7 @@
+var isCharacterAlphabetOrNonAscii = function (codePoint) {
```
siddhukrs: Can this stay a `var` since it isn't being exported?

source/core/vireo.loader.staticHelpers.js
rajsite: Looks like a good clean code change to me. With a self-documenting function name, the comment isn't really useful anymore if you want to get rid of it.

rajsite:
Can you add some tests to https://github.com/ni/VireoSDK/blob/master/test-it/karma/static/VireoCodec.Test.js

Particularly around testing the behavior of encoding the first character (make sure to include some cjk and surrogate pair emoji like 💩).

PR #639, opened 2019-08-13
Boolean input using default format causes error in StringFormat (CAR 742708)

cglzaguilar:
Passing a boolean input into StringFormat with an empty format string (default) generate the following error:

Format Into String (arg 2) in WebApp::index.gviweb<APPEND>
Invalid format specifier: "%Z"

Also, returned values should be in uppercase: TRUE or FALSE.

See CAR 742708


test-it/ViaTests/StringFormatDefaultFormat.via
```diff
@@ -0,0 +1,69 @@
+// Autogenerated Vireo assembly file
```
cglzaguilar: FormatStringDefaultFormat

cglzaguilar: FormatStringDefaultFormat

cglzaguilar: FormatStringDefaultFormat

source/core/NumericString.cpp
```diff
@@ -418,7 +422,7 @@ static Int32 GetTimeZoneOffsetFromTimeAndLocale(StaticTypeAndData *arg, const Fo
 /**
  * main format function, all the %format functionality is done through this one
  * */
-void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], StringRef buffer, ErrorCluster *errPtr)
+void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], StringRef buffer, ErrorCluster *errPtr, SubString* formatName)
```
sanmut: You may want to make the new parameter being added as a default (optional) parameter with default value of nullptr. This will help avoid fixing all the existing callers.

test-it/ViaTests/StringFormatDefaultFormat.via
```diff
@@ -0,0 +1,69 @@
...
+// Generated on 08/09/2019 13:59:40
```
sanmut: Please consider adding a test case where %s is being passed in the "format string" for the corresponding boolean arguments. As this was also incorrect before with respect to casing, it is worth capturing in a test.

test-it/ViaTests/StringFormatDefaultFormat.via
```diff
@@ -0,0 +1,103 @@
+// Autogenerated Vireo assembly file
```
cglzaguilar: FormatStringDefaultFormat

source/core/NumericString.cpp
cglzaguilar: Done

test-it/ViaTests/StringFormatDefaultFormat.via
cglzaguilar: Done

PR #640, opened 2019-08-16
Fix Queue bugs, refactor, and add tests

sanmut:
Fix memory corruption issues with some sequences of enqueue/dequeue.

- Deleted the old static queue related APIs which help simplified the implementation a bit more to handle only the supported dynamic Queue. Two pointers (front and back) are used for both growable queue and bounded queue

- Use circular array under the hood. Array is resized only when it is full

- Added various ASSERTs about the assumptions the implementation must satisfy.

- Added new test for a failure case that caused corruption


source/core/Synchronization.cpp
```diff
@@ -452,123 +452,174 @@ VIREO_FUNCTION_SIGNATURE1(SetOccurrence, OccurrenceRef)
...
+            _back = insert;
+        } else {  // Array full
+            VIREO_ASSERT(_count == _elements->Length());
+            if (_maxSize > 0 && _count == _maxSize)
+                return false;
+            if (insert == 0) {
+                // Just an optimization to grow the underlying array at the end than shifting all the elements by 1
+                insert = _elements->Length();
+            }
+            NIError err = _elements->Insert1D(insert, 1);
```
spathiwa: Maybe consider adjusting the capacity here two grow 'smartly'?  Double the number of elements until you get to some large cap (1M?) and then become additive.  I just checked and noticed that the ResizeCapacity member of TypedArrayCore is private, though, so this might require more work and I think it's okay to put it off with a TO-DO and issue files.

sanmut: Good point. Created an issue to address this later.

PR #642, opened 2019-08-22
Fix bug in returning Queue name

sanmut:
Get Queue Status and Release Queue APIs intermittently have been failing in returning the queue name correctly for named queues. Especially, this issue would show up either in AppVeyor or Travis CI but not both for the same PR very rarely and just requeuing  a build would not cause the failure to show up. 

Fixed the bug in the function that returns queue name to correctly do so even if the queue refnum is 0.

Also, added a stress test to obtain, get status, release in a loop for some iterations as this is likely to catch similar issues including memory leaks more reliably.


source/core/Synchronization.cpp
```diff
@@ -948,7 +944,7 @@ VIREO_FUNCTION_SIGNATURE9(QueueRef_GetQueueStatus, RefNumVal, Boolean, Int32, St
     if (_ParamPointer(2))
         _Param(2) = maxSize;
     if (_ParamPointer(3))  // name
-        GetQueueRefName(refnumPtr ? refnumPtr->GetRefNum() : 0, _ParamPointer(3), false);
+        GetQueueRefName(refnumPtr->GetRefNum(), _ParamPointer(3), false);
```
sanmut: This is just a minor cleanup (unrelated to the bug fix)

sanmut:
@spathiwa , can refnum value be zero for Obtain Queue to return when it is successful for "named queues" especially? If it is, then this fix seems right. If it is not, then I have to debug further to see why refnum value is getting a zero in Obtain Queue to begin with?

spathiwa:
Adding clarifying comments after our in person discussion:
The problem is that making a new refnum alias is returning 0 as a new refnum value (by wrapping a static index), when 0 should be reserved for kNotARefnum.
Normal refnums already skip zero; this is done with the code
`    if (MagicMask(++_nextMagicNum) == 0)
        _nextMagicNum = 1;
`
in RefNumStorageBase::NewRefNum().

I think the correct fix is for RefNumStorageBase::CloneRefNum() (used by NewRefnumAlias) to also have similar logic

sanmut:
Will reopen this when I get a chance to prioritize this.

PR #645, opened 2019-09-12
Vireo data item needs update implementation

rajsite:
Adds a needsUpdate flag to all parameters and the locals marked as data items in Vireo. With this change the HTML Panel is able to ask Vireo if a control associated with a variable needs an update and can avoid reading the value if there is no change.

The outcome of this change is controls with large values that are not updated frequently can avoid unnecessary work if there are no writes to their data item. The following capture demonstrates the reduction in cpu time by avoiding unnecessary work for large unchanging data items:

Example diagram showing a large number of values written once to a control:
diagram

Before needsUpdate is implemented we try to read the graphs terminal every frame even if there is no change:
before

After needsUpdate we can avoid reading the graph terminal if the diagram has not copied a new value to the data item:
after

This builds on the work of @spathiwa in https://github.com/ni/VireoSDK/pull/590 by adding some JS tests and settling on the term needsUpdate consistently.


sanmut:
> After needsUpdate we can avoid reading the graph terminal if the diagram has not copied a new value to the data item:
Sweet. Thanks for showing the performance differences with the change.

source/include/TypeAndDataManager.h
```diff
@@ -508,21 +508,23 @@ class TypeCommon
...
     //  properties unique to prototype elements. they are never merged up
-    UInt16  _elementUsageType:3;  // (7-9) ElementType::UsageType
+    UInt16  _elementUsageType:3;  // (12-14) ElementType::UsageType
+    UInt16  _isDataItem:1;        // (15) Element keeps track of updatedNeeded state for read/writes
+    UInt16  _needsUpdate:1;      // (0) Value has been written, needs display (tested by JS)
     //  properties unique to DefaultPointerType objects
-    UInt16  _pointerType:3;       // (10-12)
-    UInt16  _ownsDefDefData:1;    // (13) Owns DefaultDefault data (clusters and arrays)
-    UInt16  _opaqueReference:1;   // (14) Data is not an instance of the type it wraps or is templated from (e.g. refnum(Queue))
+    UInt16  _pointerType:3;       // (1-3)
```
sanmut: Great, thanks for fixing these offsets in the comment.

test-it/ViaTests/CheckValueNeedsUpdateTest.via
```diff
@@ -0,0 +1,60 @@
+
+define (MyVI1 dv(.VirtualInstrument (
+    Locals: c(
+        de(.Double myDouble)
```
sanmut: Add a test for an item that is not marked "de" and written to and make sure CheckValueNeedsUpdate returns false.

sanmut: Add a test for a cluster for the following cases:
1. cluster marked "de" and individual elements marked "e" and
        - write to the whole cluster and CheckValueNeedsUpdate
        - write to an element in the cluster and CheckValueNeedsUpdate
2. cluster marked "e" and some elements marked "de" and some marked "e"
        - write to the whole cluster and CheckValueNeedsUpdate
        - write to an "e" element in the cluster and CheckValueNeedsUpdate
        - write to an "de" element in the cluster and CheckValueNeedsUpdate
3. Write to an element of a data item in a Vireo data type such as ErrorCluster (or preferably AnalogWaveform) in VIA but internally defined as a cluster in Vireo and check for CheckValueNeedsUpdate

If some of the cases are not possible and need not be handled and tested because DFIR emission will not emit those cases, please document this somewhere (or close to where "de" is defined in the code) and in DFIR side PR.

source/include/TDCodecVia.h
```diff
@@ -257,7 +257,7 @@ void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], Strin
 #define tsInputParamToken       "i"   // input parameter
 #define tsOutputParamToken      "o"   // output parameter
 #define tsInputOutputParamToken "io"  // input-output parameter
-                                      // (output and input-output params also act as dataItems, see above)
+                                      // (input, output, and input-output params also act as dataItems, see above)
```
spathiwa: I think I didn't have input here intentionally, because you can't write to an input, so it could never be marked dirty.

source/core/TypeAndDataManager.cpp
```diff
@@ -1226,20 +1228,21 @@ TypeRef WrappedType::GetSubElementAddressFromPath(SubString* name, void *start,
...
 // ElementType
 //------------------------------------------------------------
-ElementType* ElementType::New(TypeManagerRef typeManager, SubString* name, TypeRef wrappedType, UsageTypeEnum usageType, Int32 offset)
+ElementType* ElementType::New(TypeManagerRef typeManager, SubString* name, TypeRef wrappedType, UsageTypeEnum usageType, Int32 offset, bool isDataItem)
 {
-    ElementType* type = TADM_NEW_PLACEMENT_DYNAMIC(ElementType, name)(typeManager, name, wrappedType, usageType, offset);
-
+    ElementType* type = TADM_NEW_PLACEMENT_DYNAMIC(ElementType, name)(typeManager, name, wrappedType, usageType, offset, isDataItem);
+    if (isDataItem)
+        return type;
```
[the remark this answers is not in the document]
spathiwa: Yes, it does. The early return prevents dataitems with the same name and type from being merged in the dictionary into one object, meaning they can't be checked for needs-update independently.

source/core/TDCodecVia.cpp
```diff
@@ -506,23 +506,31 @@ void TDViaParser::ParseAggregateElementList(std::vector<TypeRef> *elementTypesVe
...
         if (token.CompareCStr(tsElementToken)) {
             usageType = kUsageTypeSimple;
+        } else if (token.CompareCStr(tsDataitemElementToken)) {
+            usageType = kUsageTypeSimple;
+            isDataItem = true;
         } else if (token.CompareCStr(tsConstElementToken)) {
             usageType = kUsageTypeConst;
         } else if (token.CompareCStr(tsInputParamToken)) {
             usageType = kUsageTypeInput;
+            isDataItem = true;
```
spathiwa: @rajsite  Did you discover a reason inputs need to be dataItems?
The code which inserts the call to SetValueNeedsUpdate does it on the destination of a copy, which I think can never be an input.

rajsite: This is unchanged from the [previous PR](https://github.com/ni/VireoSDK/pull/590/files#diff-e8e3eac825feec1ddca7e0d263928ecdR527) made in [this commit](https://github.com/ni/VireoSDK/pull/590/commits/fef7e10c025ab8788aae3ff92f537bac7ceba9f9).

source/include/TDCodecVia.h
rajsite: Commented below how this was a consistency change. And yea they can't be marked as dirty but they are "dataitems" in the loose sense of how the term is used which is roughly that they are coupled to controls / indicators. 

I can update the comment to that effect.

source/core/TDCodecVia.cpp
spathiwa: Ah, ok, there's a conversation about why this change was made in that [commit](https://github.com/ni/VireoSDK/pull/590/commits/fef7e10c025ab8788aae3ff92f537bac7ceba9f9) that I had completely forgotten about.   It's all good, then.

source/core/TypeAndDataManager.cpp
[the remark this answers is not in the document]
cglzaguilar: Yes, I'm curious too as to why we don't follow through and do an early return. Please add a comment, explaining why in the code.

source/include/TypeAndDataManager.h
```diff
@@ -508,21 +508,23 @@ class TypeCommon
...
+    UInt16  _ownsDefDefData:1;    // (4) Owns DefaultDefault data (clusters and arrays)
+    UInt16  _opaqueReference:1;   // (5) Data is not an instance of the type it wraps or is templated from (e.g. refnum(Queue))
```
cglzaguilar: _opaqueReference appears to share the same offset with _isFlat, is that correct?

sanmut: _isFlat is part of the one UInt16 and _opaqueReference is part of the next UInt16

source/core/TypeAndDataManager.cpp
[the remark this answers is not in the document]
rajsite: Added a comment describing the early return

source/include/TDCodecVia.h
rajsite: Added a comment describing all params are dataitems

source/include/TypeAndDataManager.h
rajsite: It was all @spathiwa :)

source/include/TDCodecVia.h
```diff
@@ -251,10 +251,14 @@ void Format(SubString *format, Int32 count, StaticTypeAndData arguments[], Strin
 #define tsEnqueueTypeToken      "enqueue"
 #define tsElementToken          "e"   // used for Cluster, BitCluster, and array aggregate types for simple elements
 #define tsConstElementToken     "ce"  // used for Cluster elements in Locals: section to indicate immutable value
-#define tsAliasToken            "al"  // alias to another element.
+#define tsDataitemElementToken  "de"  // used for Cluster elements in Locals: section to indicate dataItem value
+                                      // (value will be marked 'needsUpdate' when written from Vireo,
+                                      // clean when read or written from JS, when accessed in a top-level VI)
```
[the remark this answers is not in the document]
rajsite: updated the comment to include fix

test-it/ViaTests/CheckValueNeedsUpdateTest.via
rajsite: Added test for value not marked `de` and not marked `de` for double, array, cluster, Error, and AnalogWaveform<Double>.

I did not write tests for sub-elements of a cluster in the dataspace being marked as dataitems as dfir gen will not emit the dataitem marker for sub-elements of a cluster. Added a comment to the `de` definition stating this. I'll add a similar comment to the ASW side.

rajsite:
Reset @sanmut and @cglzaguilar for the most recent commit with the CopyAndReset changes to support latched boolean

PR #646, opened 2019-10-04
Close reference support

cglzaguilar:
Close Reference instruction implementation

- Close reference supports closing a single reference or an array of references.
- Close reference doesn't close Static references, aka control references. It does nothing and it doesn't return an error.
- Invalid reference error is now generated when a closed reference is passed to a JavaScriptInvoke instruction
- Added tests


source/core/CloseReference.cpp
```diff
@@ -0,0 +1,97 @@
...
+                jsCloseJavaScriptRefNum(elementType, pArray->BeginAt(i), typeRefErrorCluster, errorClusterPtr);
+            }
+        } else {
+            jsCloseJavaScriptRefNum(elementType, pData, typeRefErrorCluster, errorClusterPtr);
+        }
+    }
+#endif
+
+    // Report close reference error if there is not an error already present
+    if (errorClusterPtr && !errorClusterPtr->status && !errorAlreadyPresent)
```
rajsite: We only need the `!errorAlreadyPresent` check right? AddCallChainToSourceIfErrorPresent does the other checks

test-it/karma/fixtures/javascriptinvoke/CloseReference.via
```diff
@@ -0,0 +1,111 @@
...
+        e(dv(.Boolean false) isNotAValidRefnum1)
+        e(dv(.Boolean false) isNotAValidRefnum2)
+        e(dv(.Boolean false) isNotAValidRefnum3)
+        e(dv(.Boolean false) isNotAValidRefnum4)
+        e(.Occurrence occurrence)
+    )
+    clump (
+        // no error in, valid reference, close reference => no error out, reference closed
+        JavaScriptInvoke(occurrence false getObjectReferenceError1 'NI_GetObjectFunction' objectReference)
+        CloseReference(objectReference noErrorInAndValidReferenceError)
```
rajsite: Can you add IsNotANumPathRefnum checks before CloseReference so we can be sure the reference was valid before it was closed?

rajsite:
All the karma tests could be moved to vtr tests by calling built-in browser functions.

For example calling `Math.pow` or `Date.now`.

These tests don't require any karma behaviors like setting values before the test run, complex parsing of values, etc.

It would be better to write them as vtr tests so they run in the browser and in node.js

rajsite:
@cglzaguilar also made the error handling tweaks I was talking about in this pr to your branch: https://github.com/cglzaguilar/VireoSDK/pull/1

test-it/karma/javascriptinvoke/CloseReference.Test.js
```diff
@@ -0,0 +1,146 @@
...
+            expect(viPathParser('getObjectReferenceError1.status')).toBeFalse();
+            expect(viPathParser('getObjectReferenceError1.code')).toBe(0);
+            expect(viPathParser('getObjectReferenceError1.source')).toBeEmptyString();
+            expect(viPathParser('getObjectReferenceError2.status')).toBeFalse();
+            expect(viPathParser('getObjectReferenceError2.code')).toBe(0);
+            expect(viPathParser('getObjectReferenceError2.source')).toBeEmptyString();
+            expect(viPathParser('noErrorInAndValidReferenceError.status')).toBeFalse();
+            expect(viPathParser('noErrorInAndValidReferenceError.code')).toBe(0);
+            expect(viPathParser('noErrorInAndValidReferenceError.source')).toBeEmptyString();
+            expect(viPathParser('isNotAValidRefnum1')).toBeTrue();
```
sanmut: I would arrange the asserts in the same order as they appear in .via. So, move isNotAValidRefnum1 after getObjectReferenceError1

test-it/karma/fixtures/javascriptinvoke/CloseReference.via
```diff
@@ -0,0 +1,111 @@
+define(DynamicReferencesVI dv(VirtualInstrument (
+    Locals: c(
+        ce(dv(.ErrorCluster (true 100 'error' )) errorIn)
```
sanmut: <nitpick> I would just pick a different string for source instead of 'error'. Too many error(s) already :-)

test-it/karma/fixtures/javascriptinvoke/CloseReference.via
```diff
@@ -0,0 +1,111 @@
...
+        IsNotANumPathRefnum(objectReference isNotAValidRefnum1)
+
+        // error in, valid reference, close reference => error out same as error in, reference closed
+        JavaScriptInvoke(occurrence false getObjectReferenceError2 'NI_GetObjectFunction' objectReference)
+        Copy(errorIn errorInAndValidReferenceError)
+        CloseReference(objectReference errorInAndValidReferenceError)
+        IsNotANumPathRefnum(objectReference isNotAValidRefnum2)
+
+        // no error in, invalid reference, close reference => invalid reference error
+        CloseReference(objectReference noErrorInAndInvalidReferenceError)
```
sanmut: Not that necessary here. If you want to do for completion either here or preferably in G tests, just add for JS Reference that is a constant (defined with "ce" here) for similar tests.

test-it/karma/fixtures/javascriptinvoke/CloseReference.via
```diff
@@ -0,0 +1,111 @@
...
+        e(c(e(.Boolean status) e(.Int32 code) e(.String source)) getObjectReferenceError1)
+        e(c(e(.Boolean status) e(.Int32 code) e(.String source)) getObjectReferenceError2)
+
+        e(c(e(.Boolean status) e(.Int32 code) e(.String source)) noErrorInAndValidReferenceError)
+        e(c(e(.Boolean status) e(.Int32 code) e(.String source)) errorInAndValidReferenceError)
+        e(c(e(.Boolean status) e(.Int32 code) e(.String source)) noErrorInAndInvalidReferenceError)
+        e(c(e(.Boolean status) e(.Int32 code) e(.String source)) errorInAndInvalidReferenceError)
+
+        e(.JavaScriptDynamicRefNum objectReference)
+        e(dv(.Boolean false) isNotAValidRefnum1)
```
sanmut: <nitpick> In general, this is good to follow. Double negation is typically harder to follow. Though the API is NotAValidRefnum, we could negate the output boolean and and print/assert booleans as isValidRefNum1. Or, even write a simple wrapper SubVI around IsNotANumPathRefnum that simply negates the output and returns and call the SubVI in all the places instead of IsNotANumPathRefnum

test-it/karma/javascriptinvoke/CloseReference.Test.js
```diff
@@ -0,0 +1,146 @@
...
+            expect(rawPrint).toBeEmptyString();
+            expect(rawPrintError).toBeEmptyString();
+            expect(viPathParser('getObjectReferenceError.status')).toBeFalse();
+            expect(viPathParser('getObjectReferenceError.code')).toBe(0);
+            expect(viPathParser('getObjectReferenceError.source')).toBeEmptyString();
+            expect(viPathParser('closeReferenceError.status')).toBeFalse();
+            expect(viPathParser('closeReferenceError.code')).toBe(0);
+            expect(viPathParser('closeReferenceError.source')).toBeEmptyString();
+            expect(viPathParser('isNotAValidRefnum1')).toBeTrue();
+            expect(viPathParser('invalidReferenceError.status')).toBeTrue();
```
sanmut: invalidReferenceError is likely the only thing this test really needs to assert. Do we want all the previous ones? If you think it is safe to keep those, then put them in a separate pre-condition function so that it is clear what the test is intending to assert (using expect statements here) and what the preconditions are. This feedback is applicable for the other test methods too.

test-it/karma/javascriptinvoke/CloseReference.Test.js
```diff
@@ -0,0 +1,146 @@
...
+        var viPathParser = vireoRunner.createVIPathParser(vireo, viName);
+        vireoRunner.enqueueVI(vireo, viName);
+
+        runSlicesAsync(function (rawPrint, rawPrintError) {
+            expect(rawPrint).toBeEmptyString();
+            expect(rawPrintError).toBeEmptyString();
+            expect(viPathParser('getObjectReferenceError.status')).toBeFalse();
+            expect(viPathParser('getObjectReferenceError.code')).toBe(0);
+            expect(viPathParser('getObjectReferenceError.source')).toBeEmptyString();
+            expect(viPathParser('closeReferenceError.status')).toBeFalse();
```
sanmut: Would be nice to have a wrapper to pass just 'closeReferenceError' and assert that there is no error. Asserts in the test method would be more succinct. Thoughts?

sanmut:
If possible, would be nice to validate against draft implementation of CloseReference in DFIR side before merging.

test-it/ViaTests/JSReferenceCloseArray.via
```diff
@@ -0,0 +1,28 @@
...
+
+        e(.JavaScriptDynamicRefNum objectReference1)
+        e(.JavaScriptDynamicRefNum objectReference2)
+        e(dv(.Boolean false) isNotAValidRefnum1)
+        e(dv(.Boolean false) isNotAValidRefnum2)
+        e(.Occurrence occurrence)
+    )
+    clump (
+        JavaScriptInvoke(occurrence false getObjectReferenceError 'Date.now' objectReference1)
+        JavaScriptInvoke(occurrence false getObjectReferenceError 'Date.now' objectReference2)
```
rajsite: before closing can you assert that the refnums are valid?

test-it/ViaTests/JSReferenceCloseStatic.via
```diff
@@ -0,0 +1,32 @@
...
+
+        e(.JavaScriptStaticRefNum staticReference)
+        e(dv(.Boolean false) isNotAValidRefnum1)
+        e(dv(.Boolean false) isNotAValidRefnum2)
+        e(.Occurrence occurrence)
+    )
+    clump (
+        // no error in, valid reference, close reference => no error out, reference not closed
+        JavaScriptInvoke(occurrence false getObjectReferenceError 'Date.now' staticReference)
+        CloseReference(staticReference noErrorInAndValidReferenceError)
```
rajsite: before closing assert the refnum is valid

test-it/ViaTests/JSReferenceInvalidParameter.via
```diff
@@ -0,0 +1,27 @@
...
+        e(c(e(.Boolean status) e(.Int32 code) e(.String source)) invalidReferenceError)
+
+        e(.JavaScriptDynamicRefNum objectReference)
+        e(dv(.Boolean false) isNotAValidRefnum1)
+        e(dv(.Boolean false) isNotAValidRefnum2)
+        e(.Occurrence occurrence)
+    )
+    clump (
+        JavaScriptInvoke(occurrence false getObjectReferenceError 'Date.now' objectReference)
+        CloseReference(objectReference closeReferenceError)
```
rajsite: before closing assert valid

test-it/ViaTests/CloseReferenceUnsupportedType.via
```diff
@@ -0,0 +1,13 @@
+define(CloseReferenceUnsupportedType dv(VirtualInstrument (
+    Locals: c(
+        e(.QueueRefNum<.String> queueReference)
+        e(c(e(.Boolean status) e(.Int32 code) e(.String source)) argumentError)
+    )
+    clump (
+        CloseReference(queueReference argumentError)
```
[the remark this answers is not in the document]
cglzaguilar: Yes. I wanted to have something that exercises this scenario and the associated code path.

source/core/CloseReference.cpp
cglzaguilar: Done

test-it/karma/fixtures/javascriptinvoke/CloseReference.via
cglzaguilar: Done

source/io/module_javaScriptInvoke.js
```diff
@@ -505,6 +533,22 @@ var assignJavaScriptInvoke;
...
+         */
+        Module.javaScriptInvoke.jsCloseJavaScriptRefNum = function (javaScriptRefnumTypeRef, javaScriptRefnumDataRef, errorTypeRef, errorDataRef) {
+            var javaScriptValueRef = Module.eggShell.createValueRef(javaScriptRefnumTypeRef, javaScriptRefnumDataRef);
+            var isDynamicReference = Module.typeHelpers.isJSObjectDynamicRefnum(javaScriptValueRef.typeRef);
+            if (isDynamicReference) {
+                var cookie = Module.eggShell.readDouble(javaScriptValueRef);
+                var keyWasPresent = cookieToJsValueMap.delete(cookie);
+                if (!keyWasPresent) {
+                    var errorValueRef = Module.eggShell.createValueRef(errorTypeRef, errorDataRef);
+                    mergeNewError(errorValueRef, 'CloseReference', ERRORS.kNIInvalidReference);
```
rajsite: This was more correct before. The mergeNewError helper function name is the JS function name, not the VI name.

Close should not use this helper

cglzaguilar: Done. Reverted to previous implementation.

test-it/ViaTests/JSReferenceCloseArray.via
cglzaguilar: Done.

test-it/ViaTests/JSReferenceCloseStatic.via
cglzaguilar: Done.

test-it/ViaTests/JSReferenceInvalidParameter.via
cglzaguilar: Done.

test-it/karma/fixtures/javascriptinvoke/CloseReference.via
cglzaguilar: True, but in this case I wanted to use a reference that was previously closed.

test-it/karma/javascriptinvoke/CloseReference.Test.js
cglzaguilar: The tests are not longer karma based. They are now vtr based and they don't have this problem.

test-it/karma/javascriptinvoke/CloseReference.Test.js
cglzaguilar: The tests are not longer karma based. They are now vtr based and this problem is not as bad.

test-it/karma/fixtures/javascriptinvoke/CloseReference.via
cglzaguilar: Done. Thanks for the suggestion.

test-it/karma/fixtures/javascriptinvoke/CloseReference.via
cglzaguilar: Done.

test-it/karma/javascriptinvoke/CloseReference.Test.js
cglzaguilar: Done.

test-it/karma/fixtures/javascriptinvoke/CloseReference.via
```diff
@@ -0,0 +1,111 @@
+define(DynamicReferencesVI dv(VirtualInstrument (
+    Locals: c(
+        ce(dv(.ErrorCluster (true 100 'error' )) errorIn)
+        e(c(e(.Boolean status) e(.Int32 code) e(.String source)) err)
```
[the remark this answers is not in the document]
cglzaguilar: Done.

PR #647, opened 2019-10-04
Removing old implementation of Property Node feature not longer used

cglzaguilar:


source/io/PropertyNode.cpp
```diff
@@ -1,170 +0,0 @@
...
-
-#include "TypeDefiner.h"
-#include "ExecutionContext.h"
-#include "StringUtilities.h"
-#include "TDCodecVia.h"
-#include "VirtualInstrument.h"
-#include "ControlRef.h"
-#include <stdio.h>
-
-#define VIREO_MODULE_PropertyNode 1
```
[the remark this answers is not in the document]
cglzaguilar: Nope, this is code that we dont' have intention on maintaining, we should delete it.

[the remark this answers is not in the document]
sanmut: Ok, go for it

rajsite:
Need to also remove https://github.com/ni/VireoSDK/blob/master/source/io/module_propertyNode.js

source/io/library_propertyNode.js
```diff
@@ -1,13 +0,0 @@
-(function () {
-    'use strict';
-    var propertyNodeCAPI = {
-        jsPropertyNodeWrite: function () {
-            Module.propertyNode.jsPropertyNodeWrite.apply(undefined, arguments);
```
rajsite: Need to delete the corresponding https://github.com/ni/VireoSDK/blob/master/source/io/module_propertyNode.js

rajsite: And the reference to it in https://github.com/ni/VireoSDK/blob/master/source/core/vireo.loader.shared.js#L6

cglzaguilar: Done. Thanks for pointing out this file

PR #649, opened 2019-10-28
Fixed issue with Search1DArray for JS References

cglzaguilar:
Implemented missing IsEqSearch for Static and Dynamic references.

Added test.


test-it/ExpectedResults/JSReferenceArraySearch.vtr
```diff
@@ -0,0 +1,3 @@
+dataItem_IndexOfElement: 2
+dataItem_TestResult_CorrectIndex: true
+dataItem_TestResult_NoError: true
```
sanmut: Why even calculate and print this? dataItem_TestResult_NoError

test-it/ViaTests/JSReferenceArraySearch.via
```diff
@@ -0,0 +1,72 @@
...
+        IsEQ(local9 c2 local10)
+        Copy(local10 dataItem_TestResult_NoError)
+        ArrayConcatenate(local13 local4 local6 local8 )
+        Search1DArray(local13 local8 c1 local14 )
+        Copy(local14 dataItem_IndexOfElement)
+        IsEQ(local14 c3 local17)
+        Copy(local17 dataItem_TestResult_CorrectIndex)
+        Printf ("dataItem_IndexOfElement: %z\n" dataItem_IndexOfElement)
+        Printf ("dataItem_TestResult_CorrectIndex: %z\n" dataItem_TestResult_CorrectIndex)
+        Printf ("dataItem_TestResult_NoError: %z\n" dataItem_TestResult_NoError)
```
sanmut: Please consider adding the following test cases:
1.  Call "Search 1D Array" with a valid JS Reference returned by a previous JavaScriptInvoke call as "element" but do not include in the input "1D array".  Return -1?
2.  Call "Search 1D Array" with a variable of type JS Reference but not used at all (so it is Invalid JSRef) as "element" and that is not in the input "1D array". Return -1?
3.  Call "Search 1D Array" with a variable of type JS Reference but not used at all (so it is Invalid JSRef) as "element" and include this invalid JS Ref also in the input "1D array". Return correct index?

test-it/ExpectedResults/JSReferenceArraySearch.vtr
```diff
@@ -0,0 +1,3 @@
+dataItem_IndexOfElement: 2
```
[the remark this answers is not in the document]
cglzaguilar: Yes. Removed this line.

test-it/ExpectedResults/JSReferenceArraySearch.vtr
cglzaguilar: We need to know that there were any errors with JSLI calls.

test-it/ViaTests/JSReferenceArraySearch.via
cglzaguilar: Added test cases as requested:

JSReferenceArraySearh

test-it/ExpectedResults/JSReferenceArraySearch.vtr
```diff
@@ -0,0 +1,5 @@
+dataItem_TestResult_NoError: true
+dataItem_TestResult_InvalidReferenceFound: true
+dataItem_TestResult_InvalidReferenceNotFound: true
+dataItem_TestResult_ElementNotIsArrayNotFound: true
```
sanmut: Typo dataItem_TestResult_ElementNotI**s**ArrayNotFound --> dataItem_TestResult_ElementNotI**n**ArrayNotFound

test-it/ExpectedResults/JSReferenceArraySearch.vtr
sanmut: You don't need to fix it now, as we have not had this process around this yet. I think it is better for us to start thinking about it together. It would be better to distinguish the "core asserts" that the the test is intending to assert vs some of the "pre-conditions that we want to assert". May be, we should prefix/suffix the pre-condition asserts with "_PreCondtion_".

PR #653, opened 2019-11-01
Fix to StringToNumber instructions to appropriately work with arrays greater than 1D

cglzaguilar:
Arrays greater than 1D generate an exception when passed as input to string to number conversion instructions.

The issue is that the output array is always resized to 1D. We should use the size of the input array.


test-it/ViaTests/DHOArrayStringToNumber.via
```diff
@@ -0,0 +1,101 @@
+// Autogenerated Vireo assembly file
```
cglzaguilar: Resize

sanmut:
@cglzaguilar , looks like there are build/test failures in both CIs. Would you mind looking into this and fixing? If it is an intermittent issue, try pushing a dummy commit and see?

.appveyor.yml
```diff
@@ -17,7 +17,7 @@ install:
   - pip install tox
   - ps: Start-Process npm "run httpbin" -PassThru
   - git clone https://github.com/juj/emsdk.git
-  - cd emsdk && git checkout 9538381d56272818cba77a808ba46fdac6c961c8 && cd ..
+  - cd emsdk && git checkout 1b1f08f356d3f10d91e50c09dda6d44372ca908c && cd ..
```
rajsite: Emscripten changed their binary feed urls so we had to update to a newer version of the sdk tools that use the newer feed urls.
This does not changed the emscripten compiler version, it just changes where the compiler is downloaded from.

test-it/ViaTests/DHOArrayStringToNumber.via
sanmut: In the PR to merge this, please include this G test.

sanmut: Until we add support for system configured decimal separator (such as comma instead of period), I wonder whether we should throw an internal Vireo exception for True input for "use system decimal point" boolean input. We could throw a SA error/warning in G if a constant is wired. Might be beyond the scope of the bug you are trying to fix. But, I think we should track this. Thoughts?

We may want to investigate using [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) in JS and export a function to find the system/locale specific decimal separator and use it here. Format String related nodes may also use the decimal separator.
"Format To String" nodes may also benefit

source/core/GenericFunctions.cpp
```diff
@@ -1279,7 +1289,7 @@ VIREO_FUNCTION_SIGNATURET(VectorOrClusterStrToNumOp, AggregateStrToNumInstructio
         elementSizeStr = VStr->ElementType()->TopAQSize();
         elementSizeDest = _ParamImmediate(VOutput)->_paramType->GetSubElement(0)->TopAQSize();
         count = VStr->Length();
-        VOutput->Resize1D(count);
+        ResizeOutputToSizeOfInput(VStr, VOutput);  // Comment
```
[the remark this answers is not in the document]
cglzaguilar: Yes, good point. Used VOutput->ResizeDimensions(VStr->Rank(), VStr->DimensionLengths(), true); instead as recommended.

source/core/GenericFunctions.cpp
```diff
@@ -1260,6 +1260,16 @@ struct AggregateStrToNumInstruction : public InstructionCore
 };
 typedef Instruction6<AQBlock1, Int32, AQBlock1, Int32, AQBlock1, AQBlock1> StrToNumInstructionArgs;
 
+void ResizeOutputToSizeOfInput(TypedArrayCoreRef arrayInput, TypedArrayCoreRef arrayOutput)
+{
+    ArrayDimensionVector newDimensionLengths;
+    IntIndex rank = 0;
+    std::vector<TypedArrayCoreRef> srcArrays;
+    srcArrays.push_back(arrayInput);
+    bool isInputArraysDimensionsSame = GetMinimumArrayDimensions(srcArrays, &newDimensionLengths, &rank);
```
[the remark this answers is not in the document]
cglzaguilar: Used VOutput->ResizeDimensions(VStr->Rank(), VStr->DimensionLengths(), true); instead, this function is not longe necessary.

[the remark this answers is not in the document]
cglzaguilar: Used VOutput->ResizeDimensions(VStr->Rank(), VStr->DimensionLengths(), true); instead, this function is not longe necessary.

.appveyor.yml
sanmut: I would recommend making the changes in .yml file as a separate PR and create a separate release.

cglzaguilar: Good call. A new PR with a Vireo release is being generated.

PR #657, opened 2020-01-03
Update to emscripten 1.39.5

rajsite:
In emscripten 1.39.0 the build infrastructure used by emscripten was changed which is seen from the name change from `sdk-<version>-64bit` to `<version>-fastcomp`.

In addition the `fastcomp` backend was changed from the default and while still supported has to be explicitly selected.

This change does not effect the underlying clang version 6.0.1 that is used but instead makes sure we are using emscripten versions released on the current infrastructure. The emscripten [changelog shows changes from 1.38.30 to 1.39.5](https://github.com/emscripten-core/emscripten/blob/incoming/ChangeLog.md).

Passing [ASW PR](https://ni.visualstudio.com/DevCentral/_build/results?buildId=272139&_a=summary&view=results).

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,8 +70,8 @@ The emsdk toolchain is used to cross-compile Vireo to Vireo.js to run in Node.js
 4. From either the extracted folder or the cloned emsdk repository run the following commands:
 
    ```console
-   emsdk install sdk-1.38.30-64bit
-   emsdk activate sdk-1.38.30-64bit
+   emsdk install 1.39.5-fastcomp
+   emsdk activate 1.39.5-fastcomp
    emcc -v # should match the sdk version
    ```
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -17,9 +17,9 @@ install:
   - pip install tox
   - ps: Start-Process npm "run httpbin" -PassThru
   - git clone https://github.com/juj/emsdk.git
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -70,8 +70,8 @@ The emsdk toolchain is used to cross-compile Vireo to Vireo.js to run in Node.js
 4. From either the extracted folder or the cloned emsdk repository run the following commands:
 
    ```console
-   emsdk install sdk-1.38.30-64bit
-   emsdk activate sdk-1.38.30-64bit
+   emsdk install 1.39.5-fastcomp
```
sanmut: As emscripten seems to be going towards  LLVM WebAssembly backend (ones with "-upstream" suffix),  should we try to switch to that instead of old "fastcomp"? I am not sure whether LLVM WebAssembly backend offers asm-js (it may not, I guess) and not sure whether we need to be able to build (as needed basis from now on) !? Thoughts?

sanmut:
Please run a CI build on the ASW side before merging.

README.md
rajsite: We should definitely consider it as a separate change. The goal of this change was to switch us to releases from the [new build infrastructure](https://groups.google.com/forum/#!msg/emscripten-discuss/WhDtqVyW_Ak/8DfDnfk0BgAJ) and be on the latest fastcomp backend so that if for some reason we do file issues to emscripten, etc it will be easier to merge them in.

The fastcomp backend is supported in maintenance mode, but we should migrate to the llvm backend when possible

PR #658, opened 2020-01-03
Update devDependencies for npm build tools

rajsite:
Updates the devDependencies used by npm for the build infrastructure. This gives us the latest packages and removes existing security warnings.


rajsite:
Looks like on appveyor the emscripten toolchain is replacing the environment node 8.16.2 with the bundled node 8.9.1 and eslint requires >=8.10.0 see https://github.com/eslint/eslint/issues/12493

So underlying issue is that the emscripten toolchain should be run in a way that does not clobber the existing node version. Closing until it's sorted out.

PR #660, opened 2020-01-13
Add registerCustomGlobal public api function

rajsite:
Adds a method to the public API that allows a user of the Vireo public API to change the object that the JavaScriptInvoke feature uses for looking up functions for the JSLI feature.

This is not a feature utilized by ASW but is instead valuable for users that access the vireo api directly. In particular this allows node.js users of WebVIs to have greater control of how JavaScript functions are resolved for the JSLI document and hacks can be avoided that [unnecessarily expose objects in the global scope](https://github.com/rajsite/webvi-experiments/blob/master/WebVINode/packages/%40webvi-node/runner/src/VireoNode.js#L5).

For ASW / normal WebVI users there should be no change in behavior of vireo. [ASW VireoTest build](https://ni.visualstudio.com/DevCentral/_build/results?buildId=277170&_a=summary&view=results).

### DIFF
--- a/source/io/module_javaScriptInvoke.js
+++ b/source/io/module_javaScriptInvoke.js
@@ -54,8 +54,6 @@ var assignJavaScriptInvoke;
         publicAPI.javaScriptInvoke = {};
 
         // Private Instance Variables (per vireo instance)
-        var internalFunctionsMap = new Map();
-
         // Every call to mergeNewError should be preceeded by the behavior for internalFunctions
         var mergeNewError = function (errorValueRef, functionName, errorToSet, exception) {
             var newError = {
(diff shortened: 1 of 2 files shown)


source/io/module_javaScriptInvoke.js
```diff
@@ -237,21 +235,44 @@ var assignJavaScriptInvoke;
...
+                return window;
+            } else if (typeof self !== 'undefined') {
+                return self;
+            } else if (typeof global !== 'undefined') {
+                return global;
+            }
+            return {};
+        };
+        var internalFunctionsMap = new Map();
+        var jsInvokeGlobal = getGlobal();
```
siddhukrs: Is there a reason why we don't use the term `scope` anymore? The new names that just end in `global` and not `globalScope` are slightly confusing, given the string in the JSLI document is also called a "global"

rajsite: Talked with sid and agreed that the name global is okay. The global object is a an object you can get a reference to and the properties of it are made accessible in all scopes. You cannot hold a reference to "the global scope" but you can hold a reference to the global object.

PR #661, opened 2020-01-21
Implement JavaScriptInvoke Array of JavaScript References

rajsite:
## Description

This PR adds support for Array of JavaScript References on the JavaScriptInvoke node. Existing code in ASW does not use this code path but it is being reserved for the future.

As part of this change the following bug fixes were made:

- Close Reference did not properly free JavaScript References when a reference was closed.

   The reference was made invalid however the JS maps containing the reference were not freed. This resulted in a memory leak for JS values that were expected to be freed.

- readJavaScriptRefNum incorrectly returned `undefined` for invalid JavaScript References instead of throwing an Error.

   This was a bug because `undefined` is also a valid value to store in a JavaScript Reference. By returning `undefined` for an invalid reference we made it impossible to distinguish between an invalid reference and a references holding the value `undefined`.

- Dynamic JavaScript references and Static JavaScript references used the same map and the same map reverse lookup code. This would cause errors in the following scenario:

   1. A static reference is made to a JS value in memory
   2. A dynamic reference is made to the same JS value in memory. By doing so it overwrites the [existing reverse lookup map entry](https://github.com/ni/VireoSDK/blob/afde64c5042eee7ed969feba8fdeb21211d10389/source/io/module_javaScriptInvoke.js#L83). 
   3. A static reference is created that retrieves the existing cookie value in the reverse lookup map that was overwritten by the dynamic reference. The static reference now has a dynamic reference cookie.
   4. The dynamic reference closes its reference which also destroys the static reference using the same cookie.

   This bug cannot currently arise in practice because static reference creation is not interleaved with dynamic reference creation as the editor emits all static reference creation first. However, it is placing the system in an invalid state.

## Implementation

- Separated the backing maps into StaticRefnumManager and DynamicRefnumManager. This lets them each place stronger asserts based on the reference type. It also encapsulates behaviors such as reverse lookups in each map code.

   This does not preclude sharing the maps in the future if needed, ie dynamic maps could fallback to shared maps for lookups if needed.

- Fixed eggshell readJavaScriptRefNum to throw for invalid refnums and added the complementing isJavaScriptRefNumValid api to allow checking before reading. Also added clearJavaScriptRefNum eggshell api for completeness allowing the eggShell api to create, read, write, and destroy a JS reference. Added tests for all of the eggShell JavaScript Reference apis since none existed.

- Added a case to parameterValueVisitor and returnValueVisitor for the JavaScriptInvoke node to support Arrays of JavaScriptReferences. Added DataValue and ReturnDataValue tests for both JavaScript References and Array of JavaScript References since those tests did not exist.

 - Added a test to verify that calls to Close Reference do get removed from the DynamicRefnumManager


source/io/module_eggShell.js
```diff
@@ -405,6 +397,8 @@ var assignEggShell;
                 subTypeRef = Module.typeHelpers.subElementByIndex(typeRef, 0);
                 if (Module.typeHelpers.isBoolean(subTypeRef)) {
                     return Uint8Array;
+                } else if (Module.typeHelpers.isJSObjectRefnum(subTypeRef)) {
```
rajsite: Added to allow bulk reading of JavaScript Refnum cookies from memory. Used by the paramter and returnValueVisitor

source/io/module_eggShell.js
```diff
@@ -10,14 +10,6 @@ var assignEggShell;
         Module.eggShell = {};
         publicAPI.eggShell = {};
 
-        Module.eggShell.readJavaScriptRefNum = publicAPI.eggShell.readJavaScriptRefNum = function (valueRef) {
```
rajsite: Updated these functions to use the same error handling pattern as our other public eggShell functions below

source/io/module_javaScriptInvoke.js
```diff
@@ -74,32 +74,121 @@ var assignJavaScriptInvoke;
             return returnValueRef;
         };
 
-        var cookieToJsValueMap = new Map();
-        var jsValueToCookieCache = new Map();
-        var jsRefNumCookieCounter = 0;
+        var generateUniqueRefNumCookie = (function () {
```
rajsite: Cookie generation is shared by both maps so all cookie values are unique

source/io/module_javaScriptInvoke.js
```diff
@@ -74,32 +74,121 @@ var assignJavaScriptInvoke;
...
+
+            get size () {
+                return this._cookieToJsValueMap.size;
+            }
+        }
+
+        var staticRefnumManager = new StaticRefnumManager();
+        var dynamicRefnumManager = new DynamicRefnumManager();
+
+        Module.javaScriptInvoke.do_not_use_debug_only_static_refnum_manager = staticRefnumManager;
```
rajsite: Placed on the internal module to make available in tests but should not be used by production code

test-it/karma/javascriptinvoke/CloseReference.Test.js
```diff
@@ -0,0 +1,84 @@
...
+        var viName = 'MyVI';
+        var runSlicesAsync = vireoRunner.rebootAndLoadVia(vireo, closeReferenceViaUrl);
+        var viPathParser = vireoRunner.createVIPathParser(vireo, viName);
+
+        expect(staticReferencesMapSize()).toBe(0);
+        expect(dynamicReferencesMapSize()).toBe(0);
+
+        const {rawPrint, rawPrintError} = await runSlicesAsync();
+
+        expect(staticReferencesMapSize()).toBe(0);
```
rajsite: Memory leak test asserts that map empty after vireo runs. The test itself verifies map clears with Close Reference, see the via code

test-it/karma/javascriptinvoke/DataTypes.Test.js
```diff
@@ -232,62 +284,4 @@ describe('A JavaScript function invoke', function () {
             done();
         });
     });
-
-    it('successfully create and use an object type', function (done) {
```
rajsite: This test exercising a lot more than JavaScript Reference parameter passing behavior like the other functions in this file. Moved to separate DiagramUsage test.

source/io/module_javaScriptInvoke.js
```diff
@@ -74,32 +74,121 @@ var assignJavaScriptInvoke;
...
+                for (i = 0; i < values.length; i += 1) {
+                    cookie = generateUniqueRefNumCookie();
+                    this._cookieToJsValueMap.set(cookie, values[i]);
+                    cookies[i] = cookie;
+                }
+            }
+
+            get (cookie) {
+                // Have to check the map because undefined and null are possible values for dynamic references
+                if (!this._cookieToJsValueMap.has(cookie)) {
```
siddhukrs: Good catch.

source/io/module_javaScriptInvoke.js
```diff
@@ -111,17 +200,35 @@ var assignJavaScriptInvoke;
...
+            }
+            return dynamicRefnumManager.has(cookie);
+        };
+
+        // static reference (static control reference) always returns true (the operation was successful, which for static references is a no-op)
+        // dynamic reference (JS opaque reference from JSLI) returns true if the reference was found + cleared and false if the reference was invalid
+        Module.javaScriptInvoke.clearJavaScriptRefNum = function (javaScriptValueRef) {
+            var isStaticReference = Module.typeHelpers.isJSObjectStaticRefnum(javaScriptValueRef.typeRef);
+            if (isStaticReference) {
+                return true;
```
siddhukrs: Why doesn't this case have to clear the `_cookieToJsValueMap`?

source/io/module_javaScriptInvoke.js
```diff
@@ -74,32 +74,121 @@ var assignJavaScriptInvoke;
...
+            constructor () {
+                this._cookieToJsValueMap = new Map();
+                this._jsValueToCookieCache = new Map();
+            }
 
-        var hasCachedRefNum = function (cookie) {
-            var refNumExists = cookieToJsValueMap.has(cookie);
-            return refNumExists;
-        };
+            set (jsValue) {
```
cglzaguilar: I find this name a little bit odd. A "set" that returns a value. How about getCookie instead of set and getValue for get?

source/io/module_javaScriptInvoke.js
rajsite: Static references are not closed by Close Reference; the diagram cannot close static references. It's a no-op. Maybe in the future we can add a boolean to force close or something to the public api but there hasn't been a need yet.

rajsite: I'll add a comment for that

source/io/module_javaScriptInvoke.js
rajsite: I went with createCookie and lookupValue thanks for the suggestions!

PR #662, opened 2020-01-23
Bug fix for Array concatenate memory corruption

cglzaguilar:
The code is not accounting for arrays that have not been initialized and it not properly resizing the output array. The code assumes there is space after the resize and when writing data we overwrite memory and get memory access out of bound error.

This a bug fix for the associated AzDO bug 922467: Vireo memory access out of bounds with build array and replace array subset.

The problem was in array concatenate code, not in the array subset code.


test-it/ViaTests/ArrayConcatenateUninitialized.via
```diff
@@ -0,0 +1,24 @@
+// Autogenerated Vireo assembly file
```
cglzaguilar: ArrayConcatenate

test-it/ExpectedResults/ArrayConcatenateUninitialized.vtr
```diff
@@ -0,0 +1 @@
+Result: (((1 2 3 4) (0 0 0 8) (0 0 0 12) (0 0 0 16) (0 0 0 20) (0 0 0 24) (0 0 0 28) (0 0 0 32) (0 0 0 36)) ((1 2 3 4) (0 0 0 8) (0 0 0 12) (0 0 0 16) (0 0 0 20) (0 0 0 24) (0 0 0 28) (0 0 0 32) (0 0 0 36)))
```
sanmut: @cglzaguilar Is this the same output when this VI is run Desktop target?

cglzaguilar: Yes, it is the same output.

cglzaguilar:
After discussing with Sankara, it appears the fix is not correct. Closing PR to continue working on it.

PR #664, opened 2020-02-03
Bug922467

spathiwa:
Vireo memory access out of bounds with build array 
(Also dup bug 902123, Build array functionality is different in WebVI and VI)


sanmut:
> I created a local build of Vireo with the changes and used it to run all the tests in AzDo. All tests passed, but two. They appear to be unrelated to the change. I ran them locally and they pass.

Thanks a lot Carlos !!!

PR #665, opened 2020-02-11
Enable usage of python 3

rajsite:
Some change to the Travis build agents was causing installation of the Python-based httpbin server (used for testing the http client) to fail and causes all of our [Travis CI builds to fail](https://travis-ci.org/rajsite/VireoSDK/builds/648570884?utm_medium=notification&utm_source=github_status) with message:

```
virtualenv 20.0.1 has requirement six<2,>=1.12.0, but you'll have six 1.10.0 which is incompatible.
```

This failure revealed that that our usage of [tox to create a virtualenv is not recommended on TravisCI](https://docs.travis-ci.com/user/languages/python/#using-tox-as-the-build-script). The travis.yml was modified to avoid tox by running commands directly (where Travis has already configured a virtualenv).

As part of this change I enabled the use of python 3 as [python 2 is deprecated as of Jan 1, 2020](https://www.python.org/dev/peps/pep-0373/#id4).

To enable python 3 usage I had to upgrade our CPPLINT version where I switched to the version maintained on https://github.com/cpplint/cpplint.

These changes currently enable developers to use **either** python 2 or python 3 (in particular the appveyor windows build still uses python 2).

### DIFF
--- a/source/core/TDCodecVia.cpp
+++ b/source/core/TDCodecVia.cpp
@@ -17,6 +17,8 @@ SDG
 #include <ctype.h>
 #include <memory>
 #include <algorithm>
+#include <vector>
+
 #include "TypeDefiner.h"
 #include "ExecutionContext.h"
 #include "TypeAndDataManager.h"
@@ -24,7 +26,6 @@ SDG
 #include "TDCodecVia.h"
 #include "ControlRef.h"
 #include "Events.h"
-#include <vector>
 
 #include "VirtualInstrument.h"  // TODO(PaulAustin): remove once it is all driven by the type system.
 #include "Variants.h"
--- a/test-it/karma/http/HttpGet.Test.js
+++ b/test-it/karma/http/HttpGet.Test.js
@@ -370,10 +370,7 @@ describe('Performing a GET request', function () {
         });
     });
 
-    it('validating a response with binary data seeded bytes', function (done) {
-        // Body bytes: '\xCB\xD2\x7C\x42\x00\xA9\x78\xC2';
-        var bodyReadAsString = '\uFFFD\uFFFD\x7C\x42\x00\uFFFD\x78\uFFFD';
-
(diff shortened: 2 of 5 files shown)


source/core/TDCodecVia.cpp
```diff
@@ -17,14 +17,15 @@ SDG
 #include <ctype.h>
 #include <memory>
 #include <algorithm>
+#include <vector>
```
rajsite: Caught by new cpplint, move system headers to be included first

test-it/karma/http/HttpGet.Test.js
```diff
@@ -396,8 +393,13 @@ describe('Performing a GET request', function () {
...
             // body
-            var responseBody = textFormat.normalizeLineEndings(viPathParser('body'));
-            expect(responseBody).toBe(bodyReadAsString);
+            // Body bytes: '\xCB\xD2\x7C\x42\x00\xA9\x78\xC2';
+            // var bodyReadAsString = '\uFFFD\uFFFD\x7C\x42\x00\uFFFD\x78\uFFFD';
+            // var responseBody = textFormat.normalizeLineEndings(viPathParser('body'));
+
+            // Httpbin seed not stable between Python versions. See https://github.com/postmanlabs/httpbin/issues/598
+            // Need an alternate endpoint to test null values in responses. See https://github.com/postmanlabs/httpbin/issues/599
+            // expect(responseBody).toBe(bodyReadAsString);
```
rajsite: Disabled assertion because result is dependent on python version. Filed bug with httpbin related to this behavior.

sanmut: Interesting. Glad you filed an issue with httpbin. Should we also create an issue in Vireo and reference the httpbin issue in that? Or, leave a TODO in the comments referencing the httpbin issue?

rajsite: Created https://github.com/ni/VireoSDK/issues/666 to track the disabled test

PR #671, opened 2020-06-24
Update README.md

rajsite:


rajsite:
Was demonstrating the fork-pull workflow for an EUIB tech talk and made an example PR. While I definitely agree, the current README title is probably sufficient. I'll close the PR :P

PR #672, opened 2020-06-26
Fix ByteArrayNode WebVI error code to match Desktop

tatwood2:
A developer noticed that the error code for the Web ByteArray node was 1 while desktop is 1396 (AzDo bug 1068756).
This change switches the error code and updates the tests.

### DIFF
--- a/source/core/NumericString.cpp
+++ b/source/core/NumericString.cpp
@@ -22,6 +22,9 @@
 
 namespace Vireo {
 
+// Encoding error codes
+enum { kInvalidUTF8CharErr = 1, kByteArrayTypeErr = 1, kCannotConvertErr = 1396 };
+
 //------------------------------------------------------------
 struct FormatOptions {
     Int32 ArgumentOrder;       // 3$, 2$ negative number means default order
@@ -2090,7 +2093,10 @@ struct ByteArrayToStringParamBlock : InstructionCore
 static bool CheckUnsupportedEncodingError(ErrorCluster* errorCluster, UInt16 stringEncoding, TypedArrayCoreRef outputArray) {
     if (stringEncoding != 0) {
         if (errorCluster != nullptr) {
-            errorCluster->SetErrorAndAppendCallChain(true, 1396, "Encoding type not supported");
+            errorCluster->SetErrorAndAppendCallChain(true, kCannotConvertErr, "String To Byte Array");
+            errorCluster->AddAppendixPreamble();
+            errorCluster->source->AppendCStr("Encoding type not supported");
+            errorCluster->AddAppendixPostamble();
         }
         if (outputArray != nullptr)
             outputArray->Resize1D(0);
@@ -2099,11 +2105,14 @@ static bool CheckUnsupportedEncodingError(ErrorCluster* errorCluster, UInt16 str
     return true;
 }
 
-static bool CheckValidUTF8Error(UInt8* begin, UInt8* end, ErrorCluster* errorCluster)
+static bool CheckValidUTF8Error(UInt8* begin, UInt8* end, ErrorCluster* errorCluster, ConstCStr source)
 {
     if (!utf8::is_valid(begin, end)) {
         if (errorCluster != nullptr) {
-            errorCluster->SetErrorAndAppendCallChain(true, 1, "Invalid UTF-8 characters found");
+            errorCluster->SetErrorAndAppendCallChain(true, kInvalidUTF8CharErr, source);
+            errorCluster->AddAppendixPreamble();
+            errorCluster->source->AppendCStr("Invalid UTF-8 characters found");
+            errorCluster->AddAppendixPostamble();
         }
         return false;
     }
@@ -2118,9 +2127,13 @@ VIREO_FUNCTION_SIGNATURET(ByteArrayToString, ByteArrayToStringParamBlock)
         return _NextInstruction();
     TypedArrayCoreRef byteArray = _Param(ByteArrayIn);
     TypeRef eltType = byteArray->ElementType();
+    ConstCStr functionName = "Byte Array To String";
     if (!eltType->IsA(&TypeCommon::TypeInt8) && !eltType->IsA(&TypeCommon::TypeUInt8)) {
         if (errorCluster != nullptr) {
-            errorCluster->SetErrorAndAppendCallChain(true, 1, "Byte Array must be of type Int8 or UInt8");
+            errorCluster->SetErrorAndAppendCallChain(true, kByteArrayTypeErr, functionName);
+            errorCluster->AddAppendixPreamble();
+            errorCluster->source->AppendCStr("Byte Array must be of type Int8 or UInt8");
+            errorCluster->AddAppendixPostamble();
         }
         return _NextInstruction();
     }
@@ -2131,7 +2144,7 @@ VIREO_FUNCTION_SIGNATURET(ByteArrayToString, ByteArrayToStringParamBlock)
     if (!CheckUnsupportedEncodingError(errorCluster, stringEncoding, stringOut)) {
         return _NextInstruction();
     }
-    if (!CheckValidUTF8Error(byteArray->RawBegin(), byteArray->RawBegin() + arrayLength, errorCluster)) {
+    if (!CheckValidUTF8Error(byteArray->RawBegin(), byteArray->RawBegin() + arrayLength, errorCluster, functionName)) {
         if (stringOut != nullptr)
             stringOut->Resize1D(0);
         return _NextInstruction();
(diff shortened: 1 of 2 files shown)


tatwood2:
This is to fix https://ni.visualstudio.com/DevCentral/_workitems/edit/1068756.  The bug only mentions changing the error code, not modifying the message string to match desktop.  When this vireo change makes it to master, I'll need to update the test mentioned there at the same time.

rajsite:
It might be overkill for this specific PR but to get familiar with the workflow can you please include a link to the ASW draft PR with the ASW tests and [run a npm pack build](https://github.com/ni/VireoSDK/blob/master/CONTRIBUTING.md#testing-vireojs-on-a-different-machine--ci) against the [Web Module Squad tests](https://ni.visualstudio.com/DevCentral/_wiki/wikis/AppCentral.wiki/2445/How-to-update-Vireo-in-NXG?anchor=trigger-the-web-module-squad-tests).

Once the PR is approved on GitHub the changes will be merged and a new release made to integrate to the ASW PR.

sanmut:
> This is to fix https://ni.visualstudio.com/DevCentral/_workitems/edit/1068756. The bug only mentions changing the error code, not modifying the message string to match desktop. When this vireo change makes it to master, I'll need to update the test mentioned there at the same time.

It is a good idea to include the name of the node in the error message to match Desktop behavior while you are at it. Example: [SetErrorAndAppendCallChain call](https://github.com/ni/VireoSDK/blob/e9a4558ff53873acabf1fa935e6cff0d7b2407aa/source/core/Variants.cpp#L97) using the name of the node "Variant To Data" correctly.

sanmut:
Great, thanks for fixing this :-)

source/core/NumericString.cpp
```diff
@@ -2090,7 +2090,7 @@ struct ByteArrayToStringParamBlock : InstructionCore
 static bool CheckUnsupportedEncodingError(ErrorCluster* errorCluster, UInt16 stringEncoding, TypedArrayCoreRef outputArray) {
     if (stringEncoding != 0) {
         if (errorCluster != nullptr) {
-            errorCluster->SetErrorAndAppendCallChain(true, 1, "Encoding type not supported");
+            errorCluster->SetErrorAndAppendCallChain(true, 1396, "Encoding type not supported");
```
cglzaguilar, line 2093: Please create an enum for the error code above the function definition. Something like this:

// Encoding error codes
enum { kCannotConvertErr = 1396 };

And use kCannotConvertErr instead of the "1396" hardcode constant.

We have been updating these hardcode numbers to something more meaningful as we change the code.

rajsite, line 2093: @sanmut described updating the SetErrorAndAppendCallChain() call to use the VI name, ie

```cpp
errorCluster->SetErrorAndAppendCallChain(true, 1396, "Byte Array To String or whatever");
```

But it looks like all the calls to SetErrorAndAppendCallChain in this file are incorrectly giving a "supplementary" error instead of the VI Name.

I believe the callsites in this file should be updated similar to how the [Format function reports errors](https://github.com/ni/VireoSDK/blob/master/source/core/NumericString.cpp#L276).

ie

```cpp
errorCluster->SetErrorAndAppendCallChain(true, 1396, "Byte Array To String or whatever");
errPtr->AddAppendixPreamble();
errPtr->source->AppendCStr("Encoding type not supported");
errPtr->AddAppendixPostamble();
```

tatwood2: @rajsite. the fake PR to pull in my changes resulted in a single failing test: https://ni.visualstudio.com/DevCentral/_git/ASW/pullrequest/80900
I'll fix it while making the rest of these changes (updating the error string will result in needing to update a few C# tests).

sanmut: @rajsite, Good point, Agreed. I think we need to cleanup the error message handling throughout that has not been prioritized yet. Now, there are different levels of badness in different places in the code.

tatwood2: @rajsite, @sanmut, and @cglzaguilar, I've made some changes per your requests.  I had to add a parameter to one function so it could correctly print out the proper error message (String To Byte vs Byte To String).  The AzDo build just started but should have results in the next hour and a half here: https://ni.visualstudio.com/DevCentral/_git/ASW/pullrequest/80900.

source/core/NumericString.cpp
```diff
@@ -2118,9 +2127,13 @@ VIREO_FUNCTION_SIGNATURET(ByteArrayToString, ByteArrayToStringParamBlock)
         return _NextInstruction();
     TypedArrayCoreRef byteArray = _Param(ByteArrayIn);
     TypeRef eltType = byteArray->ElementType();
+    ConstCStr functionName = "Byte Array To String";
     if (!eltType->IsA(&TypeCommon::TypeInt8) && !eltType->IsA(&TypeCommon::TypeUInt8)) {
```
rajsite, line 2131: This check should actually be a [hard error](https://github.com/ni/VireoSDK/blob/master/source/io/HttpClient.cpp#L160) which stops the runtime since we would never expect the editor to emit anything besides an Int8 or a UInt8.

I don't feel strongly about changing this from a runtime error since it is unlikely the editor would change types without this being caught. Do you agree or should this be changed to a hard error @sanmut?

source/core/NumericString.cpp
```diff
@@ -2165,18 +2178,22 @@ VIREO_FUNCTION_SIGNATURET(StringToByteArray, StringToByteArrayParamBlock)
...
         // If the user did not pass errorCluster, no need to validate the string, as we are copying
         // to byteArrayOut unconditionally
-        CheckValidUTF8Error(stringIn->RawBegin(), stringIn->RawBegin() + stringLength, errorCluster);
+        CheckValidUTF8Error(stringIn->RawBegin(), stringIn->RawBegin() + stringLength, errorCluster, functionName);
     }
     if (byteArrayOut != nullptr) {
         IntIndex rank = byteArrayOut->Rank();
         VIREO_ASSERT(rank == 1);
         TypeRef eltType = byteArrayOut->ElementType();
         if (!eltType->IsA(&TypeCommon::TypeUInt8)) {
```
rajsite, line 2191: Based on the other comment this may also update to a harderror instead of a runtime error

source/core/NumericString.cpp
```diff
@@ -22,6 +22,9 @@
 
 namespace Vireo {
 
+// Encoding error codes
+enum { kInvalidUTF8CharErr = 1, kByteArrayTypeErr = 1, kCannotConvertErr = 1396 };
```
rajsite, line 26: Was double checking the error code for kInvalidUTF8CharErr and looks like desktop does not error for invalid utf8 sequences in the source byte array >.<

Seems like incorrect behavior on desktop's part. I'm not sure what ends up in the string output...

source/core/NumericString.cpp
```diff
@@ -2090,7 +2093,10 @@ struct ByteArrayToStringParamBlock : InstructionCore
 static bool CheckUnsupportedEncodingError(ErrorCluster* errorCluster, UInt16 stringEncoding, TypedArrayCoreRef outputArray) {
     if (stringEncoding != 0) {
         if (errorCluster != nullptr) {
-            errorCluster->SetErrorAndAppendCallChain(true, 1, "Encoding type not supported");
+            errorCluster->SetErrorAndAppendCallChain(true, kCannotConvertErr, "String To Byte Array");
```
cglzaguilar, line 2096: CheckUnsupportedEncodingError appears to be called in both ByteArrayToString and StringToByteArray, but here we are just reporting the error as always happening with "String To Byte Array". Consider adding a parameter and pass the name to avoid this problem.

source/core/NumericString.cpp
sanmut: @rajsite It is ok to change this to a hard error. NXG before VIA generation must guarantee that input to this node is Int8 or UInt8. We should add tests in NXG to test arrays of different types such as Int16, UInt16, if we already don't have test coverage.

source/core/NumericString.cpp
sanmut: Yes, NXG must guarantee this is TypeUInt8

source/core/NumericString.cpp
tatwood2: I've changed this to a hard error in the third commit.  Any particular place the NXG array test should live?

source/core/NumericString.cpp
tatwood2: Done in commit 3.

source/core/NumericString.cpp
tatwood2: Done in update 3.

source/core/NumericString.cpp
rajsite: Minor nit, looks like kByteArrayTypeErr is unused now

PR #681, opened 2021-06-09
Update the emscripten version to 2.0.23

Devika-Tantry:
ASW work item : [Update emscripten tooling used to build Vireo](https://ni.visualstudio.com/DevCentral/_workitems/edit/1334978/)
Ran [ASW PR ](https://ni.visualstudio.com/DevCentral/_git/ASW/pullrequest/168734?path=%2FPipeline%2FSquads%2FASW-Web-Module%2FRun-SquadTests.txt)with this updated package for test.
# Changes:
- Updated emscripten version to 2.0.23 , python to 3.9
- Error fix in casting case. (It was due to an upgrade in clang it promotes all warnings to error now.)
- Flag fix as per release note :
           - `BINARYEN_TRAP_MODE=js` is depreciated.
           - ` -s NO_EXIT_RUNTIME=1` is by default set to 1 so removed.
           - `EXTRA_EXPORTED_RUNTIME_METHODS` renamed to `EXPORTED_RUNTIME_METHODS`
           -  Updated the flag in debug build to `-g` 
image
so now debug output will have DWARF info.
 Note: With this updated vireo package we can run the web gvi's in [Chrome Canary](https://www.google.com/chrome/canary/).(with setup mentioned [here](https://developer.chrome.com/blog/wasm-debugging-2020/)).
`-g` applies on debug build so we have to use debug wasm to directly interact with CPP files.

In WebVI, we use Viero SDK wasm build. Screenshots of putting breakpoints in Vireo SDK CPP files and watching values in the browser in WebVI.
image


rajsite (changes requested):
@Devika-Tantry some feedback:

1. The `Error: No tool or SDK found by name '2.0.20'.` is likely due to the version of the emsdk tools used. We checkout a specific commit ([web reference](https://github.com/ni/VireoSDK/blob/c8eb11d197b09f7ed2100bdb01cd48cd8f2fa003/.github/workflows/build.yml#L58), [web-windows reference](https://github.com/ni/VireoSDK/blob/c8eb11d197b09f7ed2100bdb01cd48cd8f2fa003/.github/workflows/build.yml#L121)) because the tools can be unreliable sometimes. Try changing to one of the newer commits from emscripten-core/emsdk. I would recommend the latest [emsdk 2.0.23 release commit](https://github.com/emscripten-core/emsdk/commit/f44b84154703d29a946d51ddee08f4e5725a61db).
2. In the PR comment can you link to the associated ASW work item
3. Can you run an ASW PR build [with your local vireo change](https://github.com/ni/VireoSDK/blob/main/docs/Workflow.md#testing-vireojs-on-a-different-machine--ci) and make sure to enable the [[VireoTest] suite](https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/2445/How-to-update-Vireo-in-NXG?anchor=trigger-the-web-module-squad-tests) for the build (see the section "Trigger the Web Module Squad tests").
4. Include a link to the ASW build result run with [VireoTest] in the PR
5. (optional but it's cool) Sujay shared a screenshot of being able to debug Vireo inside Chrome Canary which is awesome! Can you add a screenshot showing debugging in the browser to the PR description? Shows the benefit of updating the toolchain :)
6. ~For the flag changes could you add comments in the PR for those lines that points to the change description? For example it looks like this is a [reference to the NO_EXIT_RUNTIME change](https://github.com/emscripten-core/emscripten/blame/a8a0d8e79ed5d4aa640e18ae7497ef4c48e5beaa/ChangeLog.md#L1464).~ Went ahead and added some comments as I was interested in the changes.

spathiwa (approved):

make-it/EmMakefile
```diff
@@ -92,11 +92,11 @@ ifeq ($(TARGET), asmjs-unknown-emscripten)
 endif
 
 ifeq ($(TARGET), wasm32-unknown-emscripten)
-   TARGET_FLAG = -s WASM=1 -s BINARYEN_TRAP_MODE=js -s ALLOW_MEMORY_GROWTH=1
+   TARGET_FLAG = -s WASM=1 -s ALLOW_MEMORY_GROWTH=1
 endif
 
 # NO_DYNAMIC_EXECUTION=1 Allows Vireo to be used in CSP environments where eval() is forbidden: https://github.com/kripken/emscripten/blob/master/src/settings.js#L563
-EM_OPT = $(EM_OPTFLAGS) $(TARGET_FLAG) -s NO_EXIT_RUNTIME=1 -std=c++14 -fno-exceptions -Werror --memory-init-file 0 -s NO_DYNAMIC_EXECUTION=1 -s MODULARIZE=1 -s EXPORT_NAME=VireoCreateCoreModule -s EXPORT_ES6=1 -s USE_ES6_IMPORT_META=0 --minify 0 -s NO_FILESYSTEM=1 -s ABORTING_MALLOC=0
```
rajsite, line 99: Looks like [NO_EXIT_RUNTIME=1 became the default](https://github.com/emscripten-core/emscripten/blame/a8a0d8e79ed5d4aa640e18ae7497ef4c48e5beaa/ChangeLog.md#L1464) was it removed because it's not needed or was it causing some other error?

make-it/EmMakefile
```diff
@@ -92,11 +92,11 @@ ifeq ($(TARGET), asmjs-unknown-emscripten)
 endif
 
 ifeq ($(TARGET), wasm32-unknown-emscripten)
-   TARGET_FLAG = -s WASM=1 -s BINARYEN_TRAP_MODE=js -s ALLOW_MEMORY_GROWTH=1
```
rajsite, line 95: Looks like [BINARYEN_TRAP_MODE has been removed](https://github.com/emscripten-core/emscripten/issues/10365#issuecomment-856810416). As long as the [Double.Test.js tests](https://github.com/ni/VireoSDK/blob/main/test-it/karma/publicapi/Double.Test.js) pass I think we are good, they were the [reason we initially had the flag](https://github.com/ni/VireoSDK/commit/790d42a5de70b1b08f99c4a3b7c7d620ddf659d8).

make-it/EmMakefile
```diff
@@ -72,7 +72,7 @@ EMCC= emcc
 EM_OPTFLAG = -O3
 
 ifeq ($(BUILD), debug)
-   EM_OPTFLAG = -g4 -s ASSERTIONS=2 -s SAFE_HEAP=1
+   EM_OPTFLAG = -g -s ASSERTIONS=2 -s SAFE_HEAP=1
```
rajsite, line 75: Looks like [-g4 was removed](https://github.com/emscripten-core/emscripten/blame/a8a0d8e79ed5d4aa640e18ae7497ef4c48e5beaa/ChangeLog.md#L149) as it is the same as -g3 (and we don't need the source maps `-gsource-map` feature if we have the dwarf symbols). In addition [-g is the same as -g3](https://emscripten.org/docs/tools_reference/emcc.html#emcc-g).

make-it/EmMakefile
```diff
@@ -111,7 +111,7 @@ EM_EXPORTS = -s EXPORTED_FUNCTIONS="[\
     '_free' \
     ]"
 
-EM_RUNTIME_EXPORTS = -s EXTRA_EXPORTED_RUNTIME_METHODS="[\
```
rajsite, line 114: [EXTRA_EXPORTED_RUNTIME_METHODS was deprecated](https://github.com/emscripten-core/emscripten/blame/a8a0d8e79ed5d4aa640e18ae7497ef4c48e5beaa/ChangeLog.md#L129) in favor of `EXPORTED_RUNTIME_METHODS`.

make-it/EmMakefile-AAL
```diff
@@ -35,7 +35,7 @@ ifeq (,$(wildcard $(ANALYSIS_WRAPPERS_SOURCE_DIRECTORY)/AnalysisWrapper*.cpp))
 ANALYSIS_LIBRARY = $(OBJS)/NoAnalysisLibrary.bc
 $(OBJS)/NoAnalysisLibrary.bc: | no-analysis
 	touch $(OBJS)/NoAnalysis.cpp
-	$(EMCC) $(EM_ANALYSIS_FLAGS) -o $@ $(OBJS)/NoAnalysis.cpp
+	$(EMCC) $(EM_ANALYSIS_FLAGS) -o $@ $(OBJS)/NoAnalysis.cpp -c
```
rajsite, line 38: The `-c` flag runs a [subset of steps](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-c) (only run preprocess, compile, and assemble steps). Why was this change made?

rajsite:
Looks like the PR is still in-work but a couple more notes:
1. Can you add an entry to the `.gitignore` for the `.vs` directory so those files are not checked in?
2. Can we scope the changes as close as possible to just the compiler update? ie lets not update the python and node versions unless it is required

cglzaguilar (approved):
I have not additional questions. Please address the feedback given by @rajsite.

source/core/TypeAndDataManager.cpp
```diff
@@ -2637,7 +2637,8 @@ IntMax ReadIntFromMemory(TypeRef type, void* pData)
                 case 0: value = 0;                              break;
                 case 4: {
                     Single singleVal = RoundToEven(*(Single*)pData);
-                    if (singleVal >= std::numeric_limits<IntMax>::max())
+                    Single intMax = (Single)(std::numeric_limits<IntMax>::max());
```
sanmut, line 2640: const ?

source/core/TypeAndDataManager.cpp
```diff
@@ -2648,7 +2649,8 @@ IntMax ReadIntFromMemory(TypeRef type, void* pData)
                 }
                 case 8: {
                     Double doubleVal = RoundToEven(*(Double*)pData);
-                    if (doubleVal >= std::numeric_limits<IntMax>::max())
+                    Double intMax = (Double)(std::numeric_limits<IntMax>::max());
```
sanmut, line 2652: const?

source/include/TypeAndDataManager.h
```diff
@@ -366,13 +366,14 @@ template <typename TSource, typename TDest>
 TDest ConvertFloatToInt(TSource src)
 {
     TDest dest;
+    TSource maxDestValue = static_cast<TSource>(std::numeric_limits<TDest>::max());
```
sanmut, line 369: const?

sanmut (changes requested):
Thanks. Would like to take another look with feedback addressed.

make-it/EmMakefile-AAL
Devika-Tantry: For the intermediate level, we need only object file, previously by default object files were generating. Now we have to explicitly mention the -c flag.

2.0.3 realese note snipet:
image

make-it/EmMakefile
Devika-Tantry: Yes. Now it is the default setting so removed it.

make-it/EmMakefile
Devika-Tantry: Now because of the default clang trap mode, we are facing test failure in Double.Test.js as you mentioned. 
So we added an extra function in [WriteDoubleToMemory ](https://github.com/ni/VireoSDK/blob/c8eb11d197b09f7ed2100bdb01cd48cd8f2fa003/source/core/TypeAndDataManager.cpp#L2847). This function will handle overflow and underflow from double to Int as JS trap was doing. 

@rajsite  Is this an intended change or do you have any other comment?

source/core/Array.cpp
```diff
@@ -189,11 +189,11 @@ VIREO_FUNCTION_SIGNATUREV(ArrayFillNDV, ArrayFillNDVParamBlock)
             tempDimensionLengths[i] = 0;
     }
 
-    _Param(ArrayOut)->ResizeDimensions(numDimensionInputs, tempDimensionLengths, false);
-
-    IntIndex totalLength = _Param(ArrayOut)->Length();
-    TypeRef eltType = array->ElementType();
-    eltType->MultiCopyData(_ParamPointer(InitialValue), array->RawBegin(), totalLength);
+    if (_Param(ArrayOut)->ResizeDimensions(numDimensionInputs, tempDimensionLengths, false)) {
+        IntIndex totalLength = _Param(ArrayOut)->Length();
```
Devika-Tantry, line 193: During resizing due to some failure if there is no action done we have to skip succeeding dependent code. There was a test failure due to this.

source/core/TypeAndDataManager.cpp
```diff
@@ -2453,8 +2453,7 @@ Boolean TypedArrayCore::ResizeDimensions(Int32 rank, IntIndex *dimensionLengths,
     }
 
     // 3. If more actual elements are needed, initialize the new ones (or all of them if requested)
-    // TODO(PaulAustin): honor bOK status.
-    if (!preserveElements) {
+    if (!preserveElements && bOK) {
         ElementType()->InitData(BeginAt(0), newLength);
```
Devika-Tantry, line 2457: If we set the bOK to false then we already have some issue so added a check. ([This ](https://github.com/ni/VireoSDK/blob/c8eb11d197b09f7ed2100bdb01cd48cd8f2fa003/test-it/karma/publicapi/ExecuteSlicesUntilClumpsFinshed.Test.js#L118)test failing due to this )

source/include/TypeAndDataManager.h
```diff
@@ -366,20 +366,39 @@ template <typename TSource, typename TDest>
...
         dest = std::numeric_limits<TDest>::max();
     } else {
         dest = static_cast<TDest>(RoundToEven(src));
     }
     return dest;
 }
 
+template <typename TDest>
+TDest ConvertDoubleToInt(Double src)
+{
```
Devika-Tantry, line 386: Function to mimic the `BINARYEN_TRAP_MODE=js`

Devika-Tantry:
> Looks like the PR is still in-work but a couple more notes:
> 
> 1. Can you add an entry to the `.gitignore` for the `.vs` directory so those files are not checked in?
> 2. Can we scope the changes as close as possible to just the compiler update? ie lets not update the python and node versions unless it is required
 
- Added  vs files in `.gitignore`
- Without  node update the build will fail it a dependency for updated emscrption

source/include/TypeAndDataManager.h
```diff
@@ -366,20 +366,39 @@ template <typename TSource, typename TDest>
...
+    TDest dest;
```
sanmut, line 387: Curious. How did you come up with this logic? Was it documented in emscripten? I don't quite understand use of fmod. Could you explain?

sanmut (changes requested):
Waiting on review feedback.

source/core/TypeAndDataManager.cpp
Devika-Tantry: The existing vireo  [test](https://github.com/ni/VireoSDK/blob/c8eb11d197b09f7ed2100bdb01cd48cd8f2fa003/test-it/karma/publicapi/ExecuteSlicesUntilClumpsFinshed.Test.js#L118) was failing after emsdk update.
I think the intention of this test to check the overflow error report generated [here](https://github.com/ni/VireoSDK/blob/c8eb11d197b09f7ed2100bdb01cd48cd8f2fa003/source/core/TypeAndDataManager.cpp#L162). But bOK (which suppose to set to false on the error)  flag does not tested down the code and trying to allocate/access the wrong location which leads to a run time segmentation fault. So added a bOK check before accessing the location. 
For this particular test, we already know about the error before this line, and bOK is set to false so we don't want to do `InitData` which leads to a segmentation fault.

source/core/Array.cpp
Devika-Tantry: The vireo test mentioned the above comment failing after emsdk update.
I think the intention of this test to check the overflow error report generated [here](https://github.com/ni/VireoSDK/blob/c8eb11d197b09f7ed2100bdb01cd48cd8f2fa003/source/core/TypeAndDataManager.cpp#L162).
so for the particular case, `ResizeDimensions` won't happen because of the expected memory allocation error. We are accessing that array later in this function which leads to segmentation faults so added check before using it.

source/include/TypeAndDataManager.h
Devika-Tantry: As mentioned emscrripten site BINARYEN_TRAP_MODE=js  was the same as how JS handling the overflow and underflow. From the [emca ](https://262.ecma-international.org/6.0/#sec-touint8)site, we can see how ToInt32 ToInt8 etc works. I tried to get similar behavior in C++ with this function.
 By default, now it traps overflow and set it to the highest possible value and for underflow, it takes the lowest supported value.  With the latest emsdk update [this ](https://github.com/ni/VireoSDK/blob/c8eb11d197b09f7ed2100bdb01cd48cd8f2fa003/test-it/karma/publicapi/Double.Test.js#L140)vireo test was failing. We were getting mismatched values from JS Typed Array and WASM code. With this change, it is fixed now.

rajsite: The goal of the Double.Test.js was to make sure that writing values to memory using the writeDouble api and by getting access to the array buffer in memory behave the same. The writeDouble api does the float to int conversion in C++ while accessing the array buffer relies on the JS float to int conversion using the spec link devika gave.

Looking at how Vireo does float to int on the diagram shows that [Vireo saturates using ConvertFloatToInt](https://github.com/ni/VireoSDK/blob/c8eb11d197b09f7ed2100bdb01cd48cd8f2fa003/source/include/TypeAndDataManager.h#L366) instead which seems to align with NXG desktop with a quick test. @sanmut is this intentionally defined this way?

If that's the case then the test might be backwards. The WriteDoubleToMemory function should probably be calling ConvertFloatToInt instead of having it's own implementation of float to int (either the casting from before or the new approach mimicking the JS behavior). And the JS side should have a function available to help with float to int conversion when assigning to type array and should not assign values directly to the typed array.

I plan to discuss it with @sanmut in the morning to see if that makes sense.

sanmut: I agree with @rajsite. I confirmed the behavior of saturation of floating point to integral conversion with LabVIEW itself. It is better to use the existing "ConvertFloatToInt" as Milan described above.  We discussed this over a meeting with @Devika-Tantry and @rajsite to make the suggested changes in the code and modify the test.

rajsite (changes requested):
Discussed in a meeting, some outcomes:
- For WriteDoubleToMemory we should use ConvertFloatToInt instead and use it for all the num types, including 64bit
- The Double.Test.js should be updated to use the saturating values where relevant
- We should undo the failed ResizeDimension handling preferring to let the app segfault. We should create a Tech Debt to create an abort mechanism that can abort the runtime quickly closer to the allocation failure (maybe it can do so with a segfault? There are WASM opcodes to abort that we need to find out how to call from C++).
- The ExecuteSlicesUntilClumpsFinshed.Test.js error handling test should use a try catch around the execute slices and assert that the error happens. We should create a Tech Debt task to make an instruction that can intentionally fail with a seg fault and update the test to call that instruction.

source/core/Array.cpp
sanmut: Devika, Milan and I discussed about this also. We should revert this change because the node should fail with an error or the execution should stop on a fatal failure such as memory allocation failure. It is incorrect for the execution to continue. Devika will make the changes and try to fix the JS test to handle the exception.

sanmut (changes requested):

test-it/karma/publicapi/Double.Test.js
```diff
@@ -138,45 +132,26 @@ describe('The Vireo EggShell Double api can', function () {
...
-            testWriteDoubleCoerced('int8MinValue', -128, -Infinity, expectedCoercedValue(Int8Array, -Infinity));
-            testWriteDoubleCoerced('int8MinValue', -128, -0, expectedCoercedValue(Int8Array, -0));
-            testWriteDoubleCoerced('int8MinValue', -128, 0, expectedCoercedValue(Int8Array, 0));
-
-            expect(expectedCoercedValue(Int8Array, NaN)).toBe(0);
-            expect(expectedCoercedValue(Int8Array, Infinity)).toBe(0);
-            expect(expectedCoercedValue(Int8Array, -Infinity)).toBe(0);
-            expect(expectedCoercedValue(Int8Array, -0)).toBe(0);
-            expect(expectedCoercedValue(Int8Array, 0)).toBe(0);
+            testWriteDoubleCoerced('int8MinValue', -128, NaN, 127);
```
rajsite, line 150: @sanmut does this seem right to you? Double NaN [goes to IntTypeMax](https://github.com/ni/VireoSDK/blob/c8eb11d197b09f7ed2100bdb01cd48cd8f2fa003/source/include/TypeAndDataManager.h#L370)? Eventually the saturating WASM float to int will treat [NaN as zero](https://github.com/WebAssembly/spec/blob/master/proposals/nontrapping-float-to-int-conversion/Overview.md#design). We should align with LabVIEW so if LabVIEW is taking NaN to IntMax during float to int that's what we should replicate (wonder if it's compiler specific / architecture dependent).

source/include/TypeAndDataManager.h
Devika-Tantry: Done. Used `ConvertFloatToInt` for Double to Int conversion and removed all references to the Typed array in the test.

source/core/Array.cpp
Devika-Tantry: Reverted these changes and created [tech debt](https://ni.visualstudio.com/DevCentral/_workitems/edit/1507808/) item to track changes to stop the execution after fatal failure such as memory allocation failure.

Devika-Tantry:
> Discussed in a meeting, some outcomes:
> 
> * For WriteDoubleToMemory we should use ConvertFloatToInt instead and use it for all the num types, including 64bit
> * The Double.Test.js should be updated to use the saturating values where relevant
> * We should undo the failed ResizeDimension handling preferring to let the app segfault. We should create a Tech Debt to create an abort mechanism that can abort the runtime quickly closer to the allocation failure (looks like we can call the [C abort function](https://emscripten.org/docs/api_reference/module.html#Module.onAbort)).
> * The ExecuteSlicesUntilClumpsFinshed.Test.js error handling test should use a try catch around the execute slices and assert that the error happens. We should create a Tech Debt task to make an instruction that can intentionally fail with a seg fault and update the test to call that instruction.

- Updated `WriteDoubleToMemory` and `Double.Test.js` as per comment.
 - Created the tech debt to track the [`mechanism for abort for allocation failure`](https://ni.visualstudio.com/DevCentral/_workitems/edit/1507808/) and to fix the [`ExecuteSlicesUntilClumpsFinshed`](https://ni.visualstudio.com/DevCentral/_workitems/edit/1507825/) test.

test-it/karma/publicapi/Double.Test.js
```diff
@@ -56,12 +56,6 @@ describe('The Vireo EggShell Double api can', function () {
         expect(readDouble(path)).toMatchIEEE754Number(initialValue);
     };
 
-    // Expected coerced values of Doubles in Vireo should have the same behavior as assignment to typed arrays
-    // This is so a user assigning a double to a typedarray using getTypedArray will see the same effect as writing to a double using writeDouble
-    var expectedCoercedValue = function (TypedArrayConstructor, value) {
```
rajsite, line 61: Can you add a comment to [readTypedArray](https://github.com/ni/VireoSDK/blob/c8eb11d197b09f7ed2100bdb01cd48cd8f2fa003/source/io/module_eggShell.js#L472) that describes the change in assumption. Something like:


```js
// Note: If you are directly writing JavaScript numbers (floating point doubles) to 
// integer TypedArray views then be aware that integer TypedArray views will truncate
// overflowing values as specified in TypedArray conversion operations:
// https://tc39.es/ecma262/#table-the-typedarray-constructors
// Vireo requires that overflowing JavaScript numbers written to Vireo integers
// are saturated instead of truncated. Make sure to perform Vireo's float-to-int 
// conversion algorithm before directly writing to integer TypeArray views.
Module.eggShell.readTypedArray = publicAPI.eggShell.readTypedArray = function (valueRef) { ... };
```

And to [writeTypedArray](https://github.com/ni/VireoSDK/blob/c8eb11d197b09f7ed2100bdb01cd48cd8f2fa003/source/io/module_eggShell.js#L491):

```js
// Note: When creating integer TypedArrays views from JavaScript numbers to write to
// Vireo using writeTypedArray be aware that integer TypedArray views will truncate
// overflowing values as specified in TypedArray conversion operations:
// https://tc39.es/ecma262/#table-the-typedarray-constructors
// Make sure to perform Vireo's float-to-int conversion when creating integer TypedArray
// views from JavaScript numbers to write to Vireo memory.
Module.eggShell.writeTypedArray = publicAPI.eggShell.writeTypedArray = function (valueRef, typedArrayValue) {
```

rajsite (commented):
Looking good, just some minor tweaks remaining

rajsite:
@Devika-Tantry We have been creating tech debt items as GitHub issues so far: https://github.com/ni/VireoSDK/issues
I'd prefer to keep Vireo-specific tech debt issues tracked on GitHub if possible so the ideas of how the repo should look or be refactored can be localized to GitHub. Do you know if the squad wants to continue that way?

source/core/Array.cpp
```diff
@@ -190,7 +190,6 @@ VIREO_FUNCTION_SIGNATUREV(ArrayFillNDV, ArrayFillNDVParamBlock)
     }
 
     _Param(ArrayOut)->ResizeDimensions(numDimensionInputs, tempDimensionLengths, false);
-
```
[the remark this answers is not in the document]
Devika-Tantry: Reverted.

test-it/karma/publicapi/Double.Test.js
Devika-Tantry: Added above comments.

Devika-Tantry:
> @Devika-Tantry We have been creating tech debt items as GitHub issues so far: https://github.com/ni/VireoSDK/issues
> I'd prefer to keep Vireo-specific tech debt issues tracked on GitHub if possible so the ideas of how the repo should look or be refactored can be localized to GitHub. Do you know if the squad wants to continue that way?

Created two issues ([issue1](https://github.com/ni/VireoSDK/issues/682), [issue2](https://github.com/ni/VireoSDK/issues/683)) in github to track the issue and added these issue links in corresponding AzDo for squad tracking purpose.

.github/workflows/build.yml
```diff
@@ -20,10 +20,10 @@ jobs:
     # Install lint dependencies
     - uses: actions/setup-python@v2
       with:
-        python-version: '3.5'
+        python-version: '3.9'
```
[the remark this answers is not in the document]
Devika-Tantry: For windows, we don't have to change the minimum python and node version. (Also for local windows development, while installing the emsdk, it will install the required version of node and python package itself)
The problem is with the ubuntu build for that the python version has to updated to install properly (reverted node version changes)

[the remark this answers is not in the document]
rajsite: Thanks for the confirmation!

source/core/TypeAndDataManager.cpp
rajsite: Discussed in person and instead decided failed allocations should abort the runtime

test-it/karma/publicapi/Double.Test.js
rajsite: @sanmut friendly ping. Also maybe @spathiwa has some insight if this behavior is right for NaN handling in float-to-int conversions for LabVIEW?

rajsite (approved):

rajsite:
Unfortunately we don't have great performance benchmarks to check for regressions in compiler upgrades that I am aware of. There is the PiCalc test if y'all want to try it.

I did a quick check on size changes and it looks like the wasm file is about 45kb smaller (top is original and bottom is new compiler):
image

Free page size win! :)

rajsite:
@Devika-Tantry minor comment, won't block the PR, but thanks for the work on upgrading the compiler version! Make sure to add yourself to the [AUTHORS](https://github.com/ni/VireoSDK/blob/main/AUTHORS) list :)

test-it/karma/publicapi/Double.Test.js
spathiwa: @rajsite 
That's correct; NaN conversion to integer in LabVIEW saturates to maxInt, not zero.
In the CG runtime we don't rely on the compiler to do this because it would be architecture-dependent; we explicitly check for NaN/unordered.

Man, I kind of miss working on Vireo.  (I wonder if the Mac build still works.)

rajsite: > That's correct; NaN conversion to integer in LabVIEW saturates to maxInt, not zero.

Thanks for the confirmation! That resolves my question but @sanmut feel free to unresolve if you have any follow-up.

> Man, I kind of miss working on Vireo. (I wonder if the Mac build still works.)

Might be chances coming up as debugging is figured out :)

sanmut: Been busy with bugs. Craig beat me to it. Thanks Craig.
The only other thing I would recommend is to add some test cases such as conversion of different normal floating point values such as "-3.5, -2.1, 2.1, 3.5" to make sure the rounding behavior is FE_TONEAREST and ensure the default conversion behavior does not get affected with compiler upgrades. This can be added as a GVI test in ASW side too. It is not necessary to do this as part of this PR and not high priority.

sanmut (approved):

rajsite:
@Devika-Tantry Looks like the PR is almost ready 👍
Looking at the ASW PR where you tested changes yesterday one concern I have is that it only [seems to have run 8 tests](https://ni.visualstudio.com/DevCentral/_build/results?buildId=1064680&view=ms.vss-test-web.build-test-results-tab).

We should make sure that the tests run since this is a compiler change. I believe the issue may be that the Run-SquadTests.txt GUID needs to be changed for each build you want to trigger. The most recent commits in that PR do not seem to update the token.

test-it/karma/publicapi/Double.Test.js
rajsite: @sanmut that's interesting, hadn't thought about the rounding modes that are enabled. Looks like [WebAssembly currently only supports nearest](https://webassembly.github.io/spec/core/exec/numerics.html?highlight=rounding#rounding) but there is an interesting recent thread about [supporting more modes](https://github.com/WebAssembly/design/issues/1384#issuecomment-857373480).

@Devika-Tantry mentioning so that we capture the work to add tests verifying rounding modes. (mentioning just-in-case as the thread is marked resolved).

Devika-Tantry:
> My bad, that was the .NET Core build 🤦‍♂️ The [ASW PR build](https://ni.visualstudio.com/DevCentral/_build/results?buildId=1065912&view=ms.vss-test-web.build-test-results-tab) ran many more tests. Did you get a chance to evaluate the couple of failures that were in there?

These are some timeout error seems like these are common in PR build and these all not related to vireo changes. If we rerun, different set of [tests ](https://ni.visualstudio.com/DevCentral/_build/results?buildId=1073351&view=ms.vss-test-web.build-test-results-tab&runId=11448670&resultId=127705&paneView=debug) fail due to timeout.

test-it/karma/publicapi/Double.Test.js
Devika-Tantry: Done. Created the [issue ](https://github.com/ni/VireoSDK/issues/684)to add more test cases.

PR #696, opened 2021-10-26
Added flag to ignore debug point

Devika-Tantry:


source/include/TDCodecVia.h
```diff
@@ -118,22 +118,22 @@ class TDViaParser
...
+    void    ParseClump(VIClump* viClump, InstructionAllocator* cia, Boolean isDebuggingEnabled = false);
     void    PreParseClump(VIClump* viClump);
     SubString* TheString() {return &_string;}
     VirtualInstrument *CurrentVIScope() const { return _virtualInstrumentScope; }
 
  public:
-    static NIError StaticRepl(TypeManagerRef tm, SubString *replStream);
-    static void FinalizeVILoad(VirtualInstrument* vi, EventLog* pLog);
-    static void FinalizeModuleLoad(TypeManagerRef tm, EventLog* pLog);
+    static NIError StaticRepl(TypeManagerRef tm, SubString *replStream, Boolean isDebuggingEnabled = false);
```
rajsite, line 134: @shivaprasad-basavaraj Would it be possible to capture the debugging enabled flag in the TypeManager instead of passing it through as a flag? Conceptually I don't think I'd create a TypeManager instance where I'd want to parse some via with DebugPoints enabled and some where it isn't within the same TypeManager instance.
@spathiwa @sanmut do you think having that flag on the TypeManager instead is actually preferable? Is there a better place for these load options?

PR #697, opened 2021-12-09
Add an option to disable parsing of debug point instructions

shivaprasad-basavaraj:
Added an option that can be passed into the TDViaParser to instruct it to skip parsing the DebugPoint instructions. Made changes to the eggshell loadVia call as well so that the disable debugging option can be passed in.

Added tests for the new option.


shivaprasad-basavaraj:
@spathiwa @rajsite  @sanmut In the initial [draft PR](https://github.com/ni/VireoSDK/pull/696) for this change @rajsite had suggested that the option to disable debugging be part of the TypeManager so that we don't have to pass the option around. I wasn't sure if TypeManager is the right place since the option has meaning only in the context of a parse. Hence I have made it a variable in the TDViaParser class which can be used to decide whether or not to parse the debug point instruction.

Unfortunately, we create new parsers during the parse of a Via file and hence had to explicitly pass the option to the new parsers that were created. I also had to add the option as a parameter to the FinalizeVILoad and FinalizeModuleLoad methods since these were static. I am not sure why that is since these methods aren't being used outside the TDViaParser.

spathiwa:
I don't think there is any good reason for FInalizeVILoad and FInalizeModuleLoad to be static other than they didn't happen to need to access any member variables.  I think making them non-static to avoid needing to pass the new parameter would be fine.

spathiwa:
Otherwise, looks good to me.  I agree that TDViaParser is a logical place for the variable versus the type manager.

spathiwa (approved):

shivaprasad-basavaraj:
> I don't think there is any good reason for FInalizeVILoad and FInalizeModuleLoad to be static other than they didn't happen to need to access any member variables. I think making them non-static to avoid needing to pass the new parameter would be fine.

I have made the two methods non-static to avoid passing the parameter around.

source/core/TDCodecVia.cpp
```diff
@@ -1736,6 +1739,15 @@ void TDViaParser::ParseClump(VIClump* viClump, InstructionAllocator* cia)
             if (!_string.EatChar(')'))
                 return LOG_EVENT(kHardDataError, "')' missing");
             state.MarkPerch(&perchName);
+        } else if (instructionNameToken.CompareCStr("DebugPoint") && _isDebuggingDisabled) {
```
rajsite, line 1742: @shivaprasad-basavaraj Why is debugging opt out instead of opt in? Shouldn't Vireo ignore DebugPoint instructions unless we opt into debugging? There is no benefit to opting into running DebugPoint instructions by default right?

I think instead of `isDebuggingDisabled` the flag should be `debugging` and by default `debugging` is false.

source/io/module_eggShell.js
```diff
@@ -678,7 +678,7 @@ var assignEggShell;
             return str;
         };
 
-        Module.eggShell.loadVia = publicAPI.eggShell.loadVia = function (viaText) {
+        Module.eggShell.loadVia = publicAPI.eggShell.loadVia = function (viaText, isDebuggingDisabled) {
```
rajsite, line 681: Instead of a boolean flag we should have a configuration object. Coupled with the previous suggestion it should look something like:

```js
Module.eggShell.loadVia = publicAPI.eggShell.loadVia = function (viaText, config) {
    var debugging = typeof config?.debugging === 'boolean' ? config?.debugging : false;
```
and then you can pass debugging to _EggShell_REPL

test-it/karma/utilities/TestHelpers.VireoRunner.js
```diff
@@ -5,7 +5,7 @@
     'use strict';
     window.testHelpers = window.testHelpers || {};
 
-    var rebootAndLoadVia = function (vireo, viaAbsolutePath) {
+    var rebootAndLoadVia = function (vireo, viaAbsolutePath, isDebuggingDisabled) {
```
rajsite, line 8: Based on the other feedback we should change this name to config and pass it through to loadVia

test-it/karma/debugpoint/DebugPoint.Test.js
```diff
@@ -13,38 +13,67 @@ describe('A via file with debug point ', function () {
...
+        beforeAll(async function () {
+            vireo = await vireoHelpers.createInstance();
+        });
+
+        afterAll(function () {
+            vireo = undefined;
+        });
+
+        it('on executing succesfully will not set the needs update property on corresponding local ', async function () {
+            const runSlicesAsync = vireoRunner.rebootAndLoadVia(vireo, simpleAddNodeViaUrl, true);
```
rajsite, line 70: Based on the other feedback this becomes

```js
const runSlicesAsync = vireoRunner.rebootAndLoadVia(vireo, simpleAddNodeViaUrl, {debugging: true});
```

rajsite (changes requested):

source/core/TDCodecVia.cpp
shivaprasad-basavaraj: @rajsite I went back and forth on it a few times. The reason I made debugging opt out is that the user did choose to generate the DebugPoint instructions so I thought the normal flow in the execution engine should be to respect those instructions and the optional flow would be even if you did generate DebugPoint instructions don't consider them during execution. My thinking was that during a normal flow its the compiler who controls whether debug instructions should be emitted or not and the runtime just runs whatever instructions it has.

Once we have a concept of enabling/disabling debugging in the GWeb editor, I think we would skip generation of DebugPoints altogether.

rajsite: I still think it should be opt-in:
- all it does is add overhead to execution. If a user is running in a debug state they are using the runtime in an environment that does extra work to utilize the behavior of the debug instructions already. In any other context it just adds overhead.
- all else equal, the flag debugging is clearer than isDebuggingDisabled

rajsite: I can even give a somewhat concrete example, if PicoG adopts the new vireo and Gweb with Debug point it'll run slower by default https://github.com/PicoG/PicoVireo/blob/main/platform/rp2040/main.cpp#L220
They get no benefit from it being on by default unless they opt-in and add debugging features to compliment DebugPoint. By default it'll just regress in perf until they out out.

shivaprasad-basavaraj: I hadn't thought of the case where the user would use new GWeb and new Vireo in conjunction. I have now made the parsing of debugging instructions opt-in.

source/io/module_eggShell.js
shivaprasad-basavaraj: I have made the change to pass debugging option in a config object. I couldn't use the "?." operator since that is supported only from Chrome 80.

test-it/karma/utilities/TestHelpers.VireoRunner.js
shivaprasad-basavaraj: Done.

test-it/karma/debugpoint/DebugPoint.Test.js
shivaprasad-basavaraj: Done

source/CommandLine/main.cpp
```diff
@@ -42,16 +42,20 @@ int VIREO_MAIN(int argc, const char * argv[])
...
 
+        Boolean debugging = false;
         for (Int32 arg = 1; arg < argc; ++arg) {
             if (strcmp(argv[arg], "-D") == 0) {
                 gShells._pRootShell->DumpPrimitiveDictionary();
-                continue;
+            } else if (strcmp(argv[arg], "--debugging")) {
+                debugging = true;
             }
+        }
```
shivaprasad-basavaraj, line 52: I moved the processing of special command-line arguments into a separate loop so that we can detect the debugging flag and set it for all Via executions. I wanted to use a simple -D flag but that is already being used to print the primitives dictionary in use and hence went with a "--debugging" flag.

test-it/test.js
```diff
@@ -249,7 +249,7 @@
         });
 
         try {
-            vireo.eggShell.loadVia(viaCode);
+            vireo.eggShell.loadVia(viaCode, {debugging: true});
```
shivaprasad-basavaraj, line 252: I had to enable debugging for both our JS and Native test suite since we do have VTR tests which check for DebugPoint execution.

source/io/module_eggShell.js
```diff
@@ -705,7 +705,8 @@ var assignEggShell;
                 origPrintErr(textErr);
             };
 
-            var result = Module._EggShell_REPL(Module.eggShell.v_userShell, viaTextPointer, viaTextLength);
+            var debugging = (config === null || config === undefined || typeof config.debugging !== 'boolean') ? false : config.debugging;
```
rajsite, line 708: A more common pattern if we can't use optional chaining would be to check if config is truthy and then check the property. ie
```suggestion
            var debugging = (config && typeof config.debugging === 'boolean') ? config.debugging : false;
```

rajsite (approved):

PR #698, opened 2022-02-08
Bug fix for timeout event with 0ms doesn't work with webvi.

amehra-ni:
WebVI does not execute when an event structure is used with a timeout event of 0 ms.
BUG [link](https://ni.visualstudio.com/DevCentral/_workitems/edit/1474951)
Whenever event structure timeout was 0ms it goes to InitObservableTimerState with tickCount as 0x7FFFFFFFFFFFFFFF setting it to infinite wait instead of executing event structure immediately.

At first, I tried adding a condition in ReserveObservationStatesWithTimeout to check if tickCount>= 0 , pass tickCount in InitObservableTimerState so that we immediately trigger the observer whenever the timeout is 0. But during testing we found that it affected the behavior of other nodes as well such as the OneButtonDialog.
So, didn't move forward with this approach.
Now, I have added a condition if the timeout given is 0ms we execute the event structure immediately by going to the desired branch target. I have tested it with different configurations and the behavior was the same as the current vireo version.

Have added test cases and have also tested with other test cases.
@shivaprasad-basavaraj


source/core/Events.cpp
```diff
@@ -1048,6 +1048,9 @@ VIREO_FUNCTION_SIGNATUREV(WaitForEventsAndDispatch, WaitForEventsParamBlock)
         Int32 &staticCount = eventInfo->eventStructInfo[eventStructIndex].setCount;
 
         UInt32 msTimeout = timeOutPtr ? *timeOutPtr : -1;
+        if (msTimeout == 0) {
```
cglzaguilar, line 1051: In line 1058, there is a check: msTimeout > 0 ? ...

Did you try changing this line to msTimeout >= 0 ? ...

source/core/Events.cpp
```diff
@@ -1048,6 +1048,9 @@ VIREO_FUNCTION_SIGNATUREV(WaitForEventsAndDispatch, WaitForEventsParamBlock)
         Int32 &staticCount = eventInfo->eventStructInfo[eventStructIndex].setCount;
 
         UInt32 msTimeout = timeOutPtr ? *timeOutPtr : -1;
```
rajsite, line 1050: I'll delegate to @cglzaguilar @spathiwa @sanmut for these changes.

@amehra-ni make sure to turn on GitHub Actions in your Fork so you can see the CI builds in your branch. It currently does not look like your fork is testing your branches. You should see the GitHub Action configuration in: https://github.com/amehra-ni/VireoSDK/settings/actions

@amehra-ni before the PR is merged make sure to [run the squad tests](https://ni.visualstudio.com/DevCentral/_wiki/wikis/AppCentral.wiki/2445/How-to-update-Vireo-in-NXG?anchor=trigger-the-web-module-squad-tests) and link to a passing build in the PR description.

Marking approve pending the other CODEOWNERS feedback

rajsite (approved):

source/core/Events.cpp
```diff
@@ -1048,6 +1048,9 @@ VIREO_FUNCTION_SIGNATUREV(WaitForEventsAndDispatch, WaitForEventsParamBlock)
...
+        if (msTimeout == 0) {
+            return arguments->branchTarget;
```
spathiwa, line 1052: This isn't correct, you're not executing any of the event dispatch code below, so if an event is actually available, you will not dispatch to it and handle it.  Your tests only test events with a single timeout case so they do not fully test this.
Carlos's suggestion may actually be the correct fix.

spathiwa (changes requested):

sanmut (changes requested):
Please also post the VI snippet of the new tests also for each test (.via) file you add to the review.

amehra-ni:
Run the squad test - [pipeline](https://ni.visualstudio.com/DevCentral/_build/results?buildId=2144614&view=results)
Via snippet for both test files(just timeout changed)- 
test

cglzaguilar (approved):
I have not further comments, please address other feedback given

source/core/Events.cpp
amehra-ni: Took Carlos's suggestion. And also have updated the tests.

spathiwa (approved):

sanmut (approved):

PR #700, opened 2022-06-13
Unexpected Output from delete from array

amehra-ni:
Unexpected Output from delete from array [BUG](https://ni.visualstudio.com/DevCentral/_sprints/backlog/ASW%20WebVI/DevCentral/22C2/06?workitem=1757358)
Delete from array doesn't work correctly for non-flat arrays(>=2-d array) if in a loop. I have attached the behavior below. 

dfa

https://user-images.githubusercontent.com/97939637/173280839-d6393cc6-960f-4b72-b9d8-67d0e1234bf1.mp4

Delete from array works fine for the first iteration, but after the first iteration, the values in the input array somehow get deleted which also makes the deleted and output array reflect the same. In the second iteration when ResizeDimension is called on deletedArray(line 1316) and arrayOut(line 1323), and getInit happens for them, it basically initializes the input array's values as well which should not have happened. We found out this happens because deletedArray and arrayOut don't get copied correctly(from the copy subarray function) and the values are still related to the input array. So when getInit happens for the deleted and output array and their values are getting initialized and as they still point to the input array, values of the input array get changed.

In this PR, we have fixed the issue by changing how the non-flat array gets copied from the CopySubarray function. The changes are done for non-flat type arrays and for flat type it's still the same. I have used what we have for elementType->CopyData so that a proper copy(deep copy) happens.

I have tested the changes manually with different cases and also have added a test. The reason I didn't add it in ArrayDelete.via was as I have added a loop as well just to be sure. Attached the via below

CI Build in my branch - [build](https://github.com/amehra-ni/VireoSDK/actions/runs/2486274586) (Lint is failing because of 4 dead links. I tried searching why is it happening but didn't find anything significant)
Run the squad tests - [pipeline](https://ni.visualstudio.com/DevCentral/_build/results?buildId=2741969&view=results)
@shivaprasad-basavaraj


source/core/Array.cpp
```diff
@@ -1201,15 +1201,31 @@ VIREO_FUNCTION_SIGNATURET(ArrayMaxMinInternal, FindArrayMaxMinInstruction)
 }
```
rajsite, line 1201: @amehra-ni Can you update the broken links in the docs that are failing the build? https://github.com/ni/VireoSDK/runs/6856533744?check_suite_focus=true#step:8:152

It looks GitHub just moved them, if you visit the links in a browser you get redirected to the new locations ie

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/about-collaborative-development-models
https://docs.github.com/en/get-started/quickstart/fork-a-repo
https://docs.github.com/en/get-started/quickstart/fork-a-repo#cloning-your-forked-repository
https://docs.github.com/en/get-started/quickstart/fork-a-repo#configuring-git-to-sync-your-fork-with-the-original-repository

rajsite: I pushed a change to main that you can merge to disable link checking for the github docs and created the following issue: https://github.com/ni/VireoSDK/issues/701

spathiwa (approved):

sanmut:
Please do add more VI based tests in the test suite in its repo and validate. Also, it is a good idea to test those VIs against NXG Desktop or CurrentGen.

sanmut (approved):

amehra-ni:
Sure @sanmut, I'll add tests in NXG codebase.

cglzaguilar (approved):
I'll defer to Spathiwa and Samu on the changes. Please address their feedback.

Issue #703, opened 2024-03-20
-Infinity values are not handled correctly in JSON


rajsite:
A user found that when using UnflattenFromJSON for double arrays, if the JSON string encounters -Infinity then:
1. The node errors with code -375011
2. The node returns an array with partially parsed results

See forum: https://forums.ni.com/t5/G-Web-Development-Software/Bug-on-Unflatten-from-JSON/m-p/4361814?lightbox-message-images-4361814=330188i0448810F940767B1#M702

![image](https://github.com/ni/VireoSDK/assets/1588923/68634cf3-7b66-421e-92c3-67dc8e6ff51a)

Didn't thoroughly debug but scanning source the following line looks suspicious which seems to indicate that any negative value will not be treated like Infinity:

https://github.com/ni/VireoSDK/blob/673eb0dea01418f6ca20e6a2a53528a941eef7ca/source/core/TDCodecVia.cpp#L1207-L1207

spathiwa:
Agreed, it looks like that line will cause a -Inf to be treated as if it was an overflow a few lines below and return an error.

rajsite:
Created two AzDo bugs for the issue:
1. [Lack of support for -Infinity](https://dev.azure.com/ni/DevCentral/_workitems/edit/2696525)
2. [Returning partial / incorrect results on error](https://dev.azure.com/ni/DevCentral/_workitems/edit/2696528) (effectively user data corruption on error)

PR #706, opened 2024-07-16
Fix handling of -Infinity during json parsing

shivaprasad-basavaraj:
Part of #703 

The issue was that we were checking only against the leading character of the token in JSON to determine if a value was Infinity or Nan which led to us misidentifying -Infinity as not a Inf or Nan value. Changed the cheks to compare the whole token.

Added a via test for this case.


spathiwa (approved):

source/core/TDCodecVia.cpp
```diff
@@ -1074,6 +1074,21 @@ Boolean TDViaParser::EatJSONItem(SubString* input)
     return true;
 }
 
+bool IsInfinityOrNan(SubString token)
+{
+    bool isInfinityOrNan = false;
+    ConstCStr specialIEEEVals[] = { "infinity", "-infinity", "inf", "-inf", "nan", nullptr};
+    ConstCStr* specValPtr = specialIEEEVals;
+    while (*specValPtr) {
+        if (token.ComparePrefixCStrIgnoreCase(*specValPtr)) {
```
rajsite, line 1083: The behavior of the UnflattenFromJSON node should match LabVIEW's Unflatten From JSON which is stricter about the allowed representations:

image

The TDViaParser is used both for parsing VIA text content and for implementing the UnflattenFromJSON node. We kind of use `Fmt().UseFieldNames()` as a hint that we are doing the UnflattenFromJSON semantics for parsing. We should either use that as a hint to make sure that the UnflattenFromJSON node will align behavior with LabVIEW or add a new hint for this use-case, something like `Fmt().LabVIEWJSONExtensions()` since this is JSON extensions that are [LabVIEW-specific](https://www.ni.com/docs/en-US/bundle/labview-api-ref/page/functions/unflatten-from-json.html).

image

source/core/TDCodecVia.cpp
```diff
@@ -1074,6 +1074,21 @@ Boolean TDViaParser::EatJSONItem(SubString* input)
     return true;
```
rajsite, line 1074: @shivaprasad-basavaraj Our PR workflow also includes making sure the [squad workflow tests](https://ni.visualstudio.com/DevCentral/_wiki/wikis/AppCentral.wiki/2445/How-to-update-Vireo-in-NXG?anchor=trigger-the-web-module-squad-tests) are run with the [local changes](https://github.com/ni/VireoSDK/blob/main/docs/Workflow.md#testing-vireojs-on-a-different-machine--ci) and including a link to the CI build (Ideally also with some tests added on the ASW side as well). [Example Vireo PR with referenced build links](https://github.com/ni/VireoSDK/pull/700#issue-1268995457).

rajsite:
Closing for now, can reopen when ready to address changes and merge

